PR #1, opened 2014-07-11
Adding this.context and Thread.backtrace()

imcu:
Adding this.context and Thread.backtrace()

### DIFF
--- a/gum/gumreturnaddress.c
+++ b/gum/gumreturnaddress.c
@@ -13,6 +13,9 @@ gboolean
 gum_return_address_details_from_address (GumReturnAddress address,
                                          GumReturnAddressDetails * details)
 {
+
+#ifdef HAVE_SYMBOL_BACKEND
+
   GumSymbolDetails sd;
 
   if (gum_symbol_details_from_address (address, &sd))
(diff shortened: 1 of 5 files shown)


bashmd:
Hi, i have applied this patch and played around a bit, the additional _gum_symbol_util_init() seems to be needed, without it i can't get any backtraces and the target process segfaults - but it also doesn't seem to be the right thing to do, because some processes crash on me with that code, a problem which i have tracked down to load_dbghelp returning NULL. The reasons for that are beyond my skill level.

If someone with a better understanding could chime in and give me some hints as to how this could be fixed, i'd gladly test it.

oleavr:
`frida-gum` uses the `DbgHelp` library to resolve return addresses into symbolic names, using debug information. This is however not a good idea to load inside the target process. What we should do instead is to map each return address to ">modulename>+0x<offset>" or something similar. Then the debugger side can resolve those further if it wishes to.

oleavr:
Good news everyone. The latest code in git has `this.backtrace()` available in `onEnter` and `onLeave`, and there is also `DebugSymbol.fromAddress()` and `DebugSymbol.fromName()` available so you can do `this.backtrace().map(DebugSymbol.fromAddress)` to get a symbolicated backtrace.

PR #12, opened 2015-02-19
Fix reading memory from a subregion in Darwin

Tyilo:
The problem before was that if two subregions was contained inside a larger one, you would always get back the address of the first subregion even when the address you were reading from was in the second subregion.


oleavr:
Awesome! \o/

PR #15, opened 2015-02-19
Add Memory.readStlString and Memory.allocStlString

Tyilo:
Not sure if `Memory.readStlString` actually works, as trying to read the string from this tests program fails for me:

```
#include <iostream>
#include <string>
#include <malloc/malloc.h>

int main(int argc, const char *argv[]) {
	const char *a = strdup("foobarbazquz123456789");

	std::string *stl_str = new std::string(a);
	void *ptr = stl_str;

	printf("ptr: %p, size: %zu\n", ptr, malloc_size(ptr));
	printf("lengths: %zu, %zu\n", stl_str->length(), strlen(stl_str->c_str()));

	getchar();

	printf("foo: %p\n", (void *)stl_str);

    return 0;
}
```

```
$ frida-repl %Test\ cpp
Attaching...
>>> Memory.readStlString(ptr('0x7fd498c04c00'))
Error: access violation reading 0x7fd498c04c00
```


oleavr:
Awesome stuff! Could you verify that `otool -L %Test\ cpp` shows it's linked to the same C++ runtime? Recent OS X releases make use of libc++ instead of libstdc++.

oleavr:
Though looking more closely at the error message, the pointer is invalid, so the C++ runtime is not the first issue at least.

oleavr:
I'm sorry, my initial idea of how this should be implemented was way off. This feature should be implemented in JavaScript (like ObjC and Dalvik), as it will have to inspect the process and contain platform-specific logic – for example libstdc++ vs libc++ on Mac and iOS, and several different C runtimes on Windows.

oleavr:
Let's try the JS approach (in a separate PR); that should do it. Thanks!

PR #17, opened 2015-03-04
Add gum_arm64_writer_put_add_reg_reg_reg function

Tyilo:
This is needed for the arm64 fix in frida-core.

### DIFF
--- a/gum/arch-arm64/gumarm64writer.c
+++ b/gum/arch-arm64/gumarm64writer.c
@@ -722,9 +722,8 @@ gum_arm64_writer_put_add_reg_reg_reg (GumArm64Writer * self,
   g_assert_cmpuint (rd.width, ==, rr.width);
 
   guint32 flags = 0;
-  if(rd.width == 64) {
+  if (rd.width == 64)
     flags |= 0x8000000;
-  }
 
   gum_arm64_writer_put_instruction (self, rd.sf | 0xb000000 | flags | rd.index | (rl.index << 5) | (rr.index << 16));
 }
--- a/gum/arch-arm64/gumarm64writer.c
+++ b/gum/arch-arm64/gumarm64writer.c
@@ -707,6 +707,29 @@ gum_arm64_writer_put_add_reg_reg_imm (GumArm64Writer * self,
 }
 
 void
+gum_arm64_writer_put_add_reg_reg_reg (GumArm64Writer * self,
+                                      GumArm64Reg dst_reg,
+                                      GumArm64Reg left_reg,
+                                      GumArm64Reg right_reg)
+{
+  GumArm64RegInfo rd, rl, rr;
+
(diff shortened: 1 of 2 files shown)


Tyilo:
I haven't actually tested this, but I think it should work.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -707,6 +707,29 @@ gum_arm64_writer_put_add_reg_reg_imm (GumArm64Writer * self,
...
+
+  gum_arm64_writer_describe_reg (self, dst_reg, &rd);
+  gum_arm64_writer_describe_reg (self, left_reg, &rl);
+  gum_arm64_writer_describe_reg (self, right_reg, &rr);
+
+  g_assert_cmpuint (rd.width, ==, rl.width);
+  g_assert_cmpuint (rd.width, ==, rr.width);
+
+  guint32 flags = 0;
+  if(rd.width == 64) {
```
oleavr: Two nitpicks: should be space between `if` and `(`, and curly braces should align, though in this case they should be omitted to make the code more readable.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -707,6 +707,29 @@ gum_arm64_writer_put_add_reg_reg_imm (GumArm64Writer * self,
...
+    flags |= 0x8000000;
+  }
+
+  gum_arm64_writer_put_instruction (self, rd.sf | 0xb000000 | flags | rd.index | (rl.index << 5) | (rr.index << 16));
```
oleavr: Would be great to break the line at 80 cols (see style elsewhere in this file). Note that `frida-core` doesn't have an 80 col limit, as the plan is to move `frida-gum` to the same conventions later; but for now let's try to keep things consistent within each module.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -707,6 +707,29 @@ gum_arm64_writer_put_add_reg_reg_imm (GumArm64Writer * self,
...
+  GumArm64RegInfo rd, rl, rr;
+
+  gum_arm64_writer_describe_reg (self, dst_reg, &rd);
+  gum_arm64_writer_describe_reg (self, left_reg, &rl);
+  gum_arm64_writer_describe_reg (self, right_reg, &rr);
+
+  g_assert_cmpuint (rd.width, ==, rl.width);
+  g_assert_cmpuint (rd.width, ==, rr.width);
+
+  guint32 flags = 0;
```
oleavr: Would be good to move the declaration just after the `rd, rl, rr`, as declaration-after-statement is avoided through-out frida-gum due to Microsoft's terrible compiler. (I tend to declare variables in the same order as they're used.)

oleavr:
Yay, awesome!

PR #18, opened 2015-03-28
Optimize ObjC's ClassFactory.refresh

Tyilo:
It now does nothing if the number of classes hasn't changed.

### DIFF
--- a/gum/gumscript-runtime-objc.js
+++ b/gum/gumscript-runtime-objc.js
@@ -137,6 +137,12 @@
 
         this.refresh = function refresh() {
             var numClasses = api.objc_getClassList(NULL, 0);
+            if(numClasses === classNames.length) {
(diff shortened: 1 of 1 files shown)


Tyilo:
Maybe we should just call `ObjC.refreshClasses` automatically now every time a class is requested?

oleavr:
Sweet! Just one coding style nitpick and this is ready to land (space after `if`).

oleavr:
Yeah, that sounds good to me (as long as it's not in the cache).

Tyilo:
Yeah, but then it should probably be removed from the public API.

oleavr:
Agreed. Let's do that and bump the major version for the next release. (Also planned is renaming `frida-repl` to just `frida` so timing is good.)

PR #19, opened 2015-03-29
Add and & or method to NativePointer

Tyilo:


gum/gumscriptcore.cpp
```diff
@@ -899,6 +909,60 @@ gum_script_core_on_native_pointer_sub (
...
+      Local<FunctionTemplate>::New (self->isolate, *self->native_pointer));
+  if (native_pointer->HasInstance (info[0]))
+  {
+    guint64 rhs = reinterpret_cast<guint64> (
+        GUM_NATIVE_POINTER_VALUE (info[0].As<Object> ()));
+    result = GSIZE_TO_POINTER (lhs & rhs);
+  }
+  else
+  {
+    result = GSIZE_TO_POINTER (lhs & (guint64)(info[0]->ToInteger ()->Value ()));
```
oleavr: Two minor coding style issues:
- Missing space after cast
- Exceeds 80 columns (someday we'll bump this in frida-gum too, but let's keep it consistent for now). Perhaps just extract the right part of the expression into an rhs like above? (Easier to read.)

oleavr:
Sweet! Let's land this.

PR #25, opened 2015-04-15
Define _QNX_SOURCE for frida-gum

s1341:


configure.ac
```diff
@@ -69,6 +69,7 @@ if [[ "x$HAVE_DARWIN" = "xyes" ]]; then
 fi
 if [[ "x$HAVE_QNX" = "xyes" ]]; then
   AC_DEFINE(HAVE_QNX, 1, [Define to 1 if OS is QNX based.])
+  AC_DEFINE(_QNX_SOURCE, 1, [Define to 1 if OS is QNX based.])
```
oleavr: Could we instead do this in the source files where we need this? (There's only two of them or so right?)

s1341: I think there are a few locations. Isn't this the best way to do this?

PR #29, opened 2015-04-19
Move gum_try_mprotect from backend-posix to backend-linux

s1341:
In preparation for having an instance of this function in backend-qnx which behaves differently.

### DIFF
--- a/gum/backend-linux/gummemory-linux.c
+++ b/gum/backend-linux/gummemory-linux.c
@@ -6,6 +6,8 @@
 
 #include "gummemory.h"
 
+#include "gummemory-priv.h"
+
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


gum/backend-linux/gummemory-linux.c
```diff
@@ -87,6 +87,31 @@ gum_clear_cache (gpointer address,
 #endif
 }
 
+gboolean
```
oleavr: It would be great to move this function between `gum_memory_write` and `gum_clear_cache` to stay consistent with the order in the header file (gummemory.h).

gum/gummemory.h
```diff
@@ -88,6 +88,8 @@ gpointer gum_alloc_n_pages (guint n_pages, GumPageProtection page_prot);
 gpointer gum_alloc_n_pages_near (guint n_pages, GumPageProtection page_prot, GumAddressSpec * address_spec);
 void gum_free_pages (gpointer mem);
 
+gint gum_page_protection_to_posix (GumPageProtection page_prot);
```
oleavr: This header file is public and available on all OSes; we need to introduce an internal header file. Remember to prefix the function with an underscore and mark it G_GNUC_INTERNAL.

gum/gummemory-priv.h
```diff
@@ -31,4 +31,6 @@ struct _GumMatchToken
   guint offset;
 };
 
+G_GNUC_INTERNAL gint _gum_page_protection_to_posix (GumPageProtection page_prot);
```
oleavr: This line crosses the 80 cols limit, so it should be broken like this:
```C
G_GNUC_INTERNAL gint _gum_page_protection_to_posix (
    GumPageProtection page_prot);
```

PR #31, opened 2015-04-20
Make gum_try_mprotect clear the ICACHE when copying

s1341:
When copying the blob in the hack for PROC_WRITE on code sections,
we should really clear the ICACHE after doing our copy so that code
in the copied region can execute immediately.

### DIFF
--- a/gum/backend-qnx/gummemory-qnx.c
+++ b/gum/backend-qnx/gummemory-qnx.c
@@ -152,10 +152,12 @@ gum_try_mprotect (gpointer address,
     }
 
     address_mmaped = mmap (aligned_address, aligned_size,
-        PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_FIXED, NOFD, 0);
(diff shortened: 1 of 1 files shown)


gum/backend-qnx/gummemory-qnx.c
```diff
@@ -152,11 +152,13 @@ gum_try_mprotect (gpointer address,
     }
 
     address_mmaped = mmap (aligned_address, aligned_size,
-        PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_FIXED, NOFD, 0);
+        PROT_READ | PROT_WRITE | PROT_EXEC, MAP_PRIVATE | MAP_FIXED, NOFD, 0);
     g_assert (address_mmaped == aligned_address);
 
     memcpy (aligned_address, buffer, aligned_size);
 
+    gum_clear_cache(aligned_address, aligned_size);
```
oleavr: Missing space after `gum_clear_cache`.

PR #35, opened 2015-04-21
Add capstone to armrelocator; implement PC relative ldr rewrite.

s1341:
Add capstone to the ARM Relocator, allowing for parsing instructions
using the capstone API.

Also, implement rewrite of PC relative ldr instructions (where PC
is the base register for the second operand).

### DIFF
--- a/gum/arch-arm/gumarmrelocator.c
+++ b/gum/arch-arm/gumarmrelocator.c
@@ -16,18 +16,30 @@ struct _GumCodeGenCtx
 {
   const GumArmInstruction * insn;
   guint32 raw_insn;
+  cs_insn * capstone_insn;
 
   GumArmWriter * output;
 };
 
 static gboolean gum_arm_relocator_rewrite_branch_imm (GumArmRelocator * self,
     GumCodeGenCtx * ctx);
+static gboolean gum_arm_relocator_rewrite_pc_relative_ldr (GumArmRelocator * self,
+    GumCodeGenCtx * ctx);
+
+static gint gum_capstone_reg_to_arm_reg (gint cs_reg);
 
 void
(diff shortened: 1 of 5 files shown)


gum/arch-arm/gumarmrelocator.c
```diff
@@ -217,6 +256,7 @@ gum_arm_relocator_write_one (GumArmRelocator * self)
   if (!rewritten)
     gum_arm_writer_put_bytes (ctx.output, ctx.insn->address, ctx.insn->length);
 
+  cs_free (ci, 1);
```
oleavr: Empty line before return would be good.

gum/arch-arm/gumarmrelocator.c
```diff
@@ -345,3 +385,28 @@ gum_arm_relocator_rewrite_branch_imm (GumArmRelocator * self,
...
+static gboolean
+gum_arm_relocator_rewrite_pc_relative_ldr (GumArmRelocator * self,
+                                           GumCodeGenCtx * ctx)
+{
+  cs_insn * ci = ctx->capstone_insn;
+  gint dest_reg, base_reg, value;
+
+  dest_reg = gum_capstone_reg_to_arm_reg (ci->detail->arm.operands[0].reg);
+  base_reg = ci->detail->arm.operands[1].mem.base;
+  value = *((gint *) GSIZE_TO_POINTER (ctx->insn->pc + ci->detail->arm.operands[1].mem.disp));
```
oleavr: Hmm, is this safe in case of self-modifying code? Shouldn't we just load the target register with what the PC would have been at the original instruction, then add to it, and dereference?

oleavr:
Sweet!

PR #36, opened 2015-04-21
Add additional shortcuts for reading and writing memory.

SomeoneWeird:
This PR adds the following methods to `Memory`

* readShort
* writeShort
* readUShort
* writeUShort
* readInt
* writeInt
* readUInt
* writeUInt
* readLong
* writeLong
* readULong
* writeULong

Not sure what your thoughts are, but I added `longSize` to the top of `gumscript-runtime-core.js` so it's only calculated once, and not each call.

Also, the tests for **\*Long** are missing, I'm not entirely sure how to get the proper sized value for the current platform in C.

### DIFF
--- a/gum/gumscript-runtime-core.js
+++ b/gum/gumscript-runtime-core.js
@@ -6,6 +6,8 @@
         dispatcher = new MessageDispatcher();
     };
 
+    const longSize = (Process.pointerSize == 8 && Process.platform !== 'windows') ? 64 : 32;
+
     Object.defineProperty(engine, 'recv', {
         enumerable: true,
         value: function recv() {
@@ -64,6 +66,90 @@
             var result = Memory.alloc(size);
             Memory.copy(result, mem, size);
             return result;
+        }
+    });
+
+    Object.defineProperty(Memory, 'readShort', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory.readS16(mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeShort', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory.writeS16(mem, value);
+        }
+    });
+
+    Object.defineProperty(Memory, 'readUShort', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory.readU16(mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeUShort', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory.writeU16(mem, value);
+        }
+    });
+
+    Object.defineProperty(Memory, 'readInt', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory.readS32(mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeInt', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory.writeS32(mem, value);
+        }
+    });
+
+    Object.defineProperty(Memory, 'readUInt', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory.readU32(mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeUInt', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory.writeU32(mem, value);
+        }
+    });
+
+    Object.defineProperty(Memory, 'readLong', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory['readS' + longSize](mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeLong', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory['writeS' + longSize](mem, value);
+        }
+    });
+
+    Object.defineProperty(Memory, 'readULong', {
+        enumerable: true,
+        value: function (mem) {
+            return Memory['readU' + longSize](mem);
+        }
+    });
+
+    Object.defineProperty(Memory, 'writeULong', {
+        enumerable: true,
+        value: function (mem, value) {
+            return Memory['writeU' + longSize](mem, value);
         }
     });
 
--- a/tests/core/script.c
+++ b/tests/core/script.c
@@ -53,6 +53,18 @@ TEST_LIST_BEGIN (script)
   SCRIPT_TESTENTRY (s64_can_be_written)
   SCRIPT_TESTENTRY (u64_can_be_read)
   SCRIPT_TESTENTRY (u64_can_be_written)
+  SCRIPT_TESTENTRY (short_can_be_read)
+  SCRIPT_TESTENTRY (short_can_be_written)
+  SCRIPT_TESTENTRY (ushort_can_be_read)
+  SCRIPT_TESTENTRY (ushort_can_be_written)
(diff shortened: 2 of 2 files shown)


tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
   g_assert_cmpint (val, ==, G_GUINT64_CONSTANT (1201239876783));
 }
 
+SCRIPT_TESTCASE (short_can_be_read)
+{
+  gint16 val = -12123;
```
oleavr: It would be good to use the naked C types here, e.g. `short` in this case.

oleavr:
Awesome stuff! Perhaps it would be better to plug in a different JS function when defining the property instead of using a variable?

oleavr:
Or no, never mind, I think your approach is better.

SomeoneWeird:
Updated.

gum/gumscript-runtime-core.js
```diff
@@ -6,6 +6,8 @@
         dispatcher = new MessageDispatcher();
     };
 
+    var longSize = Process.arch === 'x64' || Process.platform === 'windows' ? 64 : 32;
```
oleavr: It would be good to use ES6 here and say:
`const longSize = ...`
I suspect the condition should be like this though:
```JavaScript
const longSize = (Process.pointerSize == 8 && Process.platform !== 'windows') ? 64 : 32;
```

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+SCRIPT_TESTCASE (short_can_be_written)
+{
+  short val = 0;
+  COMPILE_AND_LOAD_SCRIPT ("Memory.writeShort(" GUM_PTR_CONST ", -12123);", &val);
+  g_assert_cmpint (val, ==, -12123);
+}
+
+SCRIPT_TESTCASE (ushort_can_be_read)
+{
+  short val = 12123;
```
oleavr: Shouldn't this be `unsigned short`?

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("send(Memory.readUShort(" GUM_PTR_CONST "));", &val);
+  EXPECT_SEND_MESSAGE_WITH ("12123");
+}
+
+SCRIPT_TESTCASE (ushort_can_be_written)
+{
+  short val = 0;
```
oleavr: (and this one too)

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("Memory.writeUShort(" GUM_PTR_CONST ", 12123);", &val);
+  g_assert_cmpint (val, ==, 12123);
+}
+
+SCRIPT_TESTCASE (int_can_be_read)
+{
+  gint32 val = -120123;
```
oleavr: Here `int` would be more appropriate.

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("send(Memory.readInt(" GUM_PTR_CONST "));", &val);
+  EXPECT_SEND_MESSAGE_WITH ("-120123");
+}
+
+SCRIPT_TESTCASE (int_can_be_written)
+{
+  gint32 val = 0;
```
oleavr: (and here too)

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("Memory.writeInt(" GUM_PTR_CONST ", -120123);", &val);
+  g_assert_cmpint (val, ==, -120123);
+}
+
+SCRIPT_TESTCASE (uint_can_be_read)
+{
+  guint32 val = 120123;
```
oleavr: Here `unsigned int` would be more appropriate.

tests/core/script.c
```diff
@@ -1498,6 +1506,62 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("send(Memory.readUInt(" GUM_PTR_CONST "));", &val);
+  EXPECT_SEND_MESSAGE_WITH ("120123");
+}
+
+SCRIPT_TESTCASE (uint_can_be_written)
+{
+  guint32 val = 0;
```
oleavr: (and here too)

oleavr:
Regarding `long` tests, those should be following the same pattern, right? `long val = xxx;`

SomeoneWeird:
Updated, added tests for long (i'm super tired, of course that's going to work, I think those values are correct)

tests/core/script.c
```diff
@@ -1498,6 +1510,90 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+SCRIPT_TESTCASE (uint_can_be_written)
+{
+  unsigned int val = 0;
+  COMPILE_AND_LOAD_SCRIPT ("Memory.writeUInt(" GUM_PTR_CONST ", 120123);", &val);
+  g_assert_cmpint (val, ==, 120123);
+}
+
+SCRIPT_TESTCASE (long_can_be_read)
+{
+  unsigned long val = -2147483648;
```
oleavr: Should be just `long`, right?

tests/core/script.c
```diff
@@ -1498,6 +1510,90 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+  COMPILE_AND_LOAD_SCRIPT ("send(Memory.readLong(" GUM_PTR_CONST "));", &val);
+  EXPECT_SEND_MESSAGE_WITH ("-2147483648");
+}
+
+SCRIPT_TESTCASE (long_can_be_written)
+{
+  unsigned long val = 0;
```
oleavr: (and here too)

SomeoneWeird:
Updated

tests/core/script.c
```diff
@@ -1498,6 +1510,90 @@ SCRIPT_TESTCASE (u64_can_be_written)
...
+{
+  long val = -2147483648;
+  COMPILE_AND_LOAD_SCRIPT ("send(Memory.readLong(" GUM_PTR_CONST "));", &val);
+  EXPECT_SEND_MESSAGE_WITH ("-2147483648");
+}
+
+SCRIPT_TESTCASE (long_can_be_written)
+{
+  long val = 0;
+  COMPILE_AND_LOAD_SCRIPT ("Memory.writeLong(" GUM_PTR_CONST ", 1350966097);", &val);
```
oleavr: Please break lines at 80 cols, and indent like this:
```C
  COMPILE_AND_LOAD_SCRIPT ("Memory.writeLong(" GUM_PTR_CONST ", 1350966097);",
      &val);
```
(Continuation has a 4 space indent)

SomeoneWeird:
Done.

oleavr:
Awesome! Thank you!

PR #37, opened 2015-04-21
Fix replace_function test case

s1341:
The ret variable in this case needs to be volatile so that the
compiler doesn't aggressively optimize, resulting in an assert
0x42 == 0x42.


oleavr:
Great catch!

PR #40, opened 2015-04-23
Fix the pc relative add

s1341:
Now correct for both dest_reg == src_reg and dest_reg != src_reg.

Also fixed test.


oleavr:
Merged, thanks!

PR #61, opened 2015-06-06
Process.getModuleByName(name) implemented

marc1006:
Should we also handle it if there are modules with the same name?

### DIFF
--- a/bindings/gumjs/gumscript-runtime-core.js
+++ b/bindings/gumjs/gumscript-runtime-core.js
@@ -211,7 +211,7 @@
                 },
                 onComplete: function () {
                     if (!module) {
-                        throw new Error('Module ' + name + ' not found');
+                        throw new Error("Unable to find module '" + name + "'");
                     }
                 }
             });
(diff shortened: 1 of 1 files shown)


bindings/gumjs/gumscript-runtime-core.js
```diff
@@ -198,6 +198,27 @@
...
+            Process.enumerateModules({
+                onMatch: function (m) {
+                    if (m.name === name) {
+                        module = m;
+                        return 'stop';
+                    }
+                },
+                onComplete: function () {
+                    if (!module) {
+                        throw new Error('Module ' + name + ' not found');
```
oleavr: It would be good to make this consistent with the other error message, so something like:
```JavaScript
"Unable to find module '" + name + "'"
```

oleavr: Btw, we currently use double-quotes for strings, and single-quotes for “enums” (pretending that JavaScript has them) :)

oleavr:
Sweet! Thanks!

PR #62, opened 2015-06-07
Add Process.enumerateMallocRanges{,Sync}

Tyilo:


bindings/gumjs/gumscriptprocess.cpp
```diff
@@ -422,3 +429,90 @@ gum_script_process_handle_range_match (const GumRangeDetails * details,
...
+  GumScriptMatchContext ctx;
+
+  ctx.self = static_cast<GumScriptProcess *> (
+      info.Data ().As<External> ()->Value ());
+  ctx.isolate = info.GetIsolate ();
+
+#if !defined(HAVE_DARWIN)
+  ctx->isolate->ThrowException (Exception::Error (String::NewFromUtf8 (
+      ctx.isolate, "Process.enumerateMallocRanges: not implemented yet for "
+      GUM_SCRIPT_PLATFORM)));
```
oleavr: Missing `return` statement here?

Tyilo: Yeah

oleavr:
Looking great. The only issue we need to fix is to ifdef out all the meat, like the rest of the function plus the callbacks, to avoid warnings on non-Darwin. It would also be good to get rid of the left-over newlines.

PR #64, opened 2015-06-08
Add NativePointer.cmp for comparing pointers

Tyilo:
We should probably also implement `ptr1.equals(ptr2)`, but we can do that in javascript, by using `ptr1.cmp(ptr2) === 0`.


oleavr:
Wouldn't `compare()` be more idiomatic? Considering that there is `String#localeCompare()`.

Tyilo:
Yeah, I agree.

oleavr:
Thanks!

PR #67, opened 2015-06-11
Get object classname

marc1006:
It's useful:
- if you have a reference to an object and you don't know the class of it
- if you call a generic method in Java and you want to determine the class of the result
- to determine the real class if it's type is an interface

```
...
let instance = Dalvik.cast(p, klass);
console.log(instance.constructor.name);
console.log(instance.class);
console.log(Dalvik.getObjectClassname(instance));
console.log(Dalvik.getObjectClassname(p));
```

Results
```
ITest
interface com.example.example.ITest
com.example.example.Test1
com.example.example.Test1
```

### DIFF
--- a/bindings/gumjs/gumscript-runtime-dalvik.js
+++ b/bindings/gumjs/gumscript-runtime-dalvik.js
@@ -272,11 +272,12 @@
             let handle = obj.hasOwnProperty('$handle') ? obj.$handle : obj;
             if (handle instanceof NativePointer) {
                 let env = vm.getEnv();
-                let jklass = env.getObjectClass(obj);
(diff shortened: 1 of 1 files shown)


Tyilo:
How about using `instance.$className` to match Objective-C instances?

oleavr:
Nice! Yes, +1 on the suggestion from @Tyilo, we should try to stick to the same convention that we use in ObjC.

marc1006:
Okay - I will add it in this way.
And what should we do if we have only the (jobject) reference to it? (and nothing else)

oleavr:
Cool. In that case we need a way to construct an object wrapper from a `jobject`.

oleavr:
Sweet! Thanks!

PR #68, opened 2015-06-11
Implemented 'Dalvik.choose(classname, callbacks);'

marc1006:
Example:
```
Dalvik.perform(function () {
    Dalvik.choose("java.lang.String",
        {
            onMatch: function (instance) {
                console.log(instance);
            },
            onError: function (reson) {
            },
            onComplete: function () {
                console.log('Done.');
            }
        });
});
```

### DIFF
--- a/bindings/gumjs/gumscript-runtime-dalvik.js
+++ b/bindings/gumjs/gumscript-runtime-dalvik.js
@@ -186,6 +186,13 @@
             return classFactory.use(className);
         };
 
+        this.choose = function (className, callbacks) {
+            if (classFactory.loader === null) {
+                throw new Error("Not allowed outside Dalvik.perform() callback");
+            }
+            return classFactory.choose(className, callbacks);
+        };
+
         this.cast = function (obj, C) {
             return classFactory.cast(obj, C);
         };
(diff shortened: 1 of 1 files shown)


bindings/gumjs/gumscript-runtime-dalvik.js
```diff
@@ -258,6 +265,67 @@
...
+                                let stopMaybe = callbacks.onMatch(instance);
+                                env.deleteLocalRef(localReference);
+                                if (stopMaybe === 'stop') {
+                                    return 'stop';
+                                }
+                            });
+                        }
+                    },
+                    onError: function (reason) {
+                        callbacks.onError(reason);
```
oleavr: I suppose we don't want to propagate this?

bindings/gumjs/gumscript-runtime-dalvik.js
```diff
@@ -258,6 +265,67 @@
...
+                let size = heapSourceLimit.toInt32() - heapSourceBase.toInt32();
+                Memory.scan(heapSourceBase, size, pattern, {
+                    onMatch: function (address, size) {
+                        if (api.dvmIsValidObject(address)) {
+                            Dalvik.perform(function () {
+                                let env = vm.getEnv();
+                                let thread = Memory.readPointer(env.handle.add(JNI_ENV_OFFSET_SELF));
+                                let localReference = api.addLocalReferenceFunc(thread, address);
+                                let instance = Dalvik.cast(localReference, klass);
+                                let stopMaybe = callbacks.onMatch(instance);
```
oleavr: It would be good to use `const` instead of `let` for variables that are only assigned to once in the current scope.

bindings/gumjs/gumscript-runtime-dalvik.js
```diff
@@ -1643,13 +1711,32 @@
             return _api;
         }
 
-        var temporaryApi = {};
+        var temporaryApi = {addLocalReferenceFunc: null};
```
oleavr: Would be great to split this across three lines:

```JavaScript
var temporaryApi = {
    addLocalReferenceFunc: null
};
```

(either that or add a space on each end of `{}` for consistency).

PR #70, opened 2015-06-15
Use PT_DYNAMIC to enumerate dynamic symbols on QNX

s1341:
This is the way that the dynamic linker itself looks up symbols, as there is no .dynamic section in a stripped QNX binary.


oleavr:
LGTM except some minor style inconsistencies (missing space after cast, and block inside case that should be aligned with the case).

PR #71, opened 2015-06-16
Switch from fopen to open for memory read/writes

s1341:
fopen/fread/fwrite might be buffered, open/read/write is not. This was causing an issue where there was a partial read when doing the 'copy on mprotect to write' in gum_try_mprotect.

### DIFF
--- a/gum/backend-qnx/gummemory-qnx.c
+++ b/gum/backend-qnx/gummemory-qnx.c
@@ -55,17 +55,18 @@ gum_memory_read (GumAddress address,
                  gsize len,
                  gsize * n_bytes_read)
 {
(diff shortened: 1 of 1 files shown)


s1341:
I forgot that I also switched the allocation in gum_try_mprotect to be an alloca instead of malloc. This is because we're encountering fail when trying to Intercept malloc itself.

oleavr:
Nice. I suppose lack of EINTR handling isn't a big deal?

s1341:
Yeah. I think the likelihood of EINTR is low. I don't want to add handling until I see it actually occur in the wild.

oleavr:
:+1:

PR #74, opened 2015-06-16
Various ObjC improvements

Tyilo:
- ObjC.nil
- ObjC.api: various ObjC runtime functions
- ObjC.Object.equals
- ObjC.Object.isNil
- ObjC.Object.$superClass
- ObjC.Object.$methods: list of methods prefixed with +/-

### DIFF
--- a/bindings/gumjs/gumscript-runtime-objc.js
+++ b/bindings/gumjs/gumscript-runtime-objc.js
@@ -1089,6 +1089,9 @@
                 if (specifier.hasOwnProperty('subclasses'))
                     subclasses = specifier.subclasses;
             }
+            if (typeof cls === 'string') {
+                cls = ObjC.classes[cls];
+            }
             if (!(cls instanceof ObjCObject && (cls.$kind === 'class' || cls.$kind === 'meta-class')))
                 throw new Error("Expected an ObjC.Object for a class or meta-class");
             const ptr = cls.handle;
--- a/bindings/gumjs/gumscript-runtime-objc.js
+++ b/bindings/gumjs/gumscript-runtime-objc.js
@@ -3,6 +3,7 @@
 
     let _runtime = null;
     let _api = null;
+    let cachedObjCApi = {};
 
     Object.defineProperty(this, 'ObjC', {
         enumerable: true,
@@ -29,6 +30,16 @@
             get: function () {
                 return api !== null;
             }
+        });
+
+        Object.defineProperty(this, 'nil', {
+            enumerable: true,
+            value: new ObjCObject(NULL)
+        });
+
+        Object.defineProperty(this, 'api', {
+            enumerable: true,
+            value: cachedObjCApi
         });
 
         Object.defineProperty(this, 'classes', {
@@ -328,6 +339,22 @@
             return registry;
         }
 
+        function getPointer(obj) {
+            if (typeof obj === 'string') {
+                return new NativePointer(obj);
+            }
+            if (obj instanceof NativePointer) {
+                return obj;
+            }
+            if (obj instanceof ObjCObject) {
+                return obj.handle;
+            }
+            if (typeof obj === 'object' && obj.hasOwnProperty('handle')) {
+                return getPointer(obj.handle);
+            }
+            throw new Error("Expected NativePointer or ObjC.Object instance");
+        }
+
         const objCObjectBuiltins = {
             "prototype": true,
             "handle": true,
@@ -335,17 +362,22 @@
             "toJSON": true,
             "toString": true,
             "valueOf": true,
+            "equals": true,
+            "isNil": true,
             "$kind": true,
             "$super": true,
+            "$superClass": true,
             "$class": true,
             "$className": true,
-            "$protocols": true
+            "$protocols": true,
+            "$methods": true
         };
 
         function ObjCObject(handle, protocol, cachedIsClass, superSpecifier) {
             let cachedClassHandle = null;
             let cachedKind = null;
             let cachedSuper = null;
+            let cachedSuperClass = null;
             let cachedClass = null;
             let cachedClassName = null;
             let cachedProtocols = null;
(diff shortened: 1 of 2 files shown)


bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -328,6 +339,22 @@
             return registry;
         }
 
+        function getPointer(obj) {
```
oleavr: How about `ObjC.handleOf()`? As we're using `handle` already.

bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -328,6 +339,22 @@
...
+            if (typeof obj === 'string') {
```
oleavr: It would be good to drop the block scope for single-statement bodies for consistency.

bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -389,6 +413,15 @@
                         case "valueOf":
                             const description = target.description();
                             return description.UTF8String.bind(description);
+                        case "equals":
+                            return function equals (ptr) {
```
oleavr: How about moving this function down next to the others and just returning it here?

bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -443,7 +486,22 @@
...
+                                // Fill cachedMethodNames
+                                this.keys();
+
+                                cachedNativeMethodNames = Object.keys(cachedMethods).filter(m => m.match(/^(\+|-)/));
+                            }
+                            return cachedNativeMethodNames;
                         default:
+                            if (handle.isNull()) {
+                                return function() {
+                                    return 0;
```
oleavr: Wouldn't it be better to return `ObjC.nil` instead? So you can keep calling methods on every object returned and it will "just work".

bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -536,7 +594,7 @@
                         }
                     }
 
-                    return cachedMethodNames;
+                    return Object.keys(objCObjectBuiltins).concat(cachedMethodNames);
```
oleavr: Pedantically we shouldn't return all the builtins here; we should only return the properties and not things on the prototype.

Tyilo: We want these for autocompletion in the repl. Any other way of doing that?

bindings/gumjs/gumscript-runtime-objc.js
Tyilo: I don't think we should expose it. So maybe just `getHandle`?

oleavr: Sounds good (and I agree, we shouldn't expose it).

bindings/gumjs/gumscript-runtime-objc.js
Tyilo: The problem with this is that in Objective-C `0`, `NULL` and `nil` is the same, but there are not the same in Frida.
So if you expected the method to return an int, it would work in Objective-C, but not if we use `ObjC.nil`.
I don't know if the best choice is `0`, `NULL` or `ObjC.nil`.

bindings/gumjs/gumscript-runtime-objc.js
oleavr: Hmm, it's the same way we'd auto-complete a String object; the methods will be on the prototype and presumably not part of Object.keys() if I'm not mistaken.

bindings/gumjs/gumscript-runtime-objc.js
```diff
@@ -1031,6 +1089,9 @@
                 if (specifier.hasOwnProperty('subclasses'))
                     subclasses = specifier.subclasses;
             }
+            if (typeof cls === 'string') {
```
oleavr: It would be good to keep the API as explicit as possible, so that's why I removed this.

bindings/gumjs/gumscript-runtime-objc.js
Tyilo: It isn't on the prototype, maybe we can only add them to `enumerate`, but not `keys`?

bindings/gumjs/gumscript-runtime-objc.js
Tyilo: I see, but it is a lot longer to write `ObjC.classes.NSString` instead of just `'NSString'`, so I think we should keep it for ease of use.

bindings/gumjs/gumscript-runtime-objc.js
oleavr: Only `ObjC.nil` will give you the ability to have a chain of calls that still "work", but as you said, it's still quite broken. Maybe we shouldn't bother with `ObjC.nil` and just expect you to pass `NULL` for such arguments?

bindings/gumjs/gumscript-runtime-objc.js
oleavr: Hmm, it just makes me wonder if we should also do the same for `registerClass()`; should you be able to use strings for `super` and `protocols`? In the end my gut feeling is that we should go explicit, as that will encourage users to re-use and get caching benefits of using the same object over and over:

```JavaScript
const NSString = ObjC.classes.NSString;

ObjC.choose(NSString, {...});

Interceptor.attach(NSString.foo.implementation, {...});
```

bindings/gumjs/gumscript-runtime-objc.js
oleavr: Don't we currently have the same problem for String objects?

```JavaScript
var s = "Hello";
undefined
Object.keys(s)
["0", "1", "2", "3", "4"]
Object.getOwnPropertyNames(s)
["0", "1", "2", "3", "4", "length"]
```

oleavr:
I dropped the nil stuff for now, let's give it a bit more thought and maybe reland it in a future PR.

PR #75, opened 2015-06-18
Ensure env.popLocalFrame will be called

marc1006:
It's possible that 'toJni' or 'fromJni' throws an error so we have to ensure that env.popLocalFrame(...) will be called anyway.

### DIFF
--- a/bindings/gumjs/gumscript-runtime-dalvik.js
+++ b/bindings/gumjs/gumscript-runtime-dalvik.js
@@ -762,8 +762,12 @@
                     if (retType.fromJni) {
                         frameCapacity++;
                         returnCapture = "var rawResult = ";
-                        returnStatements = "var result = retType.fromJni.call(this, rawResult, env);" +
-                            "env.popLocalFrame(NULL);" +
+                        returnStatements = "var result;" +
+                            "try {" +
+                                "result = retType.fromJni.call(this, rawResult, env);" +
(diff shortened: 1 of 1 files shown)


bindings/gumjs/gumscript-runtime-dalvik.js
```diff
@@ -993,13 +997,21 @@
                 if (retType.toJni) {
                     frameCapacity++;
                     returnCapture = "var result = ";
+                    returnStatements = "var rawResult;" +
+                        "try {";
                     if (retType.type === 'pointer') {
-                        returnStatements = "var rawResult = retType.toJni.call(this, result, env);" +
+                        returnStatements += "rawResult = retType.toJni.call(this, result, env);" +
+                            "} catch (e) {" +
```
oleavr: Why not use `finally` here?

oleavr:
Great catch (no pun intended!) :+1:

marc1006:
Oh :) Yes you're right.

oleavr:
Oops, no, I was wrong; sorry! You only want that in the failure case, to avoid popping twice. (Would be good to squash these commits while at it.)

marc1006:
Okay I have reset it. And we have in the catch branch
```
env.popLocalFrame(NULL);
```
because we have ```rawResult``` not received.

oleavr:
Thanks!

PR #81, opened 2015-06-28
Implement side-stack for on_enter/on_leave on QNX

s1341:
QNX has very small default stack sizes - just 128kb. This is insufficient
for V8 to run on top of. We've already patched glib and v8 to create new
threads with larger stacks, but that still leaves a problem with pre-existing
threads created by the target process. This patch works around this problem
by implementing a 'side-stack' which all the on_enter/on_leave handlers run
under instead of the original/real stack. This side-stack is sufficiently
large to run v8 code without issues (currently 2Mb).

### DIFF
--- a/gum/guminterceptor.c
+++ b/gum/guminterceptor.c
@@ -67,6 +67,11 @@ struct _InterceptorThreadContext
   GumInvocationStack * stack;
 
   GumArray * listener_data_slots;
+
+#ifdef HAVE_QNX
+  gpointer thread_side_stack;
+  guint thread_side_stack_size;
+#endif
 };
 
 struct _GumInvocationStackEntry
@@ -147,6 +152,11 @@ static gboolean is_patched_function (GumInterceptor * self,
 
 static void gum_function_context_wait_for_idle_trampoline (
     FunctionContext * ctx);
+
+#ifdef HAVE_QNX
+static void _gum_exec_callback_func_with_side_stack (GumInvocationListener * listener_instance,
+    GumInvocationContext * invocation_ctx, gpointer func, gpointer side_stack);
+#endif
 
 static GMutex _gum_interceptor_mutex;
 static GumInterceptor * _the_interceptor = NULL;
@@ -804,8 +814,14 @@ _gum_function_context_on_enter (FunctionContext * function_ctx,
       state.invocation_data = stack_entry->listener_invocation_data[i];
       invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+      _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
+        entry->listener_interface->on_enter,
+        interceptor_ctx->thread_side_stack + interceptor_ctx->thread_side_stack_size - 4);
+#else
       entry->listener_interface->on_enter (entry->listener_instance,
           invocation_ctx);
+#endif
     }
 
 #ifdef G_OS_WIN32
@@ -884,8 +900,14 @@ _gum_function_context_on_leave (FunctionContext * function_ctx,
     state.invocation_data = stack_entry->listener_invocation_data[i];
     invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+    _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
+      entry->listener_interface->on_leave,
+      interceptor_ctx->thread_side_stack + interceptor_ctx->thread_side_stack_size - 4);
+#else
     entry->listener_interface->on_leave (entry->listener_instance,
         invocation_ctx);
+#endif
   }
 
 #ifdef G_OS_WIN32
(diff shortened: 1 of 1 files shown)


oleavr:
Nice! The only problem I see here is that the side stack is shared by all threads. A lock would fix that, however that would create other problems like not being able to `Thread.sleep()` or `recv().wait()` in a callback and still let other threads in. I'd suggest extending the existing per-thread struct with a side stack (for QNX only).

s1341:
SFAIK I extended the InterceptorThreadContext object, so the side-stack should be per thread already.

gum/guminterceptor.c
```diff
@@ -1085,6 +1124,12 @@ interceptor_thread_context_new (void)
   context->listener_data_slots = gum_array_sized_new (FALSE, TRUE,
       sizeof (ListenerDataSlot), GUM_MAX_LISTENERS_PER_FUNCTION);
 
+#ifdef HAVE_QNX
+  context->thread_side_stack_size = 2 * 1024 * 1024;
```
oleavr: Wouldn't it be better to use a preprocessor define here? (As we don't need to know the size once created.)

gum/guminterceptor.c
```diff
@@ -1095,6 +1140,10 @@ interceptor_thread_context_destroy (InterceptorThreadContext * context)
 
   gum_array_free (context->stack, TRUE);
 
+#ifdef HAVE_QNX
+  gum_free (context->thread_side_stack);
```
oleavr: You want to use `gum_free_pages()` here.

gum/guminterceptor.c
```diff
@@ -901,6 +923,23 @@ _gum_function_context_on_leave (FunctionContext * function_ctx,
 #endif
 }
 
+#ifdef HAVE_QNX
+__attribute__ ((naked))
+static void
```
oleavr: The attribute line and `static void` should be put on one line.

gum/guminterceptor.c
```diff
@@ -901,6 +923,23 @@ _gum_function_context_on_leave (FunctionContext * function_ctx,
...
+_gum_exec_callback_func_with_side_stack (GumInvocationListener * listener_instance,
+                               GumInvocationContext * invocation_ctx,
```
oleavr: The arguments should be visually aligned. Also, we only underscore-prefix internal functions that are non-static.

gum/guminterceptor.c
```diff
@@ -884,8 +900,14 @@ _gum_function_context_on_leave (FunctionContext * function_ctx,
     state.invocation_data = stack_entry->listener_invocation_data[i];
     invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+    _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
+      entry->listener_interface->on_leave,
```
oleavr: Continuation indents should be 4 spaces.

gum/guminterceptor.c
```diff
@@ -804,8 +814,14 @@ _gum_function_context_on_enter (FunctionContext * function_ctx,
       state.invocation_data = stack_entry->listener_invocation_data[i];
       invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+      _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
```
oleavr: Continuation indents should be 4 spaces.

gum/guminterceptor.c
```diff
@@ -148,6 +153,11 @@ static gboolean is_patched_function (GumInterceptor * self,
 static void gum_function_context_wait_for_idle_trampoline (
     FunctionContext * ctx);
 
+#ifdef HAVE_QNX
+static void _gum_exec_callback_func_with_side_stack (GumInvocationListener * listener_instance,
```
oleavr: > 80 cols?

gum/guminterceptor.c
```diff
@@ -804,8 +814,14 @@ _gum_function_context_on_enter (FunctionContext * function_ctx,
       state.invocation_data = stack_entry->listener_invocation_data[i];
       invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+      _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
+        entry->listener_interface->on_enter,
+        interceptor_ctx->thread_side_stack + interceptor_ctx->thread_side_stack_size - 4);
```
oleavr: Exceeds 80 cols?

gum/guminterceptor.c
```diff
@@ -884,8 +900,14 @@ _gum_function_context_on_leave (FunctionContext * function_ctx,
     state.invocation_data = stack_entry->listener_invocation_data[i];
     invocation_ctx->backend->data = &state;
 
+#ifdef HAVE_QNX
+    _gum_exec_callback_func_with_side_stack (entry->listener_instance, invocation_ctx,
```
oleavr: Exceeds 80 cols?

gum/guminterceptor.c
oleavr: Exceeds 80 cols?

oleavr:
Oops, sorry, `ENOCAFFEINE` :)

gum/guminterceptor.c
```diff
@@ -32,6 +32,10 @@ G_DEFINE_TYPE (GumInterceptor, gum_interceptor, G_TYPE_OBJECT);
 #define GUM_INTERCEPTOR_LOCK()   (g_mutex_lock (&priv->mutex))
 #define GUM_INTERCEPTOR_UNLOCK() (g_mutex_unlock (&priv->mutex))
 
+#ifdef HAVE_QNX
+#define GUM_THREAD_SIDE_STACK_SIZE  2 * 1024 * 1024
```
oleavr: Minor style discrepancy here, should be something like:

```C
# define GUM_THREAD_SIDE_STACK_SIZE (2 * 1024 * 1024)
```

PR #82, opened 2015-06-28
Some improvements

marc1006:
- 'isCompatible' functions improved
- Using functionnames
- Add 'Env.prototype.javaLangString'

I really don't know why there wasn't 'Env.prototype.javaLangString' already implemented.


oleavr:
Awesome work! Thanks.

PR #88, opened 2015-07-02
Replace 'var' with 'let' or 'const' and some other modifcations

marc1006:
- Replace 'var' with 'let', 'const'
 - Replace function-expressions with function-declarations
 - Add remaining Java modificators
 - Insert missing and remove unnecessary semicolons


oleavr:
:heart:

PR #92, opened 2015-07-14
Several adjustments

marc1006:
- access to instance and static fields
- improve `getTypeName` and `getArrayTypeName`
- check if it's possible to cast
- instances can access static methods (now it's the same behavior as in Java)
- some other fixes


oleavr:
Awesome work! :heart:

PR #94, opened 2015-07-17
`Dalvik.synchronized`, `Dalvik.scheduleOnMainThread` and `Dalvik.isMainThread` implemented

marc1006:
- add `Dalvik.synchronized(obj/objHandle, fn);`. It's almost equivalent to following Java code `synchronized (obj) { fn(); }`
 - add `Dalvik.scheduleOnMainThread(fn);`
 - if we set an implementation to `null`, set it really to `null`
 - add `Dalvik.isMainThread()`. Now we can check if we are in the main thread
 - refactoring: assertions
 - increase GUM_MAX_SCRIPT_SOURCE_CHUNKS to 6

### DIFF
--- a/bindings/gumjs/gumscript-runtime-dalvik.js
+++ b/bindings/gumjs/gumscript-runtime-dalvik.js
@@ -98,63 +98,92 @@
             }
         });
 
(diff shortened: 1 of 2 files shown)


oleavr:
:+1:

PR #101, opened 2015-08-24
Implement method to load classes at runtime

marc1006:
```Java.openDexFile(filePath)``` will return an object with two methods
 - ```load()```: load the dex-file
 - ```getClassNames()```: get the class names

### DIFF
--- a/bindings/gumjs/gumscript-runtime-java.js
+++ b/bindings/gumjs/gumscript-runtime-java.js
@@ -273,8 +273,8 @@
             return classFactory.use(className);
         };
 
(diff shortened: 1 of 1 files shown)


oleavr:
Nice work! My only concern is that we should avoid using the globals, i.e. `Java` and `Dalvik` from inside the implementation, because otherwise it's a lot of work to temporarily change the module name, e.g. for testing purposes.

marc1006:
Done. And thanks for this advice.

oleavr:
:+1:

PR #109, opened 2015-09-03
Fix for array conversion and fix for field setter

marc1006:


marc1006:
We have to implement a function to convert from JNI Signature to Java signature and vice versa. So we can replace  ```name.replace(/;|\[/g, "")``` with it

oleavr:
:+1:

PR #122, opened 2015-12-08
Emulate {Get/Set}LastErrror on Windows

elvanderb:


oleavr:
Some issues:
- In retrospect I think `gum_thread_get_system_error` and `gum_thread_set_system_error` would be names, as `errno` makes one think of something that only exists in the CRT on Windows. It's better to pick something neutral if it's going to map to `LastError` on Windows and `errno` on UNIX. 
- Files that are common to both 32- and 64-bit should always be added to `gum-common.props` instead of both of the `.vcxproj` files.
- Let's add this to gumprocess.h and implement it in gumprocess-windows.c, it's so close to the functionality there that introducing a separate "module" isn't necessary. I'd say add the prototypes right after the gum_process ones.
- Mixture of tabs and spaces in `tests/core/process.c`.
- The test-code doesn't belong in `process.c`.
- Function definitions should have their return type etc. on a separate line.
- Variable names should always be lower-case, even with acronyms, so `teb` not `TEB`.
- Let's get rid of the errno code-path in guminterceptor.c (and include); don't worry about implementing it on UNIX, I'll take care of that once this has landed.

PR #127, opened 2015-12-23
Support linux-arm

s1341:
Note that the workaround for the r0 being set to 0 by signal return is a bit ugly. We still don't know what exactly the root cause of the issue is, but it's possible that we are hitting this line: http://lxr.free-electrons.com/source/arch/arm/kernel/signal.c?v=3.7#L309


tests/testutil.c
```diff
@@ -408,7 +408,7 @@ test_util_get_system_module_name (void)
     char * result;
 
     _test_util_system_module_name = malloc (64);
-    p = popen ("find /lib -name \"libc-*.so\" | head -1 | xargs basename"
+    p = popen ("find /lib -name \"libc[-.]*so*\" | head -1 | xargs basename"
```
oleavr: Did you check whether this still picks the actual libc and not the symlink on a typical LSB system like Ubuntu?

s1341: Seems to work for me on gentoo64 but it appears to fail on ubuntu64... Do you have a suggestion for a better fix?

oleavr: How about running the old/current command first, and if that yields nothing we try your glob second?

s1341: I just tested two different ubuntu64 machines with the original code. On one the x86_64 libc so was first, and on the other the i386 libc so was first. In other words, the original code wasn't perfect either...

Perhaps we should try something a little more precise?

oleavr: Good point! Maybe something like `grep libc /proc/self/maps`?

gum/backend-linux/gumprocess-linux.c
```diff
@@ -1241,7 +1242,8 @@ gum_linux_unparse_ucontext (const GumCpuContext * ctx,
   sc->arm_r7 = ctx->r[7];
   sc->arm_lr = ctx->lr;
 
-  sc->arm_cpsr = (ctx->pc & 1) ? GUM_PSR_THUMB : 0;
+  sc->arm_cpsr = ctx->cpsr;
+  sc->arm_cpsr |= (ctx->pc & 1) ? GUM_PSR_THUMB : 0;
```
oleavr: I suppose we should clear the thumb bit here if not set?

gum/backend-arm/guminterceptor-arm.c
```diff
@@ -542,7 +542,7 @@ gum_interceptor_backend_write_prolog (GumThumbWriter * tw)
   gum_thumb_writer_put_push_regs (tw, 2, ARM_REG_R0, ARM_REG_R1);
 
   /* reserve space for next_hop */
```
oleavr: We should update this comment to avoid confusion.

gum/backend-arm/guminterceptor-arm.c
```diff
@@ -560,7 +560,7 @@ gum_interceptor_backend_write_epilog (GumThumbWriter * tw)
       GUM_FRAME_OFFSET_CPU_CONTEXT + G_STRUCT_OFFSET (GumCpuContext, lr));
 
   /* clear next_hop and low part of GumCpuContext */
```
oleavr: We should update this comment to avoid confusion.

oleavr:
Awesome work!

PR #129, opened 2016-01-04
Add polyfills for Array.from and Array.prototype.includes and fix a bug with NativeFunctions returning void

s1341:


bindings/gumjs/gumdukcore.c
```diff
@@ -1492,26 +1492,27 @@ GUMJS_DEFINE_FUNCTION (gumjs_native_function_invoke)
   else
   {
     duk_push_undefined (ctx);
-    result = _gumjs_get_value (ctx, -1);
-    duk_pop (ctx);
   }
 
-  switch (result->type)
+  if (result)
```
oleavr: We should stick to the style: `result != NULL` (and only use this kind of truth-checks for gbooleans).

oleavr:
Sweet!

PR #136, opened 2016-01-07
Switch to a minimal subset of the babel polyfill.

s1341:
The subset was generated using the following command in the core-js repo:
npm run grunt build:es6.promise,es6.symbol,es6.set,es6.array.from,es6.array.iterator,es7.array.includes,es6.array.find uglify


oleavr:
Sweet!

PR #137, opened 2016-01-10
Add missing libm dependency

s1341:


bindings/gumjs/generate-runtime.py
```diff
@@ -73,7 +73,7 @@ def generate_runtime_duk(output_dir, output, input_dir, inputs, polyfills):
             ]
         else:
             CC = ["gcc"]
-        subprocess.call(CC + ["-Wall", "-pipe", "-O2", "-fomit-frame-pointer"] +
+        subprocess.call(CC + ["-Wall", "-pipe", "-O2", "-fomit-frame-pointer", "-lm"] +
```
oleavr: It should only be added to the CC above; it isn't needed on Mac.

PR #145, opened 2016-02-16
Make frida-gum compile on qnx again

s1341:


gum/gumexceptor.c
```diff
@@ -29,13 +32,18 @@
...
+#endif
+
 typedef struct _GumExceptionHandlerEntry GumExceptionHandlerEntry;
 #if defined (G_OS_WIN32) || defined (HAVE_DARWIN)
 # define GUM_NATIVE_SETJMP setjmp
 # define GUM_NATIVE_LONGJMP longjmp
   typedef jmp_buf GumExceptorNativeJmpBuf;
 #else
-# ifdef sigsetjmp
+# ifdef __sigsetjmp
```
oleavr: This breaks something. The check here was intentional, because we want to detect the sigsetjmp macro and use the actual function name.

gum/gumexceptor.c
```diff
@@ -29,13 +32,18 @@
 # include <signal.h>
 #endif
 
+#ifdef HAVE_QNX
```
oleavr: These should be in the same group, i.e. no empty line.

gum/gumexceptor.c
s1341: any suggestions for how to implement this? There doesn't appear to be a __sigsetjmp on qnx

oleavr: I'd suggest just adding it right below the Windows/Darwin block:
```
#elif defined (HAVE_QNX)
  define GUM_NATIVE_SETJMP sigsetjmp
```

s1341: Ok. I'll fix that tomorrow.

oleavr:
:+1:

PR #148, opened 2016-02-24
More QNX integration patches

s1341:


tests/core/symbolutil.c
```diff
@@ -114,13 +114,13 @@ SYMUTIL_TESTCASE (find_functions_matching)
...
 
-static void GUM_CDECL
+void GUM_CDECL
 gum_dummy_function_0 (void)
 {
   g_print ("%s\n", G_STRFUNC);
 }
 
-static void GUM_STDCALL
+void GUM_STDCALL
```
oleavr: This breaks the tests above checking if we can find static functions. Is the problem here that they get inlined? We have a macro to override that, GUM_NOINLINE, e.g.:
https://github.com/frida/frida-gum/blob/4286fa106d4bd752c0af5d39c5f46cf9c3ea73d3/tests/core/interceptor.c#L91

tests/gumjs/script.c
```diff
@@ -32,7 +32,9 @@ TEST_LIST_BEGIN (script)
   SCRIPT_TESTENTRY (invocations_are_bound_on_tls_object)
   SCRIPT_TESTENTRY (invocations_provide_thread_id)
   SCRIPT_TESTENTRY (invocations_provide_call_depth)
+#if !defined (HAVE_QNX)
```
oleavr: Let's use #ifndef for consistency with the other changes in this commit.

oleavr:
Sweet!

PR #151, opened 2016-02-29
Change assertion to if, so that we silently handle some pathologicall…

s1341:
…y broken files when trying to enumerate exports


oleavr:
:+1:

PR #152, opened 2016-03-07
Don't use cbnz which is not available in arm mode on some processors

s1341:


oleavr:
Do these processors support `ldr pc, [pc, #x]` in Thumb mode?

gum/backend-linux/gumprocess-linux.c
```diff
@@ -1861,7 +1861,8 @@ gum_libc_clone (GumCloneFunc child_func,
       "mov r7, %[clone_syscall]\n\t"
       "ldmdb %[next_args]!, {r0, r1, r2, r3, r4}\n\t"
       "swi 0x0\n\t"
-      "cbnz r0, 1f\n\t"
+      "cmp r0, #0\n\t"
+      "bne 1f\n\t"
```
oleavr: Should also add `cc` to the clobber list (in the same order as other clobber lists where we include `cc`).

s1341:
oleavr: don't know. I'd have to try it.

PR #156, opened 2016-03-31
Fix missing include on QNX

s1341:


tests/gumjs/script.c
```diff
@@ -7,6 +7,10 @@
 
 #include "script-fixture.c"
 
+#ifdef HAVE_QNX
+#include <unix.h>
+#endif
+
```
oleavr: Would be good to use `# include <unix.h>` for consistency, and move this to the other includes in https://github.com/frida/frida-gum/blob/6c7a2d46c9b279f0aedadd452736535e9d3c574a/tests/gumjs/script-fixture.c#L14-L32

oleavr:
:+1:

PR #174, opened 2016-10-16
Do not recycle the InvocationReturnValue object

oleavr:
It's counter-intuitive that `onLeave` implementors need to deep-copy the
value to use it outside the callback.


oleavr:
Decided against this as even this rather optimized approach results in a severe performance regression.

PR #182, opened 2017-01-11
Accommodate sourcemap changes in frida-compile/pull/2

riverar:
Depends on: https://github.com/frida/frida-compile/pull/2

frida-compile now, by default, encodes source locations relative to where its invoked, not where the scripts live. This ensures that a debugger can correctly resolve paths using the frida-compile invocation site as the root directory.

So with that, `../node_modules` handling is no longer required. For an input script `runtime\foo.js` that calls `require ('promiscuous')`, frida-compile will resolve to `node_modules/promiscuous`. The path then, as before, gets appended to a pseudo-fs-root `frida/`. To keep things tidy (and reduce debugger confusion), the path also has its slashes tweaked to fit canonical form.

### DIFF
--- a/bindings/gumjs/generate-runtime.py
+++ b/bindings/gumjs/generate-runtime.py
@@ -155,10 +155,7 @@ def extract_source_map(filename, source_code):
     return (stripped_source_code, raw_source_map)
 
 def to_canonical_source_path(path):
-    if path.startswith('../node_modules'):
-        return 'frida/' + path[3:]
-    else:
-        return 'frida/' + path
+    return os.path.join("frida", path).replace("\\", "/")
 
 def write_bytes(data, sink):
     sink.write("\n  ")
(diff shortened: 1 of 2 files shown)


bindings/gumjs/generate-runtime.py
```diff
@@ -155,10 +155,7 @@ def extract_source_map(filename, source_code):
     return (stripped_source_code, raw_source_map)
 
 def to_canonical_source_path(path):
-    if path.startswith('../node_modules'):
-        return 'frida/' + path[3:]
-    else:
-        return 'frida/' + path
+    return os.path.join('frida', path).replace("\\","/")
```
oleavr: Missing space after comma, and, this one is my fault: let's use double-quoted strings here – I've been mixing styles in this script. (The original style was to use double-quoted strings for strings, and single-quoted strings for pretend enum/symbol values.)

riverar:
Changes made, thanks!

PR #198, opened 2017-03-07
Expose Profiler APIs to the JS bindings [WIP]

oleavr:


oleavr:
Might revisit this later; will keep the branch around.

PR #211, opened 2017-04-12
Fixed illegal instruction for physical devices

akiannillo:
Tested on Samsung S6 Edge Android API 24.
Executing a load pair instruction that use the same register for both register operands results in Illegal Instruction error.
Using a second register (X15) as second register operand, even if not used later in the code, fixes it.


oleavr:
👌🍻

Issue #195, opened 2017-02-24
SIGILL on 64 bit Stalker/Android


manizzle:
When trying to run Stalker on an application that loads up a 64 bit library, I get the following errors:

```
02-21 15:46:29.758 23245-23432/com.demos.app A/libc: Fatal signal 4 (SIGILL), code 1, fault addr 0x7f5d13e044 in tid 23432 (gum-js-loop)
02-21 15:46:29.853 22891-22891/? I/ServiceManager: Waiting for service AtCmdFwd...
02-21 15:46:29.862 607-607/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
02-21 15:46:29.862 607-607/? A/DEBUG: Build fingerprint: 'google/angler/angler:6.0/MDA89D/2294819:user/release-keys'
02-21 15:46:29.862 607-607/? A/DEBUG: Revision: '0'
02-21 15:46:29.862 607-607/? A/DEBUG: ABI: 'arm64'
02-21 15:46:29.863 607-607/? A/DEBUG: pid: 23245, tid: 23432, name: gum-js-loop  >>> com.demos.app <<<
02-21 15:46:29.863 607-607/? A/DEBUG: signal 4 (SIGILL), code 1 (ILL_ILLOPC), fault addr 0x7f5d13e808
02-21 15:46:29.879 607-607/? A/DEBUG:     x0   0000007f96ac8980  x1   0000000000000050  x2   0000000000000230  x3   0000007f9b398000
02-21 15:46:29.881 607-607/? A/DEBUG:     x4   0000007f9b398210  x5   0000007f9b398440  x6   0000000000000001  x7   0000007f7dded4f0
02-21 15:46:29.881 607-607/? A/DEBUG:     x8   0000007f7dded440  x9   0000007f7dded670  x10  0000000000000001  x11  00000000000004e1
02-21 15:46:29.881 607-607/? A/DEBUG:     x12  0000007f5d13d000  x13  ffffffffa7536631  x14  0000007f7eab716c  x15  0000007f7eac76b8
02-21 15:46:29.881 607-607/? A/DEBUG:     x16  0000007f7f0e4958  x17  0000007f9b320760  x18  00000000b5000051  x19  0000007f96ac8980
02-21 15:46:29.882 607-607/? A/DEBUG:     x20  0000007f7dded268  x21  0000007f96af2a98  x22  0000000000000000  x23  0000007f7eae1378
[...]

PR #215, opened 2017-04-19
Added clear cache code

akiannillo:
Unlike emulator, actual smartphones with actual ARM CPUs use an instruction cache that needs to be cleared every time frida dynamically writes code. Not clearing instruction cache produce intermittent and random errors (e.g., SIGILL, SIGSEGV) while executing stalker tests (errors are accelerated by using the smartphone meanwhile). The clear cache code fit perfectly in the flush function of the writer, and the range to clear is rounded to the page(s) size.

This should solve issues #196 and #195 ( I will contact @manizzle and let him test it ) 

Reading: [1](https://community.arm.com/processors/b/blog/posts/caches-and-self-modifying-code) [2](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)

Thanks to @rnatella for the chat on this issue ;)


oleavr:
Ah yes, I totally missed this when reviewing the Stalker arm64 PR. This does not belong in Arm64Writer, though, due to the current separation of concerns. A codewriter generates code that may or may not be executed in the current process, so cache invalidation should not happen there. We currently do `gum_clear_cache()` in `Interceptor`, and Stalker should do the same.

gum/backend-arm64/gumstalker-arm64.c
```diff
@@ -1153,6 +1153,8 @@ gum_exec_ctx_obtain_block_for (GumExecCtx * ctx,
 
   block->code_end = (guint8 *) gum_arm64_writer_cur (cw);
 
+  gum_clear_cache (block->code_begin, block->code_end - block->code_begin);
```
oleavr: How about doing this at the end of `gum_exec_block_commit()` instead?

akiannillo: ok 👍

PR #216, opened 2017-05-06
	adds the support of heap allocs enumeration for windows

elvanderb:


gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
   }
 }
 
+static gboolean
+gum_process_enumerate_heap_ranges (HANDLE heap, GumFoundMallocRangeFunc func,
```
oleavr: One argument per line.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+                                   gpointer user_data)
+{
+  PROCESS_HEAP_ENTRY entry;
+  GumMallocRangeDetails details;
+  GumMemoryRange gum_range;
+  gboolean carry_on, unlock_heap;
+
+  // HeapLock may fail but it doesn't seem to have any real consequences...
```
oleavr: C-style comments only.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  unlock_heap = HeapLock (heap);
+
+  details.range = &gum_range;
+  carry_on = TRUE;
+  entry.lpData = NULL;
+  while (carry_on && HeapWalk (heap, &entry)) {
```
oleavr: `{` on the wrong line.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    if ((entry.wFlags & PROCESS_HEAP_ENTRY_BUSY) != 0) {
```
oleavr: `{` on the wrong line.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+      gum_range.base_address = entry.lpData;
+      gum_range.size = entry.cbData;
+      carry_on = func (&details, user_data);
+    }
+  }
+  if (unlock_heap)
```
oleavr: Let's add a blank line before this `if` so it reads better. (Structure of this function is: lock, do stuff, unlock. Three logical groups, which should have a blank line between each of them.)

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+}
+
 void
 gum_process_enumerate_malloc_ranges (GumFoundMallocRangeFunc func,
                                      gpointer user_data)
 {
-  /* Not implemented */
-  (void) func;
-  (void) user_data;
+  DWORD nb_heaps, i;
```
oleavr: Let's rename this `num_heaps` for consistency. Also, variables should be declared ordered by their chronological use. (The only exception is if the function has a variable holding the final result – which should always be first.)

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  HANDLE process_heap;
+  HANDLE * heaps;
+
+  process_heap = GetProcessHeap ();
+  if (!gum_process_enumerate_heap_ranges (process_heap, func, user_data))
+    return;
+
+  while (TRUE) {
```
oleavr: `{` on the wrong line.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    new_nb_heaps = GetProcessHeaps (0, NULL);
+    if (!new_nb_heaps) {
+      HeapFree (process_heap, 0, heaps);
+      return;
+    }
+    if (new_nb_heaps == nb_heaps)
+      break;
+    HeapFree (process_heap, 0, heaps);
+  }
+  for (i = 0; i < nb_heaps; i++)
```
oleavr: Should always use a block if the statement spans multiple lines.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    if (heaps[i] != process_heap)
```
oleavr: Should always use a block if the statement spans multiple lines.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  DWORD nb_heaps, i;
+  HANDLE process_heap;
+  HANDLE * heaps;
+
+  process_heap = GetProcessHeap ();
+  if (!gum_process_enumerate_heap_ranges (process_heap, func, user_data))
+    return;
+
+  while (TRUE) {
+    DWORD new_nb_heaps;
```
oleavr: Missing blank line after variable declarations.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    nb_heaps = GetProcessHeaps (0, NULL);
+    if (!nb_heaps)
+      return;
+    heaps = HeapAlloc (process_heap, 0, nb_heaps * sizeof(HANDLE));
```
oleavr: Missing space after `sizeof`.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    if (!heaps)
+      return;
+    new_nb_heaps = GetProcessHeaps (0, NULL);
+    if (!new_nb_heaps) {
```
oleavr: `{` on the wrong line.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+      HeapFree (process_heap, 0, heaps);
+      return;
+    }
+    if (new_nb_heaps == nb_heaps)
+      break;
+    HeapFree (process_heap, 0, heaps);
+  }
```
oleavr: Let's add a blank line here; the two loops represent different logical operations.

tests/core/process.c
```diff
@@ -352,7 +355,6 @@ PROCESS_TESTCASE (process_malloc_ranges)
 
   {
     TestForEachContext ctx;
-
```
oleavr: This blank line was intentional and should be kept.

tests/core/process.c
```diff
@@ -340,8 +342,9 @@ PROCESS_TESTCASE (process_ranges_exclude_cloaked)
   g_assert (!ctx.found);
 }
 
-#ifdef HAVE_DARWIN
+#if defined (G_OS_WIN32) || defined (HAVE_DARWIN)
 
+#define STACK_BUFF_SIZE 50
```
oleavr: Let's rename this `TEST_STACK_BUFFER_SIZE`.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,70 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  for (i = 0; i < nb_heaps; i++)
+    if (heaps[i] != process_heap)
+      if (!gum_process_enumerate_heap_ranges (process_heap, func,
+        user_data))
+        break;
+  HeapFree (process_heap, 0, heaps);
```
oleavr: Let's add a blank line on each side of this statement.

elvanderb:
done :)

bindings/gumjs/gumdukprocess.c
```diff
@@ -357,7 +357,7 @@ gum_emit_range (const GumRangeDetails * details,
   return proceed;
 }
 
-#ifdef HAVE_DARWIN
+#if defined (HAVE_DARWIN) || defined (G_OS_WIN32)
```
oleavr: For consistency we should swap the order, Windows then Darwin.

bindings/gumjs/gumv8process.cpp
```diff
@@ -373,7 +373,7 @@ gum_emit_range (const GumRangeDetails * details,
  * TBW
  */
 
-#ifdef HAVE_DARWIN
+#if defined (HAVE_DARWIN) || defined (G_OS_WIN32)
```
oleavr: For consistency we should swap the order, Windows then Darwin.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+                                   gpointer user_data)
+{
+  gboolean carry_on;
+  gboolean unlock_heap;
+  GumMemoryRange gum_range;
+  GumMallocRangeDetails details;
+  PROCESS_HEAP_ENTRY entry;
+
+  /* HeapLock may fail but it doesn't seem to have any real consequences... */
+  unlock_heap = HeapLock (heap);
```
oleavr: Naming this `locked_heap` would be a bit clearer.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
 }
 
+static gboolean
+gum_process_enumerate_heap_ranges (HANDLE heap,
+                                   GumFoundMallocRangeFunc func,
+                                   gpointer user_data)
+{
+  gboolean carry_on;
+  gboolean unlock_heap;
+  GumMemoryRange gum_range;
```
oleavr: How about just `range`?

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  unlock_heap = HeapLock (heap);
+
+  details.range = &gum_range;
+  carry_on = TRUE;
+  entry.lpData = NULL;
+  while (carry_on && HeapWalk (heap, &entry))
+  {
+    if ((entry.wFlags & PROCESS_HEAP_ENTRY_BUSY) != 0)
+    {
+      gum_range.base_address = (GumAddress)entry.lpData;
```
oleavr: Missing space after cast.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+      break;
+    HeapFree (process_heap, 0, heaps);
+  }
+
+  for (i = 0; i < num_heaps; i++)
+  {
+    if (heaps[i] != process_heap)
+    {
+      if (!gum_process_enumerate_heap_ranges (process_heap, func,
+        user_data))
```
oleavr: Line broken prematurely (fits in 80 cols).

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+
+  while (TRUE)
+  {
+    DWORD new_num_heaps;
+
+    num_heaps = GetProcessHeaps (0, NULL);
+    if (!num_heaps)
+      return;
+    heaps = HeapAlloc (process_heap, 0, num_heaps * sizeof (HANDLE));
+    if (!heaps)
```
oleavr: Only booleans should be checked this way – `if (heaps != NULL)` is the style we're using elsewhere (always prefer explicit).

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+      return;
+    new_num_heaps = GetProcessHeaps (0, NULL);
+    if (!new_num_heaps)
```
oleavr: Only booleans should be checked this way – `if (new_num_heaps == 0)` is the style we're using elsewhere (always prefer explicit).

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,86 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    return;
+
+  while (TRUE)
+  {
+    DWORD new_num_heaps;
+
+    num_heaps = GetProcessHeaps (0, NULL);
+    if (!num_heaps)
+      return;
+    heaps = HeapAlloc (process_heap, 0, num_heaps * sizeof (HANDLE));
```
oleavr: This logic doesn't actually end up getting the heaps, so presumably this is broken? Also I don't think there's any point in retrying in case the number changes, especially since we give it the size of the buffer, so we just need to use the number actually written (in case it is lower). The reason I don't think it's worth retrying is that the number may also change while we're iterating, and in that case we can't easily restart and skip the entries we already enumerated. (That would be trying too hard IMO.)

elvanderb: The check is here just in case the number change between the 2 calls, it shouldn't be a problem. It may also change after but we cannot do lot to prevent this.

elvanderb: but okay (and yes, the code was broken :))

elvanderb:
aaaaand done (I hope)

tests/core/process.c
```diff
@@ -340,8 +342,9 @@ PROCESS_TESTCASE (process_ranges_exclude_cloaked)
   g_assert (!ctx.found);
 }
 
-#ifdef HAVE_DARWIN
+#if defined (G_OS_WIN32) || defined (HAVE_DARWIN)
 
+#define TEST_STACK_BUFFER_SIZE 50
```
oleavr: Let's add a blank line after this define. Defines shouldn't be stuck to functions following them.

tests/core/process.c
```diff
@@ -392,15 +394,15 @@ PROCESS_TESTCASE (process_malloc_ranges)
...
+    ctx.range.size = TEST_STACK_BUFFER_SIZE;
     ctx.found = FALSE;
     ctx.found_exact = FALSE;
     gum_process_enumerate_malloc_ranges (malloc_range_check_cb, &ctx);
     g_assert (!ctx.found);
     g_assert (!ctx.found_exact);
   }
 }
-
+#undef TEST_STACK_BUFFER_SIZE
```
oleavr: Let's add a blank line after this `#undef` to avoid gluing it to the `#endif`, making it look related to it.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,77 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  if (heaps == NULL)
+    return;
+  num_heaps = GetProcessHeaps (num_heaps, heaps);
+  if (num_heaps == 0)
+  {
+    HeapFree (process_heap, 0, heaps);
+    return;
+  }
+
+  for (i = 0; i < num_heaps; i++)
```
oleavr: Always use `!=` in loops. (The idea is to turn off-by-one-errors into loops that never terminate, which are easier to notice than a weird off-by-one bug.)

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,77 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+    return;
+
+  num_heaps = GetProcessHeaps (0, NULL);
+  if (num_heaps == 0)
+    return;
+  heaps = HeapAlloc (process_heap, 0, num_heaps * sizeof (HANDLE));
+  if (heaps == NULL)
+    return;
+  num_heaps = GetProcessHeaps (num_heaps, heaps);
+  if (num_heaps == 0)
```
oleavr: Let's simplify this:

```
  if (num_heaps == 0)
    goto beach;
```

and just add a `beach:` label right above the `HeapFree` below.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,77 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
...
+  locked_heap = HeapLock (heap);
+
+  details.range = &range;
+  carry_on = TRUE;
+  entry.lpData = NULL;
+  while (carry_on && HeapWalk (heap, &entry))
+  {
+    if ((entry.wFlags & PROCESS_HEAP_ENTRY_BUSY) != 0)
+    {
+      range.base_address = (GumAddress) entry.lpData;
```
oleavr: Use `GUM_ADDRESS (entry.lpData)` for casting.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -305,13 +305,77 @@ _gum_process_enumerate_ranges (GumPageProtection prot,
   }
 }
 
+static gboolean
+gum_process_enumerate_heap_ranges (HANDLE heap,
+                                   GumFoundMallocRangeFunc func,
+                                   gpointer user_data)
```
oleavr: This function is lower level than the function below that's using it, so it should be right below it instead. (And forward-declaration should be at the top of the file, ordered chronologically relative to the other forward declarations.)

elvanderb:
I think we are good now :)

gum/backend-windows/gumprocess-windows.c
```diff
@@ -309,9 +311,71 @@ void
...
+    {
+      if (!gum_process_enumerate_heap_ranges (process_heap, func, user_data))
+        break;
+    }
+  }
+
+beach:
+  HeapFree (process_heap, 0, heaps);
+
+  return;
```
oleavr: This `return` is redundant. (So we can then also remove the blank line above it.)

gum/backend-windows/gumprocess-windows.c
```diff
@@ -309,9 +311,71 @@ void
...
+
+  num_heaps = GetProcessHeaps (0, NULL);
+  if (num_heaps == 0)
+    return;
+  heaps = HeapAlloc (process_heap, 0, num_heaps * sizeof (HANDLE));
+  if (heaps == NULL)
+    return;
+  num_heaps = GetProcessHeaps (num_heaps, heaps);
+  if (num_heaps == 0)
+    goto beach;
```
oleavr: Wait, we can drop this `if` statement, the loop already checks the same condition. That also eliminates the `beach` label. Sorry I didn't spot this earlier.

elvanderb:
:musical_note: Baby baby one more time :musical_note:

elvanderb:
Actually,  because we remove the loop, I think there is now a bug in my code. Please don't merge now, I'll investigate tonight.

gum/backend-windows/gumprocess-windows.c
```diff
@@ -309,9 +311,66 @@ void
...
+  if (!gum_process_enumerate_heap_ranges (process_heap, func, user_data))
+    return;
+
+  num_heaps = GetProcessHeaps (0, NULL);
+  if (num_heaps == 0)
+    return;
+  heaps = HeapAlloc (process_heap, 0, num_heaps * sizeof (HANDLE));
+  if (heaps == NULL)
+    return;
+  num_heaps = GetProcessHeaps (num_heaps, heaps);
```
elvanderb: returned num_heaps may be greater than the input num_heaps resulting in an out of band read in the heap and possibly a crash.

oleavr: Ah yes good catch! The return value here should be assigned to e.g. `num_heaps_after`, and then `num_heaps = MIN (num_heaps, num_heaps_after)` or something like that.

elvanderb:
done :)

oleavr:
Thanks! 💥

PR #219, opened 2017-05-16
Remove check for address < page_size in gum_darwin_read

mrmacete:


oleavr:
Sweet! Maybe we should also nuke it in `write()` while at it?

mrmacete:
done, pushed --force

PR #225, opened 2017-07-24
Add block and compile events to Stalker

mrmacete:
- `compile` events are emitted as soon as a basic block gets compiled by the Stalker. This happens once per block, so it's useful to understand which parts of code gets executed but it's not a full live trace of the execution.
- `block` events are emitted each time a basic block is executed, so can be used to actually trace the execution.


gum/backend-x86/gumstalker-x86.c
```diff
@@ -2432,6 +2450,32 @@ gum_exec_block_write_exec_event_code (GumExecBlock * block,
 }
 
 static void
+gum_exec_block_write_block_event_code (GumExecBlock * block,
+                                      GumGeneratorContext * gc,
+                                      GumCodeContext cc)
```
oleavr: Missing one space indentation.

PR #228, opened 2017-08-09
Add OutputDebugString() to debug() strings on Windows

erik-smit:
Currently `debug()` strings go to stdout/stderr on Windows, which means they're usually not visible.

### DIFF
--- a/gum/gum.c
+++ b/gum/gum.c
@@ -390,13 +390,13 @@ gum_on_log_message (const gchar * log_domain,
                     const gchar * message,
                     gpointer user_data)
 {
-#if defined(HAVE_WINDOWS)
+#if defined (HAVE_WINDOWS)
   gunichar2 * message_utf16;
 
   message_utf16 = g_utf8_to_utf16 (message, -1, NULL, NULL, NULL);
   OutputDebugString (message_utf16);
   g_free (message_utf16);
-#elif defined(HAVE_ANDROID)
+#elif defined (HAVE_ANDROID)
   int priority;
 
   (void) user_data;
@@ -420,7 +420,8 @@ gum_on_log_message (const gchar * log_domain,
   }
 
   __android_log_write (priority, log_domain, message);
-#elif defined(HAVE_DARWIN)
+#else
+# ifdef HAVE_DARWIN
   static gsize api_value = 0;
   GumCFApi * api;
 
@@ -516,7 +517,7 @@ gum_on_log_message (const gchar * log_domain,
   }
 
   /* else: fall through to stdout/stderr logging */
-#else
+# endif
   FILE * file = NULL;
   const gchar * severity = NULL;
 
--- a/gum/gum.c
+++ b/gum/gum.c
@@ -515,9 +515,9 @@ gum_on_log_message (const gchar * log_domain,
 
     return;
   }
-
   /* else: fall through to stdout/stderr logging */
 # endif
+
   FILE * file = NULL;
   const gchar * severity = NULL;
 
--- a/gum/gum.c
+++ b/gum/gum.c
@@ -16,6 +16,9 @@
 #include <ffi.h>
 #include <glib-object.h>
 #include <gio/gio.h>
+#ifdef HAVE_WINDOWS
+  #include <windows.h>
+#endif
 
 #define DEBUG_HEAP_LEAKS 0
 
@@ -506,6 +509,18 @@ gum_on_log_message (const gchar * log_domain,
 
     return;
   }
+#else
+# ifdef HAVE_WINDOWS
+  gunichar2* utf16_message;
+  utf16_message = g_utf8_to_utf16 (message, -1, NULL, NULL, NULL);
+  if (utf16_message)
+  {
+    OutputDebugString (utf16_message);
+    g_free (utf16_message);
+
+    return;
+  }
+
   /* else: fall through to stdout/stderr logging */
 # endif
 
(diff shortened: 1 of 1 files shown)


gum/gum.c
```diff
@@ -16,6 +16,9 @@
 #include <ffi.h>
 #include <glib-object.h>
 #include <gio/gio.h>
+#ifdef HAVE_WINDOWS
+  #include <windows.h>
```
oleavr: We currently indent includes like this:
```C
#ifdef HAVE_WINDOWS
# include <windows.h>
#endif
```

gum/gum.c
```diff
@@ -506,6 +509,18 @@ gum_on_log_message (const gchar * log_domain,
 
     return;
   }
+#else
```
oleavr: I'd say the stdio fallback doesn't make any sense on Windows – it only does on i/macOS where some processes may not have `CoreFoundation` loaded. So we should move this to the top and do like:

```C
#if defined (HAVE_WINDOWS)
...
#elif defined (HAVE_ANDROID)
...
#else
# ifdef HAVE_DARWIN
```

gum/gum.c
```diff
@@ -506,6 +509,18 @@ gum_on_log_message (const gchar * log_domain,
...
+# ifdef HAVE_WINDOWS
+  gunichar2* utf16_message;
```
oleavr: Missing space before `*`.

gum/gum.c
```diff
@@ -506,6 +509,18 @@ gum_on_log_message (const gchar * log_domain,
...
+  utf16_message = g_utf8_to_utf16 (message, -1, NULL, NULL, NULL);
```
oleavr: Missing blank line before this one.

oleavr: For consistency if would be good to name this `message_utf16`.

gum/gum.c
```diff
@@ -506,6 +509,18 @@ gum_on_log_message (const gchar * log_domain,
...
+  if (utf16_message)
```
oleavr: No need to check the result, we are in control of the call-sites and can assume they give us valid UTF-8.

gum/gum.c
erik-smit: It wouldn't make sense to have it output to console just in case people happen to be frida-ing a console-application?

I guess it makes sense to only have 1 definite path for those messages to come out.
Duplication is confusing.

erik-smit: Why sometimes the `#if defined` and sometimes the `# ifdef`?
And why sometimes the `#elif` and sometimes the `#else # ifdef`?

gum/gum.c
```diff
@@ -16,6 +16,9 @@
 #include <ffi.h>
 #include <glib-object.h>
 #include <gio/gio.h>
+#ifdef HAVE_WINDOWS
+#  include <windows.h>
```
oleavr: Should only be one space per level here.

gum/gum.c
```diff
@@ -387,7 +390,14 @@ gum_on_log_message (const gchar * log_domain,
                     const gchar * message,
                     gpointer user_data)
 {
-#ifdef HAVE_ANDROID
+#if defined(HAVE_WINDOWS)
+  gunichar2 * message_utf16;
```
oleavr: Missing blank line between declaration and first statement.

gum/gum.c
```diff
@@ -387,7 +390,14 @@ gum_on_log_message (const gchar * log_domain,
...
+  message_utf16 = g_utf8_to_utf16 (message, -1, NULL, NULL, NULL);
+  OutputDebugString (message_utf16);
+  g_free (message_utf16);
+
+  return;
```
oleavr: This `return` and previous line should be omitted.

gum/gum.c
erik-smit: fixed, but github doesn't 'see' it, because I've just added a blank line.

gum/gum.c
```diff
@@ -506,9 +514,9 @@ gum_on_log_message (const gchar * log_domain,
 
     return;
   }
-  /* else: fall through to stdout/stderr logging */
-# endif
 
+  /* else: fall through to stdout/stderr logging */
+#else
```
oleavr: This should be reverted as it breaks the fallback on i/macOS.

gum/gum.c
```diff
@@ -411,8 +420,7 @@ gum_on_log_message (const gchar * log_domain,
   }
 
   __android_log_write (priority, log_domain, message);
-#else
-# ifdef HAVE_DARWIN
+#elif defined(HAVE_DARWIN)
```
oleavr: This should be reverted as it breaks the fallback on i/macOS.

gum/gum.c
```diff
@@ -387,7 +390,13 @@ gum_on_log_message (const gchar * log_domain,
                     const gchar * message,
                     gpointer user_data)
 {
-#ifdef HAVE_ANDROID
+#if defined(HAVE_WINDOWS)
```
oleavr: Missing space after `defined`.

gum/gum.c
```diff
@@ -387,7 +390,13 @@ gum_on_log_message (const gchar * log_domain,
...
+  gunichar2 * message_utf16;
+
+  message_utf16 = g_utf8_to_utf16 (message, -1, NULL, NULL, NULL);
+  OutputDebugString (message_utf16);
+  g_free (message_utf16);
+#elif defined(HAVE_ANDROID)
```
oleavr: Missing space after `defined`.

gum/gum.c
```diff
@@ -506,9 +515,9 @@ gum_on_log_message (const gchar * log_domain,
 
     return;
   }
+
   /* else: fall through to stdout/stderr logging */
 # endif
-
```
oleavr: This change is unrelated and should be reverted.

oleavr:
Thanks! ❤️

PR #230, opened 2017-08-20
Add group parameter to `hexdump()`

erik-smit:


bindings/gumjs/runtime/hexdump.js
```diff
@@ -67,25 +77,29 @@ function hexdump(target, options) {
...
-        resetColor
-      );
-
-      asciiChars.push(
-        isNewline ? newlineColor : dataColor,
-        (value >= 32 && value <= 126) ? String.fromCharCode(value) : '.',
-        resetColor
-      );
+      for (let groupOffset = printGroup-1; groupOffset >= 0; groupOffset--) {
+        if (bytes[offset + groupOffset] !== void 0) {
```
oleavr: Let's use `undefined` instead of `void 0`.

erik-smit:
> Missing spaces around operators in many places.

Sorry, I don't see it. Can you give an example?

> Endian should be taken into account. Feel free to add this as Process.endian, which should be either 'little' or 'big', i.e. strings.

Let it automatically decide whether or not to do the endian-conversion based on the contents of `Process.endian`?

oleavr:
> Sorry, I don't see it. Can you give an example?

E.g.:
`printGroup-1` => `printGroup - 1`
`(index+groupOffset)` => `(index + groupOffset)`

> Let it automatically decide whether or not to do the endian-conversion based on the contents of Process.endian?

Yep. I think it would make sense to write a test for this, using `Memory.alloc()` and `Memory.writeU32()` to set up a buffer with one value, and then `hexdump()` that.

erik-smit:
> Yep. I think it would make sense to write a test for this, using Memory.alloc() and Memory.writeU32() to set up a buffer with one value, and then hexdump() that.

And... have the test also be endian-aware?

Or maybe... have the endian->output conversion be a parameter, with the default depending on the systems endianness?

And then having the test do two tests, one with endian set to convert and other with not-convert?

oleavr:
> And... have the test also be endian-aware?

I suppose the test could overwrite `Process.endian` before calling `hexdump()`, but definitely not give it as a parameter, that doesn't sound like a useful thing to do.

oleavr:
Closing for now. Feel free to reopen and continue at a later point.

PR #231, opened 2017-08-25
Do not ignore interception of NativeFunction calls

fabianfreyer:
Interception is ignored for frida's own threads. However, this does not
make sense when calling NativeFunction calls.

To make `gum_interceptor_unignore_current_thread` work before
`gum_interceptor_ignore_current_thread` for theads that aren't ignored
yet, also invoke listeners for negative ignore_levels.

From IRC (these are mostly cited freely from @oleavr ):
- [x] The Interceptor instance should be retrieved from `core`’s `interceptor` field, which has the interceptor impl. as a field (instead of calling `obtain()` and leaking the reference)
- [x] Ignoring again should also happen when the call throws a native exception that we handle
- [x] Style is off
- [x] the test-harness doesn’t ignore Frida’s JS thread. We’ll have to schedule an idle callback on the JS thread to ignore, then run the script, then undo the ignore by scheduling again (so other tests aren’t affected)

### DIFF
--- a/bindings/gumjs/gumdukcore.c
+++ b/bindings/gumjs/gumdukcore.c
@@ -2742,7 +2742,9 @@ gum_duk_native_function_invoke (GumDukNativeFunction * self,
 
     if (gum_exceptor_try (core->exceptor, &exceptor_scope))
     {
+      gum_interceptor_unignore_current_thread (core->interceptor);
       ffi_call (&self->cif, implementation, rvalue, avalue);
+      gum_interceptor_ignore_current_thread (core->interceptor);
 
       if (self->enable_detailed_return)
         system_error = gum_thread_get_system_error ();
@@ -2753,7 +2755,9 @@ gum_duk_native_function_invoke (GumDukNativeFunction * self,
 
   if (gum_exceptor_catch (core->exceptor, &exceptor_scope))
   {
+    gum_interceptor_ignore_current_thread (core->interceptor);
     _gum_duk_throw_native (ctx, &exceptor_scope.exception, core);
+    gum_interceptor_unignore_current_thread (core->interceptor);
   }
 
   if (self->enable_detailed_return)
--- a/bindings/gumjs/gumv8core.cpp
+++ b/bindings/gumjs/gumv8core.cpp
@@ -2435,12 +2435,16 @@ gum_v8_native_function_invoke (GumV8NativeFunction * self,
 
   isolate->Exit ();
 
+  auto interceptor = core->script->priv->interceptor.interceptor;
+
   {
     Unlocker ul (isolate);
 
     if (gum_exceptor_try (core->exceptor, &scope))
     {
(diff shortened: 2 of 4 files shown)


fabianfreyer:
I've updated the patch to address most of the issues discussed, but I'm still not really sure about scheduling jobs on the JS thread. My naïve approach would be something along these lines:
```c
SCRIPT_TESTCASE (native_function_can_be_intercepted)
{
  GumScriptScheduler * scheduler = GUM_DUK_IS_SCRIPT_BACKEND (fixture->backend) ? 
    gum_duk_script_backend_get_scheduler (fixture->backend) : gum_v8_script_backend_get_scheduler (fixture->backend);
  gum_script_scheduler_push_job_on_js_thread (scheduler, G_PRIORITY_DEFAULT,
    (GumScriptJobFunc) gum_interceptor_ignore_current_thread, g_object_ref (interceptor), g_object_unref);
  COMPILE_AND_LOAD_SCRIPT(...);
  gum_script_scheduler_push_job_on_js_thread (scheduler, G_PRIORITY_DEFAULT,
    (GumScriptJobFunc) gum_interceptor_unignore_current_thread, g_object_ref (interceptor), g_object_unref);

  ...
}
```

However, I don't think I understand enough to be sure about this.

bindings/gumjs/gumdukcore.c
```diff
@@ -2742,7 +2742,9 @@ gum_duk_native_function_invoke (GumDukNativeFunction * self,
 
     if (gum_exceptor_try (core->exceptor, &exceptor_scope))
     {
+      gum_interceptor_unignore_current_thread (core->interceptor);
       ffi_call (&self->cif, implementation, rvalue, avalue);
+      gum_interceptor_ignore_current_thread (core->interceptor);
```
oleavr: - The `ignore` also needs to happen when `gum_exceptor_try` returns `FALSE` because of an unhandled exception resulting in us `setjmp`ing back to the call.
- `core->interceptor` is not the Interceptor, it's DukInterceptor, i.e. the JS module, so this needs an extra de-reference to get the actual `Interceptor`. Would suggest a variable for this.

bindings/gumjs/gumdukcore.c
```diff
@@ -2753,7 +2755,9 @@ gum_duk_native_function_invoke (GumDukNativeFunction * self,
 
   if (gum_exceptor_catch (core->exceptor, &exceptor_scope))
   {
+    gum_interceptor_ignore_current_thread (core->interceptor);
     _gum_duk_throw_native (ctx, &exceptor_scope.exception, core);
+    gum_interceptor_unignore_current_thread (core->interceptor);
```
oleavr: This `unignore` call is unreadable code, as Duktape's exception handling unwinds the stack.

oleavr:
Cherry-picked into 86ef4dd24cd259d3c3a612765dab3b943eaec7c0, ready for inclusion in the upcoming release 🍻

PR #232, opened 2017-08-26
Create capstone instance during module initialization

mrmacete:
This ensures in a naive way that the capstone instance is created on all code paths which need it.


oleavr:
Thanks! 💥

PR #233, opened 2017-08-27
Fix x86writer’s sub_reg_reg, add cmp_reg_reg

mrmacete:
- make sub_reg_reg compatible with 64-bits registers, by adding REX prefix
- cmp_reg_reg is exactly the same, with different opcode


oleavr:
Thanks! ❤️

PR #236, opened 2017-09-12
Poll thread state if instruction pointer is NULL

mrmacete:
This is needed to avoid stalking threads at address 0, causing exceptions and / or deadlocks. At most it tries for 20 seconds, then gives up.


oleavr:
Thanks! 🤘

mrmacete:
thank you!

PR #237, opened 2017-09-26
Expose sqlite open flags

mrmacete:
Example: `SqliteDatabase.open(fileName, {flags: [‘readwrite’, ‘create’]})`

Possible flags are `readonly`, `readwrite`, `create`. The `create` flag can be combined only with `readwrite`. The default flags are `[‘readwrite’, ‘create’]`.

### DIFF
--- a/bindings/gumjs/gumdukdatabase.c
+++ b/bindings/gumjs/gumdukdatabase.c
@@ -49,7 +49,7 @@ static void gum_push_column (duk_context * ctx, sqlite3_stmt * statement,
 
 static const duk_function_list_entry gumjs_database_module_functions[] =
 {
-  { "open", gumjs_database_open, 1 },
+  { "_open", gumjs_database_open, 2 },
   { "openInline", gumjs_database_open_inline, 1 },
 
   { NULL, NULL, 0 }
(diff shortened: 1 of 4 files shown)


bindings/gumjs/gumv8database.cpp
```diff
@@ -157,16 +157,16 @@ _gum_v8_database_finalize (GumV8Database * self)
 GUMJS_DEFINE_FUNCTION (gumjs_database_open)
 {
   gchar * path;
+  gint flags;
   sqlite3 * handle;
   gint status;
   Local<Object> object;
 
-  if (!_gum_v8_args_parse (args, "s", &path))
+  if (!_gum_v8_args_parse (args, "si", &path, &flags))
```
oleavr: I guess `guint` / `'u'` would be more appropriate?

bindings/gumjs/gumdukdatabase.c
```diff
@@ -138,16 +138,16 @@ GUMJS_DEFINE_FUNCTION (gumjs_database_open)
...
   GumDukDatabase * self;
   const gchar * path;
+  gint flags;
   sqlite3 * handle;
   gint status;
 
   self = gumjs_module_from_args (args);
 
-  _gum_duk_args_parse (args, "s", &path);
+  _gum_duk_args_parse (args, "si", &path, &flags);
```
oleavr: I guess `guint` / `'u'` would be more appropriate?

bindings/gumjs/gumv8database.cpp
mrmacete: well, i thought about that too... but from the docs it seems `sqlite3_open_v2` consumes signed flags: https://sqlite.org/c3ref/open.html

bindings/gumjs/runtime/core.js
```diff
@@ -762,4 +762,43 @@ SourceMap.prototype.resolve = function (generatedPosition) {
...
+  readwrite: 2,
+  create: 4,
+};
+
+Object.defineProperties(SqliteDatabase, {
+  open: {
+    enumerable: true,
+    value: function (file, options) {
+      if (options === undefined)
+        options = {};
```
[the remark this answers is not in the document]
mrmacete: awesome!

bindings/gumjs/gumv8database.cpp
oleavr: Ahh, nice catch!

PR #246, opened 2017-12-07
Added some missing calls to gum_clear_cache

akiannillo:


gum/backend-arm64/gumstalker-arm64.c
```diff
@@ -1822,6 +1824,8 @@ gum_exec_ctx_ensure_helper_reachable (GumExecCtx * ctx,
   write (ctx, cw);
 
   gum_arm64_writer_flush (cw);
+  gum_clear_cache (slab->data, gum_arm64_writer_offset (cw));
```
oleavr: This may clear the wrong area, because the writer was initialized with `slab->data + slab->offset` and not `slab->data`. Use `cw.base` instead.

tests/core/arch-arm64/stalker-arm64-fixture.c
```diff
@@ -159,6 +159,8 @@ test_arm64_stalker_fixture_follow_and_invoke (TestArm64StalkerFixture * fixture,
   gum_arm64_writer_put_pop_reg_reg (&cw, ARM64_REG_X29, ARM64_REG_X30);
   gum_arm64_writer_put_ret (&cw);
 
+  gum_arm64_writer_flush (&cw);
+  gum_clear_cache (code, gum_arm64_writer_offset (&cw));
```
oleavr: Probably a bit more maintainable to use `cw.base` here.

oleavr:
Awesome! <3

PR #252, opened 2017-12-22
Expose ModuleMap’s modules to javascript

mrmacete:
Add ModuleMap.enumerateModules{Sync} with the same semantics as Process.enumerateModules{Sync}


mrmacete:
made it with a property instead, here: https://github.com/frida/frida-gum/pull/253

PR #253, opened 2017-12-27
Expose ModuleMap values() method

mrmacete:
It returns a copy of the current internal `modules` array. Example:

```
var mm = new ModuleMap();
mm.values(); // get a copy of current modules array
```

### DIFF
--- a/bindings/gumjs/gumdukmodule.c
+++ b/bindings/gumjs/gumdukmodule.c
@@ -50,7 +50,7 @@ GUMJS_DECLARE_FUNCTION (gumjs_module_map_find)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_name)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_path)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_update)
-GUMJS_DECLARE_FUNCTION (gumjs_module_map_get_modules)
+GUMJS_DECLARE_FUNCTION (gumjs_module_map_copy_values)
 
 static void gum_duk_module_filter_free (GumDukModuleFilter * filter);
 static gboolean gum_duk_module_filter_matches (const GumModuleDetails * details,
@@ -69,17 +69,6 @@ static const duk_function_list_entry gumjs_module_functions[] =
   { NULL, NULL, 0 }
 };
 
-static const GumDukPropertyEntry gumjs_module_map_values[] =
-{
-  {
-    "modules",
-    gumjs_module_map_get_modules,
-    NULL
-  },
-
-  { NULL, NULL, NULL }
-};
-
 static const duk_function_list_entry gumjs_module_map_functions[] =
 {
   { "has", gumjs_module_map_has, 1 },
@@ -87,6 +76,7 @@ static const duk_function_list_entry gumjs_module_map_functions[] =
   { "findName", gumjs_module_map_find_name, 1 },
   { "findPath", gumjs_module_map_find_path, 1 },
   { "update", gumjs_module_map_update, 0 },
+  { "values", gumjs_module_map_copy_values, 0 },
 
   { NULL, NULL, 0 }
 };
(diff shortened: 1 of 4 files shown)


bindings/gumjs/gumdukmodule.c
```diff
@@ -50,6 +50,7 @@ GUMJS_DECLARE_FUNCTION (gumjs_module_map_find)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_name)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_path)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_update)
+GUMJS_DECLARE_FUNCTION (gumjs_module_map_copy_values)
```
oleavr: This function name is nice and descriptive, but for consistency we should name it `gumjs_module_map_values`.

bindings/gumjs/gumdukmodule.c
```diff
@@ -571,6 +573,27 @@ GUMJS_DEFINE_FUNCTION (gumjs_module_map_update)
   return 0;
 }
 
+GUMJS_DEFINE_FUNCTION (gumjs_module_map_copy_values)
+{
+  GumModuleDetails * details;
+  int i;
```
oleavr: This should be `guint` to match the type of `values->len`. Variables should also be declared chronologically, so this one should go after `values`.

bindings/gumjs/gumdukmodule.c
```diff
@@ -571,6 +573,27 @@ GUMJS_DEFINE_FUNCTION (gumjs_module_map_update)
   return 0;
 }
 
+GUMJS_DEFINE_FUNCTION (gumjs_module_map_copy_values)
+{
+  GumModuleDetails * details;
```
oleavr: Would be good to move this into the loop, as it's only used there.

gum/gummodulemap.c
```diff
@@ -120,6 +120,12 @@ gum_module_map_update (GumModuleMap * self)
   g_array_sort (self->modules, (GCompareFunc) gum_module_details_compare_base);
 }
 
+const GArray *
+gum_module_map_get_values (GumModuleMap * self)
+{
+    return self->modules;
```
oleavr: Indentation should be 2 spaces here.

gum/gummodulemap.h
```diff
@@ -27,6 +27,8 @@ GUM_API const GumModuleDetails * gum_module_map_find (GumModuleMap * self,
 
 GUM_API void gum_module_map_update (GumModuleMap * self);
 
+GUM_API const GArray * gum_module_map_get_values (GumModuleMap * self);
```
oleavr: Even though it's the most correct, the `const` part here should be omitted, so the GArray API can be used without casting. E.g. `g_array_get_element_size`, which expects a non-const pointer.

bindings/gumjs/gumv8module.cpp
```diff
@@ -742,6 +744,26 @@ GUMJS_DEFINE_CLASS_METHOD (gumjs_module_map_update, GumV8ModuleMap)
...
+  auto values = gum_module_map_get_values (self->handle);
+  auto result = Array::New (isolate, values->len);
+
+  for (guint i = 0; i != values->len; i++)
+  {
+    auto details = &g_array_index (values, GumModuleDetails, i);
+    auto module = Object::New (isolate);
+    _gum_v8_object_set_ascii (module, "name", details->name, core);
+    _gum_v8_object_set_pointer (module, "base", details->range->base_address,
+      core);
```
oleavr: Hanging indentation should be 4 spaces.

bindings/gumjs/gumv8module.cpp
```diff
@@ -99,6 +99,7 @@ GUMJS_DECLARE_FUNCTION (gumjs_module_map_find)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_name)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_find_path)
 GUMJS_DECLARE_FUNCTION (gumjs_module_map_update)
+GUMJS_DECLARE_FUNCTION (gumjs_module_map_copy_values)
```
oleavr: Would be good to adjust the function name here also.

bindings/gumjs/gumdukmodule.c
oleavr: Oops, great point! Let's keep it like this 👍

oleavr:
Thanks! 🤘

PR #258, opened 2018-01-24
remove unused variable in gumelfmodule.c

elvanderb:
bloom_shift is initialized but not used, that makes the compiler sad (-Wunused-but-set-variable)


oleavr:
Thanks!

PR #259, opened 2018-01-24
fix process_threads test

elvanderb:
Make sure that threads are started before counting them

PR #263, opened 2018-01-30
Fix process threads test

oleavr:
Finalized version of #259.


oleavr:
Cherry-picked instead.

PR #264, opened 2018-02-03
Expose import offset for i/macOS

mrmacete:
The offset is the address (in the module which imports the symbol) where it gets bound.

Initialized in darwin backend for now, for other architectures is set to 0 (not shown in javascript).


oleavr:
Awesome! I'm wondering if there's a more suitable name, because `offset` makes it sound like it's an offset from some address, and not a pointer to a memory location. Perhaps `slot`, `entry` or something similar?

mrmacete:
yeah i was wondering the same, let's try with `slot` ?

oleavr:
Sounds good!

oleavr:
Released in 10.6.51 🎉

PR #265, opened 2018-02-05
Allow emitting i/macOS imports more than once per symbol

mrmacete:
- because they’ll have different values for `slot`


gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -542,21 +537,7 @@ gum_emit_import (const GumDarwinBindDetails * details,
     d.slot = 0;
   }
 
-  key = g_strconcat (
```
oleavr: We should also remove the local variable.

mrmacete: oops

PR #269, opened 2018-03-16
Skip DT_GNU_HASH in gum_store_symtab_params if unavailable

krytarowski:
DT_GNU_HASH is not available on all ELF platforms and it is a GNU extension.

This corrects a build on NetBSD/amd64 8.99.12.


oleavr:
Thanks!

PR #270, opened 2018-03-17
Handle empty or undefined error stack in v8

mrmacete:


bindings/gumjs/runtime/error-handler-v8.js
```diff
@@ -25,6 +25,11 @@ global._setUnhandledExceptionCallback(function (error) {
 });
 
 Error.prepareStackTrace = function (error, stack) {
+  if (!stack || stack.length === 0) {
```
oleavr: Can it ever be `undefined`? If so we should do it like this for consistency:

```js
  if (stack === undefined || stack.length === 0) {
```

mrmacete: no it seems it can't be `undefined`, i did hit an error like `Cannot read property 'length' of undefined` during testing this, but it turns out it was unrelated.

oleavr: Ahh ok, nice!

PR #271, opened 2018-03-19
Add OutputStream.writeMemoryRegion

mrmacete:
to stream directly from memory without making copies of the data

### DIFF
--- a/bindings/gumjs/gumdukstream.c
+++ b/bindings/gumjs/gumdukstream.c
@@ -112,6 +112,7 @@ static void gum_duk_close_output_operation_finish (GOutputStream * stream,
     GAsyncResult * result, GumDukCloseOutputOperation * self);
 GUMJS_DECLARE_FUNCTION (gumjs_output_stream_write)
 GUMJS_DECLARE_FUNCTION (gumjs_output_stream_write_all)
+GUMJS_DECLARE_FUNCTION (gumjs_output_stream_write_memory_region)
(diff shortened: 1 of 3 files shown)


bindings/gumjs/gumdukstream.c
```diff
@@ -666,6 +668,29 @@ GUMJS_DEFINE_FUNCTION (gumjs_output_stream_write_all)
       GUM_DUK_WRITE_ALL);
 }
 
+GUMJS_DEFINE_FUNCTION (gumjs_output_stream_write_memory_region)
+{
+  GumDukObject * self;
+  GumDukHeapPtr callback;
+  GumDukWriteOperation * op;
+  gconstpointer address;
+  gsize length;
```
oleavr: Nit: let's order these chronologically by use.

bindings/gumjs/gumdukstream.c
```diff
@@ -666,6 +668,29 @@ GUMJS_DEFINE_FUNCTION (gumjs_output_stream_write_all)
...
+
+  (void) ctx;
```
oleavr: We used to treat unused arguments as warnings, but only on Windows, but I ended up disabling that as there's so many cases where it's unavoidable and we just end up with "noisy" code. For now I think we should avoid adding these to newly written code. (Would be nice to remove all of these "suppressors" in one swoop.)

PR #272, opened 2018-03-20
Let go js lock during findExportByName

mrmacete:


oleavr:
Thanks! 🤘Landed in 0b14e28 together with the corresponding change to the V8 runtime.

PR #277, opened 2018-04-23
Stop worker thread before modifying old_ports

mrmacete:
This is to avoid corrupting `old_ports->count` while the worker thread is potentially iterating on it. That could have produced an infinite loop in the worker thread leading to indefinite hang or memory corruption.


gum/backend-darwin/gumexceptor-darwin.c
```diff
@@ -312,11 +312,10 @@ gum_exceptor_backend_detach (GumExceptorBackend * self)
         old_ports->flavors[port_index]);
     g_assert_cmpint (kr, ==, KERN_SUCCESS);
   }
+  gum_exceptor_backend_stop_worker_thread (self);
```
oleavr: Nit: perhaps a blank line above and below this would make it stand out a bit more and improve readability.

PR #279, opened 2018-04-24
Added carriage return replacing

DavidCatalan:
Added replace("\r", "") after reading gum headers contents to avoid issues with different OS/git configurations.


oleavr:
Thanks!

PR #287, opened 2018-06-29
Add memory search with mask

mrmacete:
The mask is bitwise AND-ed against both the needle and the haystack.

To specify the mask it's now possible to append a `:` char after the needle, then the mask with the same syntax (radare2-style), example:

```
Memory.scanSync(address, size, "13 37 13 37 : 1f ff ff f1");
```

As a plus it's now possible to specify nibble-level wildcards, which are translated to masks under the hood, example:

```
Memory.scanSync(address, size, "?3 37 13 ?7");
```

Obviously this kind of search introduces a bit of overhead, but the design principle is to avoid all overhead when not actually using masks.

### DIFF
--- a/gum/gummemory.c
+++ b/gum/gummemory.c
@@ -48,8 +48,8 @@ static GumMatchToken * gum_match_pattern_get_longest_token (
     const GumMatchPattern * self, GumMatchType type);
 static gboolean gum_match_pattern_try_match_on (const GumMatchPattern * self,
     guint8 * bytes);
-static gint gum_memcmp_mask (guint8 * haystack, guint8 * needle,
-    guint8 * mask, gsize len);
+static gint gum_memcmp_mask (const guint8 * haystack, const guint8 * needle,
+    const guint8 * mask, guint len);
 static GumMatchToken * gum_match_pattern_push_token (GumMatchPattern * self,
     GumMatchType type);
 static gboolean gum_match_pattern_seal (GumMatchPattern * self);
@@ -291,18 +291,18 @@ GumMatchPattern *
 gum_match_pattern_new_from_string (const gchar * match_combined_str)
 {
   GumMatchPattern * pattern = NULL;
-  gchar ** splitted;
+  gchar ** parts;
   const gchar * match_str, * mask_str;
   gboolean has_mask = FALSE;
   GumMatchToken * token = NULL;
   const gchar * ch, * mh;
 
-  splitted = g_strsplit (match_combined_str, ":", 2);
-  match_str = splitted[0];
+  parts = g_strsplit (match_combined_str, ":", 2);
+  match_str = parts[0];
   if (match_str == NULL)
     goto parse_error;
 
-  mask_str = splitted[1];
+  mask_str = parts[1];
   has_mask = mask_str != NULL;
   if (has_mask && strlen (mask_str) != strlen (match_str))
     goto parse_error;
@@ -379,15 +379,17 @@ gum_match_pattern_new_from_string (const gchar * match_combined_str)
   if (!gum_match_pattern_seal (pattern))
     goto parse_error;
 
-  g_strfreev (splitted);
+  g_strfreev (parts);
+
   return pattern;
 
   /* ERRORS */
 parse_error:
   {
-    g_strfreev (splitted);
+    g_strfreev (parts);
     if (pattern != NULL)
       gum_match_pattern_free (pattern);
+
     return NULL;
   }
 }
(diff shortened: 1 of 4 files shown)


oleavr:
Wow, awesome! I will review later this weekend – need a clearer head :-)

gum/gummemory.c
```diff
@@ -263,54 +288,106 @@ gum_memory_scan (const GumMemoryRange * range,
...
-  GumMatchPattern * pattern;
+  GumMatchPattern * pattern = NULL;
+  gchar ** splitted;
+  const gchar * match_str, * mask_str;
+  gboolean has_mask = FALSE;
   GumMatchToken * token = NULL;
-  const gchar * ch;
+  const gchar * ch, * mh;
+
+  splitted = g_strsplit (match_combined_str, ":", 2);
```
oleavr: Nit: `split` would be grammatically better, but is also a verb so maybe it's better to name this `parts`?

gum/gummemory.c
```diff
@@ -263,54 +288,106 @@ gum_memory_scan (const GumMemoryRange * range,
...
+  match_str = splitted[0];
+  if (match_str == NULL)
+    goto parse_error;
+
+  mask_str = splitted[1];
+  has_mask = mask_str != NULL;
+  if (has_mask && strlen (mask_str) != strlen (match_str))
```
oleavr: This compares the inputs before any ignored spaces have been considered. I suppose we could fix this by moving the check later, but OTOH enforcing consistency is a good thing so we could also just leave this as it is.

gum/gummemory.c
```diff
@@ -263,54 +288,106 @@ gum_memory_scan (const GumMemoryRange * range,
...
+    }
 
     ch++;
+    mh++;
   }
 
   if (!gum_match_pattern_seal (pattern))
     goto parse_error;
 
+  g_strfreev (splitted);
```
oleavr: Nit: a blank line between “unwind” and “return” would improve readability slightly.

gum/gummemory.c
```diff
@@ -263,54 +288,106 @@ gum_memory_scan (const GumMemoryRange * range,
...
   return pattern;
 
   /* ERRORS */
 parse_error:
   {
-    gum_match_pattern_free (pattern);
+    g_strfreev (splitted);
+    if (pattern != NULL)
+      gum_match_pattern_free (pattern);
```
oleavr: Nit: a blank line between “unwind” and “return” would improve readability slightly.

gum/gummemory.c
```diff
@@ -379,6 +456,7 @@ gum_match_pattern_try_match_on (const GumMatchPattern * self,
                                 guint8 * bytes)
 {
   guint i;
+  gboolean there_are_masks = FALSE;
```
oleavr: Nit: How about flipping this to `there_are_masks` => `no_masks`.

gum/gummemory.c
```diff
@@ -396,11 +474,55 @@ gum_match_pattern_try_match_on (const GumMatchPattern * self,
...
+    GumMatchToken * token;
+
+    token = (GumMatchToken *) g_ptr_array_index (self->tokens, i);
+    if (token->type == GUM_MATCH_MASK)
+    {
+      gchar * p;
+
+      p = (gchar *) bytes + token->offset;
+      if (gum_memcmp_mask ((guint8 *) p, (guint8 *) token->bytes->data,
+         (guint8 *) token->masks->data, token->masks->len) != 0)
```
oleavr: Missing one space of indentation.

gum/gummemory.c
```diff
@@ -396,11 +474,55 @@ gum_match_pattern_try_match_on (const GumMatchPattern * self,
...
   return TRUE;
 }
 
+static gint
+gum_memcmp_mask (guint8 * haystack,
+                 guint8 * needle,
+                 guint8 * mask,
+                 gsize len)
+{
+  guint i;
```
oleavr: I'd say `i` should be the same type as `len`. Maybe using `guint` for `len` is better since `GArray`'s `len` field is a `guint`.

gum/gummemory.c
```diff
@@ -396,11 +474,55 @@ gum_match_pattern_try_match_on (const GumMatchPattern * self,
...
+    }
   }
 
   return TRUE;
 }
 
+static gint
+gum_memcmp_mask (guint8 * haystack,
+                 guint8 * needle,
+                 guint8 * mask,
```
oleavr: These could all be `const` right?

PR #289, opened 2018-07-18
Removed unnecessary (and bug inducing) definition of getcontext and setcontext

r0ck3tAKATrashPanda:
We don't need these with the proper toolchains available. Part of fixing mips support.


oleavr:
Thanks!

PR #290, opened 2018-08-09
Let gum_file_open_tmp fail

mrmacete:
Instead of failing the assertion, that will let the try_realize fail and take the alternative path before giving up.

This is enough to fix attach to `com.apple.WebKit.*` processes on iOS 11.x


gum/backend-darwin/gumcodesegment-darwin.c
```diff
@@ -274,6 +274,9 @@ gum_code_segment_try_realize (GumCodeSegment * self)
 
   self->fd = gum_file_open_tmp ("frida-XXXXXX.dylib", &dylib_path);
```
oleavr: Let's drop this blank line so it's grouped with the assignment.

PR #291, opened 2018-08-10
Free linux auxv struct

korniltsev:
g_file_get_contents allocates buffer for /proc/$pid/auxv file and it was not freed


oleavr:
Thanks! 🤘Great catch! That function is apparently full of fail :-) Feel free to open another PR for the loop fix btw – or I can do it, just let me know.

korniltsev:
I will be happy to do another PR.

PR #296, opened 2018-09-03
More kernel improvements

mrmacete:
- add scan, enumerateModules, enumerateModuleRanges
- add kernel base getter
- fix overflow on section / segment names

this works on 64-bit devices, tested on versions from iOS 9.0 to iOS 11.3.

## Added interfaces

`Kernel.scan()` / `Kernel.scanSync()` work exactly like their `Memory` counterparts, supporting bit masks too.

`Kernel.enumerateModules()` / `Kernel.enumerateModulesSync()` lists all the KEXTs, and the main kernel module (called `"Kernel"`), example:

```
[iPhone::SystemSession]-> Kernel.enumerateModulesSync()
[
    {
        "base": "0xfffffff016604000",
        "name": "Kernel",
        "path": null,
        "size": 31886248
    },
    {
        "base": "0xfffffff0152bc940",
        "name": "com.apple.driver.AppleEffaceableStorage",
        "path": null,
        "size": 28550
    },
    {
        "base": "0xfffffff015272dc0",
        "name": "com.apple.driver.AppleMultitouchSPI",
        "path": null,
        "size": 111560
    },
...
```

`Kernel.enumerateModuleRanges()` / `Kernel.enumerateModuleRangesSync()` list sections of a module by name (taken by the previous call). The semantic is the same of `Module.enumerateRanges()` Example:

```
[iPhone::SystemSession]-> Kernel.enumerateModuleRangesSync('com.apple.driver.AppleMultitouchSPI', '---')
[
    {
        "address": "0xfffffff0152733c8",
        "name": "__TEXT.__const",
        "protection": "r--",
        "size": 288
    },
    {
        "address": "0xfffffff0152734e8",
        "name": "__TEXT.__cstring",
        "protection": "r--",
        "size": 14675
    },
    {
        "address": "0xfffffff015276e3b",
        "name": "__TEXT.__os_log",
        "protection": "r--",
        "size": 1577
    },
    {
        "address": "0xfffffff015a44000",
        "name": "__TEXT_EXEC.__text",
        "protection": "r-x",
        "size": 78788
    },
...
```

the `Kernel.base` property returns a pointer to the base of the kernel in memory (intended as the address of its `__TEXT` segment).

### DIFF
--- a/bindings/gumjs/gumdukkernel.c
+++ b/bindings/gumjs/gumdukkernel.c
@@ -9,6 +9,7 @@
 #include "gumdukmacros.h"
 
 typedef struct _GumDukMatchContext GumDukMatchContext;
+typedef struct _GumKernelScanContext GumKernelScanContext;
 
 struct _GumDukMatchContext
 {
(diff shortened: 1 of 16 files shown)


bindings/gumjs/gumdukkernel.c
```diff
@@ -258,6 +415,191 @@ GUMJS_DEFINE_FUNCTION (gumjs_kernel_write_byte_array)
...
+static void
+gum_kernel_scan_context_run (GumKernelScanContext * self)
+{
+  GumDukCore * core = self->core;
+  GumExceptor * exceptor = core->exceptor;
+  GumExceptorScope exceptor_scope;
+  GumDukScope script_scope;
+  duk_context * ctx;
+
+  if (gum_exceptor_try (exceptor, &exceptor_scope))
```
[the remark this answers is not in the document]
mrmacete: LOL yeah, it directly panics

gum/gumkernel.h
```diff
@@ -20,8 +33,17 @@ GUM_API guint8 * gum_kernel_read (GumAddress address, gsize len,
...
+    const GumMatchPattern * pattern, GumMemoryScanMatchFunc func,
+    gpointer user_data);
 GUM_API void gum_kernel_enumerate_ranges (GumPageProtection prot,
     GumFoundRangeFunc func, gpointer user_data);
+GUM_API void gum_kernel_enumerate_module_ranges (const gchar * module_name,
+    GumPageProtection prot, GumFoundKernelModuleRangeFunc func,
+    gpointer user_data);
+GUM_API void gum_kernel_enumerate_modules (GumFoundModuleFunc func,
+    gpointer user_data);
+GUM_API gpointer gum_kernel_find_base_address (void);
```
oleavr: Just occurred to me that this should probably return `GumAddress` too for consistency with the other APIs?

mrmacete:
nice, only doubt: `g_strlcpy` assumes the `src` string is nul terminated, which isn't the case here (it can use all the 16 bytes reserved for that field), from the [docs](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy):

```
src must be nul-terminated; dest_size is the buffer size, not the number of bytes to copy.
```

probably it works anyways, though. i'll give it a try!

oleavr:
Ooh good point! The only drawback with that is that it assumes that the destination buffer is exactly n + 1, and that isn't obvious, so I'd say the current code is clearer.

bindings/gumjs/gumv8value.cpp
```diff
@@ -1503,6 +1503,18 @@ _gum_v8_object_set_pointer (Handle<Object> object,
       core);
 }
 
+gboolean
+_gum_v8_object_set_uint64 (Handle<Object> object,
+                            const gchar * key,
+                            GumAddress value,
+                            GumV8Core * core)
```
oleavr: Indentation off by one.

gum/backend-darwin/gumkernel-darwin.c
```diff
@@ -105,13 +105,19 @@ static GumDarwinModule * gum_kernel_find_module_by_name (
...
     GumKernelKextByNameContext * ctx);
 static GumDarwinModule * gum_kernel_get_module (void);
-static gpointer gum_kernel_do_find_base_address (void);
+static GumAddress gum_kernel_do_find_base_address (void);
 static float gum_kernel_get_version (void);
-static gpointer gum_kernel_bruteforece_base (GumAddress unslid_base);
+
+#ifdef HAVE_ARM64
+
+static GumAddress gum_kernel_bruteforece_base (GumAddress unslid_base);
```
oleavr: `bruteforece` => `bruteforce`

gum/backend-darwin/gumkernel-darwin.c
```diff
@@ -179,14 +185,13 @@ gum_kernel_try_mprotect (gpointer address,
     return FALSE;
 
   page_size = vm_kernel_page_size;
-  aligned_address = GSIZE_TO_POINTER (
-      GPOINTER_TO_SIZE (address) & ~(page_size - 1));
+  aligned_address = address & ~(page_size - 1);
```
oleavr: Nit: `page_size` should be a `GumAddress` too, or cast to it, so the bitwise operation works on 32-bit.

gum/backend-darwin/gumkernel-darwin.c
```diff
@@ -628,28 +633,28 @@ gum_kernel_get_module (void)
...
-gpointer
+GumAddress
 gum_kernel_find_base_address (void)
 {
   static GOnce get_base_once = G_ONCE_INIT;
 
   g_once (&get_base_once, (GThreadFunc) gum_kernel_do_find_base_address, NULL);
 
-  return (gpointer) get_base_once.retval;
+  return (GumAddress) get_base_once.retval;
```
oleavr: This will truncate the address on 32-bit.

oleavr:
Merged.

PR #298, opened 2018-09-05
ARM64 relocator x16 to x18 migration

hexploitable:
Modified the arm64 relocator to use `x18` instead of `x16` since `x18` is actually unreserved.
`x16` is used for supervisor calls so when we clobber it we can cause applications to crash. 
E.g. using interceptor with the address of an svc instruction will clobber the syscall value held in `x16`.

@oleavr this works, have tried and tested it and I can now see the values in x18 instead, also x16 remains unclobbered. I'm not sure if all of the replacements were necessary, i'll let you comment since you're the brains 😄.

As discussed previously, there is definitely a more elegant way to preserve state without clobbering registers but this works for now. I think `x18` is definitely most suitable for iOS at least.


oleavr:
Thanks, let's move to X18! 👍

PR #304, opened 2018-10-01
Add optional backtracer argument.

leonjza:


oleavr:
Thanks! How about also adding the enum while at it?

PR #315, opened 2018-12-06
Show the actual start address when viewing a pointer.

FrankSpierings:
I have only tested this on 32bits. So using the `NativePointer.toInt32()` works. Not sure how this should work on 64-bits adressing. I'm also not sure why NativePointer will only return signed integers. Isn't it more logical to return an unsigned int?

### DIFF
--- a/bindings/gumjs/runtime/hexdump.js
+++ b/bindings/gumjs/runtime/hexdump.js
@@ -11,6 +11,7 @@ function hexdump(target, options) {
   const useAnsi = options.hasOwnProperty('ansi') ? options.ansi : false;
 
   let buffer;
+  let startAddress = NULL;
   if (target instanceof ArrayBuffer) {
     if (length === undefined)
       length = target.byteLength;
@@ -19,6 +20,7 @@ function hexdump(target, options) {
     if (length === undefined)
       length = 256;
     buffer = Memory.readByteArray(target, length);
+    startAddress = target;
   }
 
   const bytes = new Uint8Array(buffer);
@@ -60,7 +62,7 @@ function hexdump(target, options) {
       result.push('\n');
 
     result.push(
-      offsetColor, pad(offset.toString(16), leftColumnWidth, '0'), resetColor,
+      offsetColor, pad(startAddress.add(offset).toString(16), leftColumnWidth, '0'), resetColor,
       columnPadding
     );
 


bindings/gumjs/runtime/hexdump.js
```diff
@@ -19,6 +20,7 @@ function hexdump(target, options) {
     if (length === undefined)
       length = 256;
     buffer = Memory.readByteArray(target, length);
+    startAddress = target.toInt32() >>> 0;
```
oleavr: This will truncate some pointers in 64-bit processes. In general never turn a pointer into a JavaScript number like here, because those are backed by `double`, which means your pointer might get rounded to a different address (lol).

bindings/gumjs/runtime/hexdump.js
```diff
@@ -59,8 +61,9 @@ function hexdump(target, options) {
     if (bufferOffset !== 0)
       result.push('\n');
 
+    let address = (startAddress + offset).toString(16);
     result.push(
-      offsetColor, pad(offset.toString(16), leftColumnWidth, '0'), resetColor,
+      offsetColor, pad(address, leftColumnWidth, '0'), resetColor,
```
oleavr: Should be possible to simplify this to just:

```suggestion
      offsetColor, pad(startAddress.add(offset).toString(16), leftColumnWidth, '0'), resetColor,
```

bindings/gumjs/runtime/hexdump.js
```diff
@@ -11,6 +11,7 @@ function hexdump(target, options) {
   const useAnsi = options.hasOwnProperty('ansi') ? options.ansi : false;
 
   let buffer;
+  let startAddress = 0;
```
oleavr:
```suggestion
  let startAddress = NULL;
```

bindings/gumjs/runtime/hexdump.js
oleavr:
```suggestion
    startAddress = target;
```

bindings/gumjs/runtime/hexdump.js
```diff
@@ -59,8 +61,9 @@ function hexdump(target, options) {
     if (bufferOffset !== 0)
       result.push('\n');
 
+    let address = (startAddress + offset).toString(16);
```
oleavr:
```suggestion
```

bindings/gumjs/runtime/hexdump.js
FrankSpierings: Agreed!

bindings/gumjs/runtime/hexdump.js
FrankSpierings: What happens if the address is from an array, not a pointer? NULL + offset in Javascript?

oleavr: It's just a pointer value used for display purposes, i.e. it's never dereferenced, so it's safe also in the `NULL` case.

bindings/gumjs/runtime/hexdump.js
FrankSpierings: Agreed

bindings/gumjs/runtime/hexdump.js
FrankSpierings: You are far the better programmer, so I believe you ;).

oleavr:
Yay! Did you test this final version?

FrankSpierings:
No will

> Yay! Did you test this final version?

No I will, but It might be tomorrow.

oleavr:
Thanks! 🤘

FrankSpierings:
I managed to compile it into the new version. I've tested it inside an Android (ARM 32bits) and it seems to work correctly on `console.log(hexdump(new ArrayBuffer(64)))` and `console.log(hexdump(Module.findExportByName("libc.so", "open")))`.

Any way to inject the gum scripts without a full recompilation of the gadget?

oleavr:
> I managed to compile it into the new version. I've tested it inside an Android …

Cool, thanks!

> Any way to inject the gum scripts without a full recompilation of the gadget?

No, but you can assign a new function to `global.hexdump` to replace the builtin version with a local one, e.g. hacked on using a [frida-compile-based agent](https://github.com/oleavr/frida-agent-example).

oleavr:
One last thing before merging – could you squash your commits and then fix the author so it's exactly like this: `Full Name <foo@bar.com>`. E.g. first squash then `git commit --amend --author="Full Name <foo@bar.com>"`.

FrankSpierings:
I hope that I have done what you asked, but I consider myself a git noob once you start talking about squashing etc. So it seems fine to me, but I might have created something horrible, please check my work ;)

oleavr:
Thanks! :) I tweaked the commit message style for consistency with [the style we use](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

PR #322, opened 2019-01-29
Fix trailing junk data in ansi_string_to_utf8

clouds56:
See the [doc](https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/nf-stringapiset-multibytetowidechar)

### DIFF
--- a/bindings/gumjs/gumdukmemory.c
+++ b/bindings/gumjs/gumdukmemory.c
@@ -737,6 +737,7 @@ gum_ansi_string_to_utf8 (const gchar * str_ansi,
                          gint length)
 {
   guint str_utf16_size;
+  gint str_utf16_length;
   WCHAR * str_utf16;
   gchar * str_utf8;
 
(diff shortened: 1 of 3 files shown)


clouds56:
Haven't been tested, just proposing the change and needs review.

oleavr:
Good catch. Please write a test-case demonstrating the desired behavior.

clouds56:
I think it is a little complicate to test this behavior, the CP_ACP value differs between computers. The issue would only occurs in the CodePages that have [multi-bytes](https://docs.microsoft.com/en-us/windows/desktop/intl/double-byte-character-sets). (cp1250 for english is [single byte](https://docs.microsoft.com/en-us/windows/desktop/intl/single-byte-character-sets) codepage).

clouds56:
I add a test case (and flag G_OS_ACP_936 to enable).
There's no guide for build and test, so I haven't tried it myself. (why is there no CI for the repo?)

oleavr:
Thank you! Sorry for the delay here.

We do have a CI but it's not yet wired up to GitHub. Could you try running this test locally? All you need is VS2017 (Community edition will do), then run `gum-tests`, and to run only one test pass it `-p /Path/to/test`.

oleavr:
Timed out. Please reopen in the future if you find the time to look into this.

oleavr:
Actually let me take a quick look at this now, as I need to land another PR on Windows.

oleavr:
Finalized version landed in cdf6643225cc45a004bb0c5cf7d22b2a28e59cb0. Thanks for reporting and helping fix this! ❤️

PR #323, opened 2019-02-06
Fix have_v8 undefined

xiaoyur347:
If the options enable_gumpp and enable_gumjs in meson_options.txt are set to false, it will fail to compile as have_v8 is undefined.


oleavr:
Nice catch! Thanks!

PR #327, opened 2019-03-17
add missing Java.array api.

henrayluo:


oleavr:
Thanks! Made a couple of minor tweaks while at it.

PR #329, opened 2019-03-21
Tweak darwinmodule for quick app check

mrmacete:
- add _enumerate_dependencies api call
- add the “header-only” property, to load only the header and the commands instead of the whole binary, applies to file system loading
- enable gumdarwinmodule to load executables too, not only libs

### DIFF
--- a/gum/backend-darwin/gumdarwinmodule.c
+++ b/gum/backend-darwin/gumdarwinmodule.c
@@ -19,6 +19,9 @@
     (self->is_kernel ? gum_kernel_read (addr, len, out_size) \
       : gum_darwin_read (task, addr, len, out_size))
 
+#define HAS_HEADER_ONLY(self) \
+    ((self->flags & GUM_DARWIN_MODULE_FLAGS_HEADER_ONLY) != 0)
+
 typedef struct _GumResolveSymbolContext GumResolveSymbolContext;
 
 typedef struct _GumEmitImportContext GumEmitImportContext;
@@ -44,7 +47,7 @@ enum
   PROP_SOURCE_PATH,
   PROP_SOURCE_BLOB,
   PROP_CACHE_FILE,
-  PROP_HEADER_ONLY,
+  PROP_FLAGS,
 };
 
 struct _GumResolveSymbolContext
@@ -237,9 +240,9 @@ gum_darwin_module_class_init (GumDarwinModuleClass * klass)
       g_param_spec_boxed ("cache-file", "CacheFile", "Cache file used by dyld",
       G_TYPE_MAPPED_FILE,
       G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS));
-  g_object_class_install_property (object_class, PROP_HEADER_ONLY,
-      g_param_spec_boolean ("header-only", "HeaderOnly", "Load only the header",
-      FALSE,
+  g_object_class_install_property (object_class, PROP_FLAGS,
+      g_param_spec_uint ("flags", "Flags", "Optional flags", 0, G_MAXUINT,
+      GUM_DARWIN_MODULE_FLAGS_NONE,
       G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS));
 }
 
@@ -326,7 +329,7 @@ gum_darwin_module_initable_init (GInitable * initable,
         !gum_darwin_module_try_load_image_from_cache (self, self->source_path,
         self->cpu_type, self->cache_file))
     {
-      if (self->header_only)
+      if (HAS_HEADER_ONLY (self))
       {
         if (!gum_darwin_module_load_image_header_from_filesystem (self,
             self->source_path, self->cpu_type, error))
@@ -428,8 +431,8 @@ gum_darwin_module_get_property (GObject * object,
     case PROP_CACHE_FILE:
       g_value_set_boxed (value, self->cache_file);
       break;
-    case PROP_HEADER_ONLY:
-      g_value_set_boolean (value, self->header_only);
+    case PROP_FLAGS:
+      g_value_set_uint (value, self->flags);
       break;
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
@@ -474,8 +477,8 @@ gum_darwin_module_set_property (GObject * object,
       g_clear_pointer (&self->cache_file, g_mapped_file_unref);
       self->cache_file = g_value_dup_boxed (value);
       break;
-    case PROP_HEADER_ONLY:
-      self->header_only = g_value_get_boolean (value);
+    case PROP_FLAGS:
+      self->flags = g_value_get_uint (value);
       break;
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
@@ -488,7 +491,7 @@ gum_darwin_module_new_from_file (const gchar * path,
                                  GumCpuType cpu_type,
                                  guint page_size,
                                  GMappedFile * cache_file,
-                                 gboolean header_only,
+                                 GumDarwinModuleFlags flags,
                                  GError ** error)
 {
   return g_initable_new (GUM_DARWIN_TYPE_MODULE, NULL, error,
@@ -496,7 +499,7 @@ gum_darwin_module_new_from_file (const gchar * path,
       "cpu-type", cpu_type,
       "page-size", page_size,
       "source-path", path,
-      "header-only", header_only,
+      "flags", flags,
       NULL);
 }
 
@@ -505,6 +508,7 @@ gum_darwin_module_new_from_blob (GBytes * blob,
                                  mach_port_t task,
                                  GumCpuType cpu_type,
                                  guint page_size,
+                                 GumDarwinModuleFlags flags,
                                  GError ** error)
 {
   return g_initable_new (GUM_DARWIN_TYPE_MODULE, NULL, error,
@@ -512,6 +516,7 @@ gum_darwin_module_new_from_blob (GBytes * blob,
       "cpu-type", cpu_type,
       "page-size", page_size,
       "source-blob", blob,
+      "flags", flags,
       NULL);
 }
 
@@ -521,6 +526,7 @@ gum_darwin_module_new_from_memory (const gchar * name,
                                    GumCpuType cpu_type,
                                    guint page_size,
                                    GumAddress base_address,
+                                   GumDarwinModuleFlags flags,
                                    GError ** error)
 {
   return g_initable_new (GUM_DARWIN_TYPE_MODULE, NULL, error,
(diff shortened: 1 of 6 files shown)


gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1268,6 +1298,51 @@ gum_emit_section_term_pointers (const GumDarwinSectionDetails * details,
...
+          return;
+
+        break;
+      }
+      default:
+        break;
+    }
+
+    command += lc->cmdsize;
+  }
```
oleavr: Wondering if we should roll this into the load step, so we can populate `dependencies` like we normally do.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
 
+static gboolean
+gum_darwin_module_load_image_header_from_filesystem (GumDarwinModule * self,
+                                                     const gchar * path,
+                                                     GumCpuType cpu_type,
+                                                     GError ** error)
+{
+  gboolean success;
+  GMappedFile * file;
+  gsize size, size_in_pages, page_size;
```
oleavr: `page_size` should be first of the three, as it's used first.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
+  gpointer data;
+  GBytes * blob;
+  gsize header_offset = 0, header_size, cursor = 0;
```
oleavr: - `header_size` can be moved into the loop's scope.
- For clarity it would be good to initialize `cursor` just before the loop.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
+        {
+          case CPU_TYPE_ARM:
+          case CPU_TYPE_X86:
+            *out_size = sizeof (struct mach_header);
+            found = self->cpu_type == GUM_CPU_ARM || self->cpu_type == GUM_CPU_IA32;
+            break;
+          case CPU_TYPE_ARM64:
+          case CPU_TYPE_X86_64:
+            *out_size = sizeof (struct mach_header_64);
+            found = self->cpu_type == GUM_CPU_ARM64 || self->cpu_type == GUM_CPU_AMD64;
```
oleavr: Exceeds 80 cols.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
+
+        guint32 offset = GUINT32_FROM_BE (fat_arch->offset);
+        guint32 cpu_type = GUINT32_FROM_BE (fat_arch->cputype);
+        *out_offset = offset;
+        switch (cpu_type)
+        {
+          case CPU_TYPE_ARM:
+          case CPU_TYPE_X86:
+            *out_size = sizeof (struct mach_header);
+            found = self->cpu_type == GUM_CPU_ARM || self->cpu_type == GUM_CPU_IA32;
```
oleavr: Exceeds 80 cols.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
+      *out_size = sizeof (struct mach_header) + header_32->sizeofcmds;
+      found = self->cpu_type == GUM_CPU_ARM || self->cpu_type == GUM_CPU_IA32;
+      break;
+    case MH_MAGIC_64:
+      header_64 = (struct mach_header_64 *) data;
+      if (data + sizeof (struct mach_header_64) > data_end)
+        goto invalid_blob;
+      *out_offset = 0;
+      *out_size = sizeof (struct mach_header_64) + header_64->sizeofcmds;
+      found = self->cpu_type == GUM_CPU_ARM64 || self->cpu_type == GUM_CPU_AMD64;
```
oleavr: Exceeds 80 cols.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -1400,6 +1475,150 @@ gum_darwin_module_load_image_from_filesystem (GumDarwinModule * self,
...
+    return FALSE;
+
+  page_size = gum_query_page_size ();
+  data = gum_alloc_n_pages (1, GUM_PAGE_RW);
+  size = page_size;
+
+  do
+  {
+    memcpy (data, g_mapped_file_get_contents (file) + cursor, size);
+    if (!gum_darwin_module_get_header_offset_size (self, data, size, &header_offset, &header_size, error))
```
oleavr: Exceeds 80 cols.

oleavr:
Almost forgot: vapi also needs to be updated.

gum/backend-darwin/gumdarwinmodule.c
mrmacete: uh, actually `header_size` is used also after the loop

gum/backend-darwin/gumdarwinmodule.c
mrmacete: yeah right, that was my first thought too since we're already storing dependencies in `_take_image`. my first implementation was:
```
  const gchar * dep;
  gint index = 0;

  if (!gum_darwin_module_ensure_image_loaded (self, NULL))
    return;

  while ((dep = g_ptr_array_index (self->dependencies, index++)) != NULL)
  {
    if (!func (dep, user_data))
      return;
  }
```
but this seems to cause some kind of corruption and everything ends up hanging and i have to restart SpringBoard. i'm looking into that - not sure why it's behaving like this since the pointers should be stable as long as the image is there.

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -44,6 +44,7 @@ enum
   PROP_SOURCE_PATH,
   PROP_SOURCE_BLOB,
   PROP_CACHE_FILE,
+  PROP_HEADER_ONLY,
```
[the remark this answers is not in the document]
mrmacete: cool! any example of this i can inspire from?

gum/backend-darwin/gumdarwinmodule.c
oleavr: Oops, sorry, scratch that part 😄

gum/backend-darwin/gumdarwinmodule.c
oleavr: It's because `g_ptr_array_index()` intentionally doesn't do any bounds-checking, so you are meant to use `->len`.

gum/backend-darwin/gumdarwinmodule.c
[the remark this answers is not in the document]
oleavr: Perhaps [GDBusConnectionFlags](https://developer.gnome.org/gio//2.32/GDBusConnection.html#GDBusConnectionFlags)

gum/backend-darwin/gumdarwinmodule.c
mrmacete: ahh "nice" !!

gum/backend-darwin/gumdarwinmodule.c
mrmacete: lol yeah it was `header_offset` instead 😅

oleavr: Yeah another case of me thinking one thing and saying something else 😂

mrmacete:
i'm especially unsure about how i tweaked vapi

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -406,6 +431,9 @@ gum_darwin_module_get_property (GObject * object,
     case PROP_CACHE_FILE:
       g_value_set_boxed (value, self->cache_file);
       break;
+    case PROP_FLAGS:
+      g_value_set_uint (value, self->flags);
```
oleavr: Should be using `g_value_set_flags`.

gum/backend-darwin/gumdarwinmodule.h
```diff
@@ -94,6 +101,8 @@ struct _GumDarwinModule
 
   GPtrArray * dependencies;
   GPtrArray * reexports;
+
+  GumDarwinModuleFlags flags;
```
oleavr: Should be moved right after `cache_file`, to follow the order that the properties are defined in.

vapi/frida-gum-darwin-1.0.vapi
```diff
@@ -52,9 +52,16 @@ namespace Gum.Darwin {
 			get;
 		}
 
-		public Module.from_file (string path, Port task, Gum.CpuType cpu_type, uint page_size, GLib.MappedFile? cache_file = null) throws GLib.Error;
-		public Module.from_blob (GLib.Bytes blob, Port task, Gum.CpuType cpu_type, uint page_size) throws GLib.Error;
-		public Module.from_memory (string? name, Port task, Gum.CpuType cpu_type, uint page_size, Gum.Address base_address) throws GLib.Error;
+		[Flags]
+		[CCode (cprefix = "GUM_DARWIN_MODULE_FLAGS_")]
```
oleavr: The `CCode` annotation is not needed as the it's the same as the default.

vapi/frida-gum-darwin-1.0.vapi
```diff
@@ -52,9 +52,16 @@ namespace Gum.Darwin {
...
+		public enum Flags {
+			NONE = 0,
+			HEADER_ONLY	  = (1 << 0),
```
oleavr: The values should not be specified in a vapi. (TIL – I made the same mistake 😅)

mrmacete:
> The get_type() for the enum is done like this.

ohh nice, yeah i wonder if in this case we should use a [GFlagsValue](https://developer.gnome.org/gobject/stable/gobject-Enumeration-and-Flag-Types.html#GFlagsValue) instead

mrmacete:
also, i'm not sure to get who is calling `get_type()` for that, eventually

mrmacete:
> also, i'm not sure to get who is calling get_type() for that, eventually

ahh ok nvm, that's passed to `g_param_spec_flags()`

gum/backend-darwin/gumdarwinmodule.c
```diff
@@ -2328,3 +2328,26 @@ gum_dyld_cache_offset_from_address (GumAddress address,
...
+gum_darwin_module_flags_get_type (void)
+{
+  static volatile gsize gonce_value;
+
+  if (g_once_init_enter (&gonce_value))
+  {
+    static const GFlagsValue values[] =
+    {
+      { GUM_DARWIN_MODULE_FLAGS_NONE, "GUM_DARWIN_MODULE_FLAGS_NONE", "default" },
+      { GUM_DARWIN_MODULE_FLAGS_HEADER_ONLY, "GUM_DARWIN_MODULE_FLAGS_HEADER_ONLY", "optional" },
```
oleavr: `default` and `none` should be replaced here.

PR #336, opened 2019-05-25
Make libc API public

iGio90:


iGio90:
pt1 for comments:
https://github.com/frida/frida-core/pull/244

PR #340, opened 2019-07-28
Stalker: fix uaf and allow unfollow when transforming

iGio90:
Actually, there is no way to unfollow a stalked thread which uses transform.
Calling unfollow on the same thread will trigger an uaf which crash frida (it attempt to use a freed writer)
Calling unfollow on a different thread will trigger another unknown bug which is gonna be investigated later if needed but this, will open one little door to detach a stalked thread.
From my understanding, we should just unfollow when it's safe to do it. There is also missing logic when unfollow is triggered from a callout in example, which trigger another bug

todo: missing test case

### DIFF
--- a/gum/backend-arm64/gumstalker-arm64.c
+++ b/gum/backend-arm64/gumstalker-arm64.c
@@ -136,6 +136,8 @@ struct _GumExecCtx
   GumEvent tmp_event;
 
   gboolean unfollow_called_while_still_following;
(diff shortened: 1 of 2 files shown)


oleavr:
This and other edge-cases have been addressed in c4ab181cc7f4ad56ea05c650edfe9403b1f29631. Thanks! 🤘

Issue #343, opened 2019-08-20
Segmentation fault by gum_memory_patch_code() with buffer exceed page size on ARM64


gebing:
If invoking `gum_memory_patch_code()` with buffer size exceed page size on Android ARM64, it will make frida crash with SIGSEGV signal. And this is the reason of 'Java.enumerateLoadedClasses() terminates the process' [https://github.com/frida/frida-java-bridge/issues/132](url), by the code below:

```
const codeSize = 65536;
const code = Memory.alloc(codeSize);
Memory.patchCode(code, codeSize, buffer => {
  perform = recompile(buffer, code, exceptionClearImpl, nextFuncImpl, exceptionOffset, neuteredOffsets, callback);
});
```

The bug is because calling `gum_clear_cache()` after `gum_try_mprotect (start_page, range_size, GUM_PAGE_RX)`, and should be called before `gum_try_mprotect`.

@oleavr 
I will make a pull request later.

gebing:
This bug seems only on Android ARM64, and not on ARM32/X86/X64, but i do not know the reason why.

gebing:
If the buffer size is less than page size, this bug will not occur, but if buffer size is lager than 0x1100, it will definitely occur on my test phone. My test phone is RedMi 4, Android 6.1, ARM64.

PR #344, opened 2019-08-20
Fix for segmentation fault by gum_memory_patch_code()

gebing:
Fix for Issue #343, Segmentation fault by gum_memory_patch_code() with buffer exceed page size on ARM64


oleavr:
Wow, that's weird. But let's make this change, sure won't hurt.

oleavr:
Thanks, merged!

PR #345, opened 2019-08-21
Fix for duktape's Reflect.get() restriction

gebing:
Fix for duktape's Reflect.get() restriction, which will cause Java.choose throw unsupported error.

### DIFF
--- a/bindings/gumjs/runtime/error-handler-duktape.js
+++ b/bindings/gumjs/runtime/error-handler-duktape.js
@@ -122,3 +122,11 @@ function findSourceMap(source) {
   else
     return null;
 }
+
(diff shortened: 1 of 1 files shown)


oleavr:
Thank you so much for tracking this down, much appreciated!

As mentioned in frida/frida#989, I gave this some thought and realized that this is similar to the issue we were dealing with in 83e98b4, so I figured it's better to disable the builtin instead – so I went ahead and did that in fab4744. This way, since we're using frida-compile w/ Babel under the hood, its polyfill will use its fully standard-compliant version instead of the non-standard one built into Duktape.

I think we should aim to disable as many of Duktape's non-standard builtins as possible.

Thanks again for your help on this; this was a great catch!

PR #347, opened 2019-09-01
'hexdump()' has error when option's length bigger than ArrayBuffer.length

gebing:
'hexdump()' will throw an error when option's length bigger than ArrayBuffer.length.


oleavr:
Nice, thanks! How about using `Math.min()` instead?

gebing:
OK. commit again.

oleavr:
Thanks!

PR #352, opened 2019-09-18
Add support for MemoryAccessMonitor to linux

alvarofe:
Besides adding MemoryAccessMonitor to Linux I also fixed a `FIXME` in the signal handler when reporting memory operation - it was needed for MemoryAccessMonitor to function correctly. Now, it is possible to check if the signal was due to a read or write - using capstone to disassemble the instruction.

Likewise, `gum_linux_unparse_ucontext` was using pc to set THUMB bit. However, on the signal context CPSR already contains that information plus PC will always be even. Therefore, with the previous logic when resuming after handling the exception it will always return to ARM mode instead of the one that correspond - let me know if that breaks something I am not aware. This should also be added to darwin and other ARM backends. Unfortunately, I don't have devices to test but let me know if you want me to send the same fix for darwin.

### DIFF
--- a/gum/backend-linux/gumlinux.h
+++ b/gum/backend-linux/gumlinux.h
@@ -8,18 +8,27 @@
 #define __GUM_LINUX_H__
 
 #include "gumprocess.h"
+#include "gummemory.h"
 
 #include <ucontext.h>
 
 G_BEGIN_DECLS
 
 typedef struct _GumLinuxNamedRange GumLinuxNamedRange;
+typedef struct _GumLinuxRange GumLinuxRange;
 
 struct _GumLinuxNamedRange
 {
   const gchar * name;
   gpointer base;
   gsize size;
+};
+
+struct _GumLinuxRange
+{
+  GumAddress base;
+  gsize size;
+  GumPageProtection prot;
 };
 
 GUM_API GumCpuType gum_linux_cpu_type_from_file (const gchar * path,
@@ -29,6 +38,7 @@ GUM_API void gum_linux_enumerate_modules_using_proc_maps (
     GumFoundModuleFunc func, gpointer user_data);
 GUM_API void gum_linux_enumerate_ranges (pid_t pid, GumPageProtection prot,
     GumFoundRangeFunc func, gpointer user_data);
+GUM_API GList * gum_linux_collect_ranges (void);
 GUM_API GHashTable * gum_linux_collect_named_ranges (void);
 
 GUM_API gboolean gum_linux_module_path_matches (const gchar * path,
--- a/gum/backend-linux/gumlinux.h
+++ b/gum/backend-linux/gumlinux.h
@@ -7,28 +7,28 @@
 #ifndef __GUM_LINUX_H__
 #define __GUM_LINUX_H__
 
-#include "gumprocess.h"
 #include "gummemory.h"
+#include "gumprocess.h"
 
 #include <ucontext.h>
 
 G_BEGIN_DECLS
 
-typedef struct _GumLinuxNamedRange GumLinuxNamedRange;
 typedef struct _GumLinuxRange GumLinuxRange;
-
-struct _GumLinuxNamedRange
-{
-  const gchar * name;
-  gpointer base;
-  gsize size;
-};
+typedef struct _GumLinuxNamedRange GumLinuxNamedRange;
 
 struct _GumLinuxRange
 {
   GumAddress base;
   gsize size;
   GumPageProtection prot;
+};
+
+struct _GumLinuxNamedRange
+{
+  const gchar * name;
+  gpointer base;
+  gsize size;
 };
 
 GUM_API GumCpuType gum_linux_cpu_type_from_file (const gchar * path,
--- a/gum/backend-linux/gummemoryaccessmonitor-linux.c
+++ b/gum/backend-linux/gummemoryaccessmonitor-linux.c
@@ -34,8 +34,7 @@ struct _GumMemoryAccessMonitor
   gint pages_total;
 
   GumPageProtection access_mask;
-  GumPageDetails * pages_details;
-  guint num_pages;
+  GArray * pages;
   gboolean auto_reset;
 
   GumMemoryAccessNotify notify_func;
@@ -72,8 +71,7 @@ static gboolean gum_linux_clear_flag (const GumPageDetails * details,
 static gboolean gum_memory_access_monitor_on_exception (
     GumExceptionDetails * details, gpointer user_data);
 
-G_DEFINE_TYPE (GumMemoryAccessMonitor, gum_memory_access_monitor,
-    G_TYPE_OBJECT)
+G_DEFINE_TYPE (GumMemoryAccessMonitor, gum_memory_access_monitor, G_TYPE_OBJECT)
 
 static void
 gum_memory_access_monitor_class_init (GumMemoryAccessMonitorClass * klass)
@@ -98,10 +96,8 @@ gum_memory_access_monitor_dispose (GObject * object)
   gum_memory_access_monitor_disable (self);
 
   if (self->notify_data_destroy != NULL)
-  {
-    self->notify_data_destroy (self->notify_data);
-    self->notify_data_destroy = NULL;
-  }
+    g_clear_pointer (&self->notify_data, self->notify_data_destroy);
+
   self->notify_data = NULL;
   self->notify_func = NULL;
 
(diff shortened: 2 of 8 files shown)


alvarofe:
Just testing against my application I noticed that duktape runtime does not have support for MemoryAccessMonitor whereas V8 does - although only for windows which I have to change the #ifdef. Is there any reason for that ?

alvarofe:
Working from JS using V8 by now - android arm64
screenshot-2019-09-19-18:21:30

oleavr:
@alvarofe No reason just lack of time to implement it there. (And lack of motivation as the feature was Windows-only anyway.)

alvarofe:
Thinking about this, it should also work on macos - although I should change minor things to get the mapping. With that, both will share a common implementation. So give me some days to advance work and test it on mac osx (I don't have an iphone to test though, you or someone else should take care of that).

Likewise, I will try to add support for duktape runtime.

alvarofe:
Ready to review - lets merge this one along with the others PR I have open (duktape included). Once done, will do mac os support on spare time.

alvarofe:
Let me review this once more to make sure to fix all the coding styles issues so we can save some time while reviewing this.

oleavr:
Thanks! Just let me know when it's ready for another review.

alvarofe:
Ready to review - I tried to address as many coding styles issues as I could but likely I left some. Waiting your comments

gum/backend-posix/gumexceptor-posix.c
```diff
@@ -378,6 +396,236 @@ gum_exceptor_backend_on_signal (int sig,
...
+    case ARM_INS_STRH:
+    case ARM_INS_STRHT:
+    case ARM_INS_STRT:
+      return GUM_MEMOP_WRITE;
+    default:
+      return GUM_MEMOP_READ;
+  }
+}
+
+#endif
```
oleavr: For consistency, ordering of architecture-specific code should be:
- x86/x86_64
- arm
- arm64
- mips

gum/backend-posix/gumexceptor-posix.c
```diff
@@ -378,6 +396,236 @@ gum_exceptor_backend_on_signal (int sig,
...
+  return op;
+}
+
+static cs_insn *
+disassemble_instruction_at (gconstpointer address, 
+                            GumCpuContext * cpu)
+{
+  cs_err err;
+  csh capstone;
+  cs_insn * insn = NULL;
```
oleavr: The variable holding the return value should be first. (Others should be chronologically sorted.)

gum/backend-posix/gumexceptor-posix.c
```diff
@@ -335,7 +353,7 @@ gum_exceptor_backend_on_signal (int sig,
       if (siginfo->si_addr == ed.address)
         md->operation = GUM_MEMOP_EXECUTE;
       else
-        md->operation = GUM_MEMOP_READ; /* FIXME */
+        md->operation = gum_exceptor_check_mem_access (ed.address, cpu_context);
```
oleavr: How about naming it `gum_exceptor_infer_memory_operation`.

gum/backend-posix/gumexceptor-posix.c
```diff
@@ -50,6 +53,21 @@ static int gum_exceptor_backend_replacement_sigaction (int sig,
...
+    GumCpuContext * cpu);
+
+#if defined (HAVE_I386)
+static GumMemoryOperation gum_exceptor_x86_check_mem_operation (cs_insn * insn);
+#elif defined (HAVE_ARM64)
+static GumMemoryOperation gum_exceptor_arm64_check_mem_operation (
+    cs_insn * insn);
+#elif defined (HAVE_ARM)
+static GumMemoryOperation gum_exceptor_arm_check_mem_operation (cs_insn * insn);
+#endif
```
oleavr: See below comment on ordering.

gum/backend-posix/gumexceptor-posix.c
```diff
@@ -17,13 +17,16 @@
...
 
+#include <capstone.h>
 #include <signal.h>
 #include <stdlib.h>
 #ifdef HAVE_QNX
 # include <sys/debug.h>
 # include <unix.h>
 #endif
 
+#define X86OP(n) insn->detail->x86.operands[n]
```
oleavr: Defines should be namespaced, i.e. `GUM_X86OP`. Macros like this are typically placed next to where they're used. But in this case this is only used twice, so ISTM we don't really need a macro.

gum/backend-linux/gumlinux.h
```diff
@@ -8,12 +8,14 @@
...
 
 #include "gumprocess.h"
+#include "gummemory.h"
 
 #include <ucontext.h>
 
 G_BEGIN_DECLS
 
 typedef struct _GumLinuxNamedRange GumLinuxNamedRange;
+typedef struct _GumLinuxRange GumLinuxRange;
```
oleavr: Because the function is before the named one, the struct should also be.

gum/backend-linux/gumlinux.h
```diff
@@ -22,13 +24,21 @@ struct _GumLinuxNamedRange
   gsize size;
 };
 
+struct _GumLinuxRange
+{
+  GumAddress base;
+  gsize size;
+  GumPageProtection prot;
+};
```
oleavr: Here as well.

gum/backend-linux/gummemoryaccessmonitor-linux.c
```diff
@@ -0,0 +1,464 @@
...
+static void gum_linux_free_ranges (gpointer data);
+static gboolean gum_linux_set_flag (const GumPageDetails * details,
+    gpointer user_data);
+static gboolean gum_linux_clear_flag (const GumPageDetails * details,
+    gpointer user_data);
+static gboolean gum_memory_access_monitor_on_exception (
+    GumExceptionDetails * details, gpointer user_data);
+
+G_DEFINE_TYPE (GumMemoryAccessMonitor, gum_memory_access_monitor,
+    G_TYPE_OBJECT)
```
oleavr: This would fit on one line (80 cols).

gum/backend-linux/gummemoryaccessmonitor-linux.c
```diff
@@ -0,0 +1,464 @@
...
+  gum_mprotect (GSIZE_TO_POINTER (range->base), range->size, new_prot);
+
+  return TRUE;
+}
+
+static gboolean
+gum_linux_clear_flag (const GumPageDetails * details,
+                      gpointer user_data)
+{
+  guint i;
```
oleavr: Variables should be ordered chronologically (except for the one holding the return value, which should be first).

gum/backend-linux/gummemoryaccessmonitor-linux.c
```diff
@@ -0,0 +1,464 @@
...
+  {
+    GumPageDetails * page = &self->pages_details[i];
+    const GumLinuxRange * range = &page->range;
+    const GumMemoryRange * r = &self->ranges[page->range_index];
+    guint operation_mask;
+    guint operations_reported;
+    guint pages_remaining;
+
+    if ((GPOINTER_TO_SIZE (d.address) >= range->base) &&
+        GPOINTER_TO_SIZE (d.address) < (range->base + self->page_size))
```
oleavr:
```suggestion
        GPOINTER_TO_SIZE (d.address) < range->base + self->page_size)
```

gum/backend-linux/gummemoryaccessmonitor-linux.c
```diff
@@ -0,0 +1,464 @@
...
+          break;
+        case GUM_MEMOP_EXECUTE:
+          if ((gum_orig_prot & GUM_PAGE_WRITE) == 0)
+            return FALSE;
+          break;
+        default:
+          g_assert_not_reached ();
+      }
+
+      /* Restore original flags */
```
oleavr: I think this is obvious, so the comment is redundant.

gum/backend-linux/gummemoryaccessmonitor-linux.c
```diff
@@ -0,0 +1,464 @@
...
+      if (self->auto_reset)
+        gum_mprotect (GSIZE_TO_POINTER (range->base), range->size, range->prot);
+
+      operation_mask = 1 << d.operation;
+      operations_reported = g_atomic_int_or (&page->completed, operation_mask);
+      if ((operations_reported != 0) && self->auto_reset)
```
oleavr:
```suggestion
      if (operations_reported != 0 && self->auto_reset)
```

alvarofe:
All the comments should have been fixed. Ready for the second round

oleavr:
Thanks! 🤘

Landed in c227af0, which also makes it generic so it works on any UNIX system. (In theory – only tested on macOS and tests pass, albeit slow – probably due to it walking each page of every range and there's a ton of them.)

PR #353, opened 2019-09-18
Handle multiple devices and linux host test for Android

alvarofe:


oleavr:
Closing this as I think we need to take a step back and refactor these scripts as a whole.

PR #354, opened 2019-09-22
fix android 10 linker issue

Alien-AV:
lookup of linker location happened with hardcoded paths(, which change every few versions?).
changed the lookup to be based on file name only and not full path

### DIFF
--- a/gum/backend-linux/gumandroid.c
+++ b/gum/backend-linux/gumandroid.c
@@ -650,7 +650,7 @@ gum_try_parse_linker_proc_maps_line (const gchar * line,
   gchar perms[5] = { 0, };
   gchar path[PATH_MAX];
   gint n;
-  const gchar * new_path, * old_path, * linker_path;
+  const gchar * suffix, * linker_path;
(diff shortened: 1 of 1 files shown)


Alien-AV:
Meanwhile I verified only that frida-ps works.

It seems that ADB remote session disconnects when starting the frida server. (Maybe phone reboots?)

In any case, it works better with this fix than without it.

oleavr:
Thanks for looking into this! Unfortunately this fix may introduce false positives, if an app happens to have mapped a file suffixed with `linker` / `linker64`. So I think we need to take a step back and rethink how we locate the linker.

Alien-AV:
How high are the chances of that? We can push it now, as it fixes an actual and current issue.
And if someone ever hits the issue of "mapped a file suffixed with `linker` (and it's matched before the actual linker because it's closer to vdso)", then in a different issue we can take a step back and rethink.

oleavr:
I think we can go with the solution suggested [here](https://github.com/frida/frida/issues/1026#issuecomment-539111292).

(Sorry for the delay, I've been heads down in code for a while.)

PR #360, opened 2019-10-10
Make frida-gum linker seeking future-proof on Android

strazzere:
Per https://github.com/frida/frida/issues/1026 and https://github.com/frida/frida-gum/pull/354 -
I looked into how the linker is meant to be established in Android going forward. According to
https://source.android.com/devices/architecture/modular-system/runtime#bionic-changes - Google
intends to maintain both `/system/bin/linker` and `/system/bin/linker64` as a symlink, pointing
to what the real file will be.

This means if we look for those files and they are symlinks, the linker should be mapped to memory
using the files realpath. So for future changes, looking to see what the symlink resolves as should
make this work in all future version (and keep its backwards compatibility).

Tested this on Android 5, 7.2, 8.0.1, 9 and 10.

### DIFF
--- a/gum/backend-linux/gumandroid.c
+++ b/gum/backend-linux/gumandroid.c
@@ -13,7 +13,9 @@
 #include <link.h>
 #include <pthread.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <sys/system_properties.h>
+#include <unistd.h>
 
 #if (defined (HAVE_I386) && GLIB_SIZEOF_VOID_P == 4) || defined (HAVE_ARM)
 # define GUM_ANDROID_LEGACY_SOINFO 1
@@ -650,7 +652,9 @@ gum_try_parse_linker_proc_maps_line (const gchar * line,
   gchar perms[5] = { 0, };
   gchar path[PATH_MAX];
   gint n;
-  const gchar * new_path, * old_path, * linker_path;
+  struct stat linker_stat;
+  gchar linker_true_path[PATH_MAX];
+  const gchar * linker_path, *linker_expected_path;
   const guint8 elf_magic[] = { 0x7f, 'E', 'L', 'F' };
 
   n = sscanf (line,
@@ -666,19 +670,32 @@ gum_try_parse_linker_proc_maps_line (const gchar * line,
 
   if (sizeof (gpointer) == 4)
   {
-    new_path = "/apex/com.android.runtime/bin/linker";
-    old_path = "/system/bin/linker";
+    linker_expected_path = "/system/bin/linker";
   }
   else
   {
-    new_path = "/apex/com.android.runtime/bin/linker64";
-    old_path = "/system/bin/linker64";
+    linker_expected_path = "/system/bin/linker64";
   }
 
-  if (strcmp (path, new_path) == 0)
-    linker_path = new_path;
-  else if (strcmp (path, old_path) == 0)
-    linker_path = old_path;
+  // If there are no linker files, fail
+  if(lstat(linker_expected_path, &linker_stat) == -1) {
+    return FALSE;
+  }
+
+  if (S_ISLNK(linker_stat.st_mode)) {
+    if(readlink(linker_expected_path, linker_true_path, sizeof(linker_true_path)) < 0) {
+      // Issue reading the link, fall back to backwards compatible path
+      // linker_true_path = &linker_expected_path[0];
+      strcpy(linker_true_path, linker_expected_path);
+    }
+  } else {
+    // Expected path is a real file, use it
+    // linker_true_path = &linker_expected_path[0];
+      strcpy(linker_true_path, linker_expected_path);
+  }
+
+  if (strcmp (path, linker_true_path) == 0)
+    linker_path = linker_true_path;
   else
     return FALSE;
 
--- a/gum/backend-linux/gumandroid.c
+++ b/gum/backend-linux/gumandroid.c
@@ -639,7 +639,6 @@ no_vdso:
 
 beach:
   g_strfreev (lines);
-  g_free (linker_path);
   g_free (maps);
 
   return result;
@@ -651,41 +650,26 @@ gum_find_linker_path ()
   gchar *linker_expected_path, *linker_path = NULL;
 
   if (sizeof (gpointer) == 4)
-  {
     linker_expected_path = "/system/bin/linker";
-  }
   else
-  {
     linker_expected_path = "/system/bin/linker64";
-  }
 
   // If there are no linker file, fail
   if (!g_file_test (linker_expected_path, G_FILE_TEST_EXISTS))
-  {
-    goto result;
-  }
+    return NULL;
 
   if (g_file_test (linker_expected_path, G_FILE_TEST_IS_SYMLINK))
   {
     linker_path = g_file_read_link (linker_expected_path, NULL);
-    if (linker_path == NULL)
-    {
-      // Issue reading the link, fall back to backwards compatible path
-      // linker_true_path = &linker_expected_path[0];
-      strcpy (linker_path, linker_expected_path);
-    }
-  }
-  else
-  {
-    // Expected path is a real file, use it
-    // linker_true_path = &linker_expected_path[0];
-      strcpy (linker_path, linker_expected_path);
+
+    // Issue reading the link, fall back to backwards compatible path
+    if (linker_path != NULL)
+      return linker_path;
   }
 
-  result:
-  g_free (linker_expected_path);
-
-  return linker_path;
+  // If file was not a symlink or symlink resolution doesn't work, fall back
+  // to expected file
+  return linker_expected_path;
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


gum/backend-linux/gumandroid.c
```diff
@@ -666,19 +670,32 @@ gum_try_parse_linker_proc_maps_line (const gchar * line,
...
-    old_path = "/system/bin/linker64";
+    linker_expected_path = "/system/bin/linker64";
   }
 
-  if (strcmp (path, new_path) == 0)
-    linker_path = new_path;
-  else if (strcmp (path, old_path) == 0)
-    linker_path = old_path;
+  // If there are no linker files, fail
+  if(lstat(linker_expected_path, &linker_stat) == -1) {
```
oleavr: Style should look like this:

```suggestion
  if (lstat (linker_expected_path, &linker_stat) == -1)
  {
```

Though in case of a single statement – unless the statement needs to be split across multiple lines – curly braces should be omitted.

(Sorry for not documenting it anywhere yet.)

oleavr: However, it would be good to use GLib's [file utilities](https://developer.gnome.org/glib/stable/glib-File-Utilities.html) here:

```C
  if (g_file_test (linker_expected_path, G_FILE_TEST_IS_SYMLINK))
  {
    …
  }
```

And [g_file_read_link()](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-read-link) to read the link. (Return value needs to be `g_free()`d.)

gum/backend-linux/gumandroid.c
```diff
@@ -666,19 +670,32 @@ gum_try_parse_linker_proc_maps_line (const gchar * line,
...
+    if(readlink(linker_expected_path, linker_true_path, sizeof(linker_true_path)) < 0) {
+      // Issue reading the link, fall back to backwards compatible path
+      // linker_true_path = &linker_expected_path[0];
+      strcpy(linker_true_path, linker_expected_path);
+    }
+  } else {
+    // Expected path is a real file, use it
+    // linker_true_path = &linker_expected_path[0];
+      strcpy(linker_true_path, linker_expected_path);
+  }
```
oleavr: Now that we need to do some I/O inside of this function, which is called for each line of `/proc/self/maps`, we should move this logic outside the function and pass it in instead.

gum/backend-linux/gumandroid.c
```diff
@@ -13,7 +13,9 @@
 #include <link.h>
 #include <pthread.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <sys/system_properties.h>
+#include <unistd.h>
```
oleavr: These can be omitted when using the GLib APIs.

gum/backend-linux/gumandroid.c
strazzere: Makes sense, will switch to using glib's file utils and fix the styling.

gum/backend-linux/gumandroid.c
strazzere: Yup - I sort of thought about this, but didn't bother in the moment. Will figure this one out.

strazzere:
I believe I've fixed everything and put it into the normal frida styling.

strazzere:
Erm... Do not merge yet please, I need to do a few more tests to validate things.

strazzere:
This should be in a good state for a second pass by you @oleavr when you have a moment or two.

gum/backend-linux/gumandroid.c
```diff
@@ -641,16 +644,44 @@ gum_try_init_linker_details (void)
...
+
+    // Issue reading the link, fall back to backwards compatible path
+    if (linker_path != NULL)
+      return linker_path;
+  }
+
+  // If file was not a symlink or symlink resolution doesn't work, fall back
+  // to expected file
+  return linker_expected_path;
+}
```
oleavr: How about something like:

```C
static gchar *
gum_find_linker_path (void)
{
  gchar * result = NULL;
  const gchar * traditional_path;

  traditional_path = (sizeof (gpointer) == 4)
      ? "/system/bin/linker"
      : "/system/bin/linker64";

  if (g_file_test (traditional_path, G_FILE_TEST_IS_SYMLINK))
    result = g_file_read_link (traditional_path, NULL);

  if (result == NULL)
    result = g_strdup (traditional_path);

  return result;
}
```

Basically:
- Always return a newly allocated string, not just in the symlink-case.
- Never return `NULL`, as it's not clear what we should do when the file doesn't exist (or the underlying `stat` fails so it looks like it doesn't) – so perhaps it's better to not deal with it here until we need a fallback strategy for that case.
- Shorten the variable names as the function now gives them context. (Now that they're no longer part of a larger function.)

gum/backend-linux/gumandroid.c
```diff
@@ -571,9 +572,11 @@ static const GumModuleDetails *
 gum_try_init_linker_details (void)
 {
   const GumModuleDetails * result = NULL;
-  gchar * maps, ** lines;
+  gchar * maps, ** lines, *linker_path;
```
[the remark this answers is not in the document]
strazzere: Meaning, you would prefer both `** lines` and `* linker_path` to be place before `* maps`?

gum/backend-linux/gumandroid.c
strazzere: Works for me.

gum/backend-linux/gumandroid.c
```diff
@@ -571,9 +572,11 @@ static const GumModuleDetails *
...
   gint num_lines, vdso_index, i;
 
+  linker_path = gum_find_linker_path ();
```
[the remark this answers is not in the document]
strazzere: I'm putting it above the potential `goto no_vdso;` otherwise it potentially may try to perform a `g_free(linker_path)` on an uninitialized variable.

gum/backend-linux/gumandroid.c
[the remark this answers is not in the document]
oleavr: Ah no, sorry, “based on first use” should have been “based on the order in which they're referenced” (including assignments). I.e.:

```
  const GumModuleDetails * result = NULL;
  gchar * maps, ** lines;
  gint num_lines, vdso_index, i;
  gchar * linker_path;
```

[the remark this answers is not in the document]
oleavr: Update: Also initialize it to `NULL` so the `g_free ()` can be done regardless at `beach`.

gum/backend-linux/gumandroid.c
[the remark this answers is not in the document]
oleavr: It's safe to call `g_free ()` on a `NULL` value, so it's better to initialize it to `NULL`. (It's the convention used by GLib.)

gum/backend-linux/gumandroid.c
```diff
@@ -572,11 +572,9 @@ static const GumModuleDetails *
 gum_try_init_linker_details (void)
 {
   const GumModuleDetails * result = NULL;
-  gchar * maps, ** lines, *linker_path;
+  gchar ** lines, * linker_path, * maps;
   gint num_lines, vdso_index, i;
```
oleavr:
```
  gchar * maps, ** lines;
  gint num_lines, vdso_index, i;
  gchar * linker_path = NULL;
```

strazzere:
/me cringes

I think it's ready? :)

gum/backend-linux/gumandroid.c
```diff
@@ -321,8 +321,9 @@ struct _GumFunctionSignature
 };
 
 static const GumModuleDetails * gum_try_init_linker_details (void);
+static gchar * gum_find_linker_path (void);
 static gboolean gum_try_parse_linker_proc_maps_line (const gchar * line,
-    GumModuleDetails * module, GumMemoryRange * range);
+    const gchar * linker_path, GumModuleDetails * module, GumMemoryRange * range);
```
oleavr:
```suggestion
    const gchar * linker_path, GumModuleDetails * module,
    GumMemoryRange * range);
```

(Overflows 80 cols by two spaces, so we need to break it.)

gum/backend-linux/gumandroid.c
```diff
@@ -598,13 +600,16 @@ gum_try_init_linker_details (void)
       break;
     }
   }
+
```
oleavr:
```suggestion
```

(To keep it part of the same "operation".)

gum/backend-linux/gumandroid.c
strazzere: This is already fixed in the latest commit (I believe, this view is getting cluttered for me)

gum/backend-linux/gumandroid.c
strazzere: Inadvertent when I was moving it around, good catch.

gum/backend-linux/gumandroid.c
oleavr: Ah yes sorry, my bad.

gum/backend-linux/gumandroid.c
strazzere: Hah, you should post your linter config :grin:

strazzere:
I think it's all done? `+35 −24` the review has now eclipsed the code changes :laughing: 

You should definitely post your linter config, but also I need to skim the code more just to get a feel for it :smirk:

strazzere:
:raised_hands: :raised_hands: :raised_hands:

oleavr:
Yay! 💥🙌

> the review has now eclipsed the code changes :laughing: 

😂

> You should definitely post your linter config

Unfortunately I don't have any – I'm a biological linter 😅

PR #361, opened 2019-10-13
Fix linker lookup in Android

muhzii:
That g_free call shouldn't be here as `linker_path` is being accessed later from `GumModuleDetails`
Fixes: https://github.com/frida/frida/issues/1046


oleavr:
Oops, good catch. Thanks a lot!

This solution would unfortunately replace this use-after-free with a memory leak, so I have implemented a slightly different solution in 1f97c98.

PR #377, opened 2019-11-27
Fix rounding issue with time.

Manouchehri:


oleavr:
Thanks, good catch! Fixed in e6a590f807b9b99fd05cd2e90b9379cd6d1fe601, which also fixes a similar flaw in `MonotonicallyIncreasingTime()`.

PR #379, opened 2019-12-04
Adds size support for ELF symbols.

Manouchehri:
Not sure if this breaks building on other platforms, but seems to work on Linux.

### DIFF
--- a/bindings/gumjs/gumv8module.cpp
+++ b/bindings/gumjs/gumv8module.cpp
@@ -476,6 +476,7 @@ gum_emit_symbol (const GumSymbolDetails * details,
 
   _gum_v8_object_set_ascii (symbol, "name", details->name, core);
   _gum_v8_object_set_pointer (symbol, "address", details->address, core);
+  _gum_v8_object_set_uint(symbol, "size", details->size, core);
 
   return mc->OnMatch (symbol);
 }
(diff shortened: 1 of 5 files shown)


jmis1337:
This would be very useful to port jemalloc parser to frida.

bindings/gumjs/gumv8module.cpp
```diff
@@ -476,6 +476,7 @@ gum_emit_symbol (const GumSymbolDetails * details,
 
   _gum_v8_object_set_ascii (symbol, "name", details->name, core);
   _gum_v8_object_set_pointer (symbol, "address", details->address, core);
+  _gum_v8_object_set_uint(symbol, "size", details->size, core);
```
oleavr:
```suggestion
  _gum_v8_object_set_uint (symbol, "size", details->size, core);
```

gum/backend-elf/gumelfmodule.h
```diff
@@ -79,6 +79,7 @@ struct _GumElfSymbolDetails
   GumElfSymbolType type;
   GumElfSymbolBind bind;
   GumElfSectionHeaderIndex section_header_index;
+  gsize size;
```
oleavr: Let's put this right after `address`.

oleavr:
Actually, scratch that. Doesn't seem to be available on Darwin, at least not without parsing the DWARF. We can make this a Linux/ELF-only feature initially.

oleavr:
Will take a quick stab at landing this.

oleavr:
Landed in 2da7de17a125ca2881e795f4211d6b5da42eeb82. Thanks!

Issue #116, opened 2015-10-16
Thread enumeration does not work on Android


stevielavern:
When attached to a process on Android, thread list is always empty.
For instance, with the contact application:

```
$ frida -U com.android.contacts
...
[USB::com.android.contacts]-> Process.enumerateThreadsSync()
[]
```

When gdb, attached to the same process, shows the following:

```
 (gdb) info threads
  Id   Target Id    
  23   Thread 9967  
  22   Thread 9948  
  21   Thread 9946  
  20   Thread 9945  
  19   Thread 9944  
  18   Thread 9940  
  17   Thread 9165  
  16   Thread 9158  
  15   Thread 9156  
  14   Thread 9155  
  13   Thread 9154  
  12   Thread 9153  
  11   Thread 9152  
  10   Thread 9150  
  9    Thread 9149  
  8    Thread 9148  
  7    Thread 9147  
  6    Thread 9146  
  5    Thread 9145  
  4    Thread 9144  
  3    Thread 9143  
  2    Thread 9142  
* 1    Thread 9138  
```

Precision: same behavior on Android 4.4 and Android 5.0

oleavr:
Oops! Thanks for the heads up.

oleavr:
Should be fixed in latest Frida 9.x.

PR #382, opened 2019-12-06
Fix thread enumaration issue on Linux when using Yama

jonwilson030981:
With Yama LSM enabled processes by default can only ptrace their children (unless otherwise configured by sysctl). This appears to be the default setting on Ubuntu Linux. More aggressive changes may be required for other modes.

This patch uses prctl(PR_SET_PTRACER) to give permission to the ptracing thread (in another process) to attach.

- https://www.kernel.org/doc/Documentation/security/Yama.txt
- https://github.com/torvalds/linux/blob/master/security/yama/yama_lsm.c

Note also that Process.enumerateThreads() doesn't raise any exceptions on failure, instead simply returning an empty or reduced set. This may help to identify the cause if subsequent issues are raised. The following issues possibly relate:

- https://github.com/frida/frida/issues/489
- https://github.com/frida/frida-gum/issues/116

### DIFF
--- a/gum/backend-linux/gumprocess-linux.c
+++ b/gum/backend-linux/gumprocess-linux.c
@@ -103,7 +103,8 @@ typedef gint (* GumCloneFunc) (gpointer arg);
 
 enum _GumModifyThreadAck
 {
-  GUM_ACK_ATTACHED = 1,
+  GUM_ACK_PTRACER = 1,
+  GUM_ACK_ATTACHED,
   GUM_ACK_STOPPED,
   GUM_ACK_READ_CONTEXT,
   GUM_ACK_MODIFIED_CONTEXT,
@@ -434,6 +435,9 @@ gum_process_modify_thread (GumThreadId thread_id,
         NULL);
     g_assert (child > 0);
 
+    prctl (PR_SET_PTRACER, child);
+    gum_put_ack (fd, GUM_ACK_PTRACER);
+
     if (gum_await_ack (fd, GUM_ACK_ATTACHED))
     {
       GumThreadState state;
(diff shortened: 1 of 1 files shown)


gum/backend-linux/gumprocess-linux.c
```diff
@@ -434,6 +435,10 @@ gum_process_modify_thread (GumThreadId thread_id,
         NULL);
     g_assert (child > 0);
 
+    prctl(PR_SET_PTRACER, child);
```
oleavr:
```suggestion
    prctl (PR_SET_PTRACER, child);
```

gum/backend-linux/gumprocess-linux.c
```diff
@@ -434,6 +435,10 @@ gum_process_modify_thread (GumThreadId thread_id,
...
+
+    gum_put_ack(fd, GUM_ACK_PTRACER);
```
oleavr:
```suggestion
    gum_put_ack (fd, GUM_ACK_PTRACER);
```

gum/backend-linux/gumprocess-linux.c
```diff
@@ -434,6 +435,10 @@ gum_process_modify_thread (GumThreadId thread_id,
         NULL);
     g_assert (child > 0);
 
+    prctl(PR_SET_PTRACER, child);
+
```
oleavr:
```suggestion
```

jonwilson030981:
Sorry, old coding style habits die hard! Hope this is better and by forcing an amended commit I don't make the merge hard!

oleavr:
No worries! Thanks for digging into and fixing this 🤘

Issue #383, opened 2019-12-09
a crash report on android(cased by bne.n instrument)


bigboysun:
I want hook a function,it‘s code is(12bytes head of function):

294e38:	6dc0      	ldr	r0, [r0, #92]	; 0x5c
294e3a:	2800      	cmp	r0, #0
294e3c:	bf18      	it	ne
294e3e:	f03b b88f 	bne.w	2cff60 
294e42:	4770      	bx	lr

It generated a trampoline (on_invoke_trampoline),  code is:
0xf3b3911c:	ldr	r0, [r0, #92]	; 0x5c
   0xf3b3911e:	cmp	r0, #0
   0xf3b39120:	it	ne
=> 0xf3b39122:	bne.n	0xf3b39126	; unpredictable <IT:ne>
   0xf3b39124:	b.n	0xf3b39130
   0xf3b39126:	push	{r0}
   0xf3b39128:	push	{r0}
   0xf3b3912a:	ldr	r0, [pc, #12]	; (0xf3b39138)
   0xf3b3912c:	str	r0, [sp, #4]
   0xf3b3912e:	pop	{r0, pc}

it crash at ==>,  crash report this “bne.n” is Illegal instrument

report：

signal 4 (SIGILL), code -6 (SI_TKILL), fault addr 0xdbc4e222 
I/AEE/AED (  484):     r0 c7c2d3f8  r1 00000010  r2 00000010  r3 ca50dab0
I/AEE/AED (  484):     r4 efe0d300  r5 000334b5  r6 000334b4  r7 ca50dad8
I/AEE/AED (  484):     r8 ca50daac  r9 00018f3c  sl d6c4226b  fp d6d40a71
I/AEE/AED (  484):     ip d59f6e39  sp ca50da68  lr dbc4e209  pc dbc4e222  cpsr a00f1830
I/AEE/AED (  484): 
I/AEE/AED (  484): backtrace:
I/AEE/AED (  484):     #00 pc 00000222  <unknown>
I/AEE/AED (  484):     #01 pc 00000205  <unknown>

bigboysun:
b instrment with condition can't follow it instrument!!

bigboysun:
will any body fix? or, i do

oleavr:
Please submit a PR ❤️ This should be handled by ThumbRelocator in frida-gum.

bigboysun:
This is an important patch. Many function headers will have IT instruction.

oleavr:
[...]

PR #384, opened 2019-12-10
Bigboysun patch 1

bigboysun:
fixed https://github.com/frida/frida-gum/issues/383
Has been tested


bigboysun:
The situation is far more complicated than expected. The jump of the IT instruction only allows one instruction, and the remote jump is not allowed. It needs to be modified.

oleavr:
Thanks for looking into this! I believe we already have some IT handling in our relocator, so it sounds like there's something missing there.

Issue #320, opened 2019-01-12
Provide an api to get raw data pointer from ArrayBuffer?


ChiChou:
Some api only returns ArrayBuffer (like `frida-fs`).
When I have to pass them to ffi, looks like the best way is

```
const p = Memory.alloc(bytes.byteLength);
Memory.writeByteArray(p, bytes);
```

That is an unnecessary extra `memcpy`.

jmis1337:
can't you use type arrays like this?

new Uint8Array(bytes)[0]=0xff

ChiChou:
> can't you use type arrays like this?
> 
> new Uint8Array(bytes)[0]=0xff

You basically have no idea what I was talking about

jmis1337:
nope.  but I guess the ticket was not clear.  It looked like you are trying to read from an ArrayBuffer. That's how you read/write to an array buffer. 
Just trying to help...

Manouchehri:
@ChiChou Hmm, would adding a Memory.sharedByteArray() that returns a SharedArrayBuffer work for your use case?

ChiChou:
@Manouchehri Thanks a lot!

ChiChou:
@Manouchehri Oh wait. This is not what I mean. 

I know nothing about internal implementation of v8 and Duktape but I guess there must be a vector in the backing store of an ArrayBuffer (otherwise I'll submit a PR). What I need is to get the **raw** pointer of this continuous storage. If that's not clear, I mean something like `std::string::data()` in C++

oleavr:
@ChiChou That's exactly what I implemented in a1034d38f7cc4e9716a1b3b1a93c43f5a3cf0135. You can now pass an `ArrayBuffer` to `NativeFunction` for any `'pointer'` argument.

I also landed @Manouchehri's PR in eec44a1bf4ef6dee7397ba866674581a044e0d29, which is a different thing: when you want to wrap an `ArrayBuffer` around a region of memory to easily peek/poke at it without going through read*()/write*() APIs for each access.

oleavr:
@ChiChou But thinking more about it, maybe we should just add a function to get the underlying pointer, and revert `a1034d38f7cc4e9716a1b3b1a93c43f5a3cf0135`. It's a bit more dangerous, but it's also more explicit... What do you think? Any good names come to mind?

oleavr:
A few ideas:
- `ptr(buf)`
- `Memory.unwrap(buf)`

ChiChou:
[...]

PR #385, opened 2019-12-10
Add NativePointer#mapByteArray()

Manouchehri:
Useful if you want to avoid needing to constantly use .writeByteArray(...). Might solve #320.


Manouchehri:
@oleavr Could you do the final touches to make it a part of the NativePointer API? =)

PR #394, opened 2020-01-07
Fix linker path lookup on Android 10 when inside a sandbox.

### LINKED ISSUE #393: #360 breaks attaching to Chrome sandbox
I think it's a seccomp issue, but for some reason attaching to a sandboxed Chrome process no longer works.

```
frida -U -n "com.android.chrome:sandboxed*" --runtime=v8 
```

```
5320  5376 E linker  : readlink("/proc/self/fd/54") failed: Operation not permitted [fd=54]
5320  5376 E linker  : warning: unable to get realpath for the library "/data/local/tmp/re.frida.server/frida-agent-32.so". Will use given path.
5320  5376 F Frida   : Unable to locate the Android linker; please file a bug
5378  5378 E chromium: [0107/005740.532352:ERROR:exception_handler_server.cc(102)] capget: Invalid argument (22)
5378  5378 E chromium: [0107/005740.535704:ERROR:process_memory_linux.cc(42)] open: No such file or directory (2)
```
After git revert 74cca04 1f97c98 2de84803f918e86b73b4b8234dfad1fbed62357b , I'm able to attach again.

Manouchehri:
Fixes #393. Not the prettiest but it works and I need that. =P

### DIFF
--- a/gum/backend-linux/gumandroid.c
+++ b/gum/backend-linux/gumandroid.c
@@ -677,6 +677,15 @@ gum_find_linker_path (void)
   if (g_file_test (traditional_path, G_FILE_TEST_IS_SYMLINK))
     result = g_file_read_link (traditional_path, NULL);
 
+  // Android 10 workaround
+  if (result == NULL) {
+    traditional_path = (sizeof (gpointer) == 4)
+        ? "/apex/com.android.runtime/bin/linker"
+        : "/apex/com.android.runtime/bin/linker64";
+
+    result = g_file_read_link (traditional_path, NULL);
+  }
+
   if (result == NULL)
     result = g_strdup (traditional_path);
 


gum/backend-linux/gumandroid.c
```diff
@@ -677,6 +677,15 @@ gum_find_linker_path (void)
   if (g_file_test (traditional_path, G_FILE_TEST_IS_SYMLINK))
     result = g_file_read_link (traditional_path, NULL);
 
+  // Android 10 workaround
+  if (result == NULL) {
+    traditional_path = (sizeof (gpointer) == 4)
+        ? "/apex/com.android.runtime/bin/linker"
+        : "/apex/com.android.runtime/bin/linker64";
+
+    result = g_file_read_link (traditional_path, NULL);
```
oleavr: Are those links though?

Manouchehri: Nope, but it somehow works. ;p

oleavr:
Hmm, I'm leaning towards just ditching all the cleverness and going for something stupid: pick either new or old hard-coded path based on Android API level.

Manouchehri:
I have no real preference as long as it works. =)

oleavr:
My mind is in a different context so I'm not able to think about a good solution here, but this is a regression so I don't want it to linger and bite users. Reverted to the proposed solution in 3ca8cd3. Sorry @strazzere - hopefully we'll come up with something better once we've had some time to chew on this one :-)

PR #399, opened 2020-03-02
Handle nested signals

bannsec:
Fixed the issue here: https://github.com/frida/frida/issues/1193

Short is that by default linux won't allow handling of the same signal inside itself. I.e.: segfault handling inside segfault handling.

NOTE: The way I have to test this involves sending a signal, which kills the process after being caught. The test passes but based on the way the tests for frida-gum work, any test after it fails due to the process crashing.

Not sure how to handle that, since it's not really a failed test, just how the harness handles tests.

### DIFF
--- a/gum/backend-posix/gumexceptor-posix.c
+++ b/gum/backend-posix/gumexceptor-posix.c
@@ -209,7 +209,7 @@ gum_exceptor_backend_attach (GumExceptorBackend * self)
 
   action.sa_sigaction = gum_exceptor_backend_on_signal;
   sigemptyset (&action.sa_mask);
-  action.sa_flags = SA_SIGINFO;
+  action.sa_flags = SA_SIGINFO | SA_NODEFER;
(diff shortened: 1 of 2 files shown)


oleavr:
Nice! I think one way to test this is to allocate a page of memory (look for `gum_alloc_n_pages` in other tests) that's initially read-only, and have the test-code try to write to that. Then in the handler it can be flip the page to read-write and `return true`.

bannsec:
Ok. Updated it to use access exception as you mentioned and it seems to pass now without segfaulting out.

bannsec:
@oleavr -- just rebased my changes on master. Anything else you need?

oleavr:
@bannsec Thanks! Nope, will get to this soon. Nearly out of the iOS rabbithole I've been in these last few weeks.

oleavr:
Landed with a few tiny tweaks as f9d71b163b1460fca3d8f2624045d1567810ea6d. Thanks! :raised_hands:

PR #417, opened 2020-04-17
Improve address resolution in the libdwarf backend

WorksButNotTested:
GUM previously relied upon 'dladdr' to find symbol names when trying to make a debug symbol from an address. This only works when symbols are in the .dynamic section. It will not incorporate those in the .symtab section.

The dwarf engine already walks the .symtab and .dynamic section to build a hashtable indexed on names to lookup addresses.This patch builds a hashtable of the reverse, but only for the .symtab section (we can rely on the existing implementation working for .dynamic). We can then use this hash table to do the reverse lookup.

This issue will not affect binaries built with -rdynamic, and may also not affect those built with -pie. Obviously, this doesn't affect shared objects by their nature.

### DIFF
--- a/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
+++ b/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
@@ -634,7 +634,7 @@ static gboolean
 gum_collect_symbol_if_function (const GumElfSymbolDetails * details,
                                 gpointer user_data)
 {
-  gboolean include_symbols = (gboolean) user_data;
+  gboolean include_symbols = GPOINTER_TO_INT (user_data);
   const gchar * name;
   gpointer address;
   GArray * addresses;
--- a/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
+++ b/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
@@ -363,7 +363,7 @@ gum_find_nearest_symbol_by_address (gpointer address,
   GHashTable * table;
   gchar * name;
   GList * keys, * cur;
-  gpointer current_address, closest_address;
+  gpointer current_address, nearest_address;
 
   table = gum_get_address_symbols ();
 
@@ -375,31 +375,31 @@ gum_find_nearest_symbol_by_address (gpointer address,
     return TRUE;
   }
 
-  closest_address = NULL;
+  nearest_address = NULL;
   keys = g_hash_table_get_keys (table);
   for (cur = keys; cur != NULL; cur = cur->next)
   {
     current_address = cur->data;
-    if (closest_address == NULL)
+    if (nearest_address == NULL)
     {
-      closest_address = current_address;
+      nearest_address = current_address;
       continue;
     }
 
     if (address < current_address)
       continue;
 
-    if (address - current_address >= address - closest_address)
+    if (address - current_address >= address - nearest_address)
       continue;
 
-    closest_address = current_address;
+    nearest_address = current_address;
   }
   g_list_free (keys);
 
-  if (closest_address != NULL)
+  if (nearest_address != NULL)
   {
-    nearest->address = closest_address;
-    nearest->name = g_hash_table_lookup (table, closest_address);
+    nearest->address = nearest_address;
+    nearest->name = g_hash_table_lookup (table, nearest_address);
     return TRUE;
   }
 
@@ -639,7 +639,6 @@ gum_collect_symbol_if_function (const GumElfSymbolDetails * details,
   gpointer address;
   GArray * addresses;
   gboolean already_collected;
-  gchar * symbol_name;
 
   if (details->section_header_index == SHN_UNDEF || details->type != STT_FUNC)
     return TRUE;
@@ -674,7 +673,7 @@ gum_collect_symbol_if_function (const GumElfSymbolDetails * details,
 
   if (include_symbols)
   {
-    symbol_name = g_hash_table_lookup (gum_address_symbols, address);
+    gchar * symbol_name = g_hash_table_lookup (gum_address_symbols, address);
     if (symbol_name == NULL)
       g_hash_table_insert (gum_address_symbols, address, g_strdup (name));
   }
--- a/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
+++ b/gum/backend-libdwarf/gumsymbolutil-libdwarf.c
@@ -93,6 +93,8 @@ struct _GumDieDetails
   Dwarf_Debug dbg;
 };
 
+static gboolean gum_find_nearest_symbol_by_address (const gpointer address,
+    GumNearestSymbolDetails * nearest);
 static GumModuleEntry * gum_module_entry_from_address (gpointer address,
     GumNearestSymbolDetails * nearest);
 static GumModuleEntry * gum_module_entry_from_path_and_base (const gchar * path,
@@ -101,10 +103,13 @@ static Dwarf_Addr gum_module_entry_virtual_address_to_file (
     GumModuleEntry * self, gpointer address);
 
 static GHashTable * gum_get_function_addresses (void);
+static GHashTable * gum_get_address_symbols (void);
 static gboolean gum_collect_module_functions (const GumModuleDetails * details,
     gpointer user_data);
 static gboolean gum_collect_symbol_if_function (
     const GumElfSymbolDetails * details, gpointer user_data);
+static gboolean gum_collect_symbol (const GumElfSymbolDetails * details,
+    gpointer user_data);
 
 static void gum_symbol_util_ensure_initialized (void);
 static void gum_symbol_util_deinitialize (void);
@@ -146,6 +151,7 @@ static gint gum_compare_pointers (gconstpointer a, gconstpointer b);
 G_LOCK_DEFINE_STATIC (gum_symbol_util);
 static GHashTable * gum_module_entries = NULL;
 static GHashTable * gum_function_addresses = NULL;
+static GHashTable * gum_address_symbols = NULL;
 static GTimer * gum_cache_timer = NULL;
 
 gboolean
@@ -220,6 +226,12 @@ no_debug_info:
 
     g_strlcpy (details->module_name, entry->module->name,
         sizeof (details->module_name));
+
+    if (nearest.name == NULL)
+    {
+      gum_find_nearest_symbol_by_address (address, &nearest);
+    }
+
     if (nearest.name != NULL)
     {
       offset = GPOINTER_TO_SIZE (address) - GPOINTER_TO_SIZE (nearest.address);
@@ -295,6 +307,11 @@ no_debug_info:
   {
     gsize offset;
 
+    if (nearest.name == NULL)
+    {
+      gum_find_nearest_symbol_by_address (address, &nearest);
+    }
+
     if (nearest.name != NULL)
     {
       offset = GPOINTER_TO_SIZE (address) - GPOINTER_TO_SIZE (nearest.address);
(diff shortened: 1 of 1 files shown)


gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -551,6 +644,9 @@ gum_collect_module_functions (const GumModuleDetails * details,
   gum_elf_module_enumerate_symbols (entry->module,
       gum_collect_symbol_if_function, NULL);
 
+  gum_elf_module_enumerate_symbols (entry->module,
+      gum_collect_symbol, NULL);
```
oleavr, line 648: How about adding this logic to `gum_collect_symbol_if_function` instead? (To avoid walking the symbols twice. The predicate is the same as well.)

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -534,6 +601,32 @@ gum_get_function_addresses (void)
...
+  }
+  else
+  {
+    need_update = g_timer_elapsed (gum_cache_timer, NULL) >= GUM_MAX_CACHE_AGE;
+  }
+
+  if (need_update)
+  {
+    gum_process_enumerate_modules (gum_collect_module_functions, NULL);
+  }
```
oleavr, line 625: This common part should be extracted into a function and called from both. E.g. `gum_maybe_refresh_symbol_caches ()`

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +361,56 @@ gum_find_function (const gchar * name)
...
+  name = g_hash_table_lookup (table, address);
+
+  if (name != NULL)
+  {
+    nearest->name = name;
+    nearest->address = address;
+    return TRUE;
+  }
+
+  keys = g_hash_table_get_keys (table);
```
oleavr, line 385: This leaks `keys`.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -220,6 +226,12 @@ gum_symbol_details_from_address (gpointer address,
 
     g_strlcpy (details->module_name, entry->module->name,
         sizeof (details->module_name));
+
+    if (nearest.name == NULL)
+    {
+      gum_find_nearest_symbol_by_address (address, &nearest);
+    }
```
oleavr, line 233: Unnecessary block.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -295,6 +307,11 @@ gum_symbol_name_from_address (gpointer address)
   {
     gsize offset;
 
+    if (nearest.name == NULL)
+    {
+      gum_find_nearest_symbol_by_address (address, &nearest);
+    }
```
oleavr, line 313: Unnecessary block.

WorksButNotTested:
Changes made.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -93,6 +93,8 @@ struct _GumDieDetails
   Dwarf_Debug dbg;
 };
 
+static gboolean gum_find_nearest_symbol_by_address (const gpointer address,
```
oleavr, line 96:
```suggestion
static gboolean gum_find_nearest_symbol_by_address (gconstpointer address,
```

But seeing as we're using `gpointer` in the surrounding code it's probably better to be consistently doing it wrong until we fix it in the other places :)

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
+                                    GumNearestSymbolDetails * nearest)
+{
+  GHashTable * table;
+  gchar * name;
+  GList * keys;
+  GList * current;
+  gpointer current_address;
+  gpointer closest_address = NULL;
+
+  table = gum_get_address_symbols ();
```
oleavr, line 370:
```suggestion
  table = gum_get_address_symbols ();

```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
+  name = g_hash_table_lookup (table, address);
+
```
oleavr, line 372:
```suggestion
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
 }
 
+static gboolean
+gum_find_nearest_symbol_by_address (const gpointer address,
+                                    GumNearestSymbolDetails * nearest)
+{
+  GHashTable * table;
+  gchar * name;
+  GList * keys;
+  GList * current;
```
oleavr, line 366:
```suggestion
  GList * keys, * cur;
```

For consistency with other places where we walk such lists.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
+  gpointer current_address;
+  gpointer closest_address = NULL;
```
oleavr, line 368:
```suggestion
  gpointer current_address, closest_address;
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
+  name = g_hash_table_lookup (table, address);
+
+  if (name != NULL)
+  {
+    nearest->name = name;
+    nearest->address = address;
+    return TRUE;
+  }
+
+  keys = g_hash_table_get_keys (table);
```
oleavr, line 380: Perhaps doing `closest_address = NULL;` right before `keys =` would make it clearer what we're doing here.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,58 @@ gum_find_function (const gchar * name)
...
+
+    if (address < current_address)
+      continue;
+
+    if ((address - current_address) >= (address - closest_address))
+      continue;
+
+    closest_address = current_address;
+  }
+
```
oleavr, line 398:
```suggestion
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -546,10 +622,10 @@ gum_collect_module_functions (const GumModuleDetails * details,
     return TRUE;
 
   gum_elf_module_enumerate_dynamic_symbols (entry->module,
-      gum_collect_symbol_if_function, NULL);
+      gum_collect_symbol_if_function, (gpointer) FALSE);
 
   gum_elf_module_enumerate_symbols (entry->module,
-      gum_collect_symbol_if_function, NULL);
+      gum_collect_symbol_if_function, (gpointer) TRUE);
```
oleavr, line 628: Can't we do both in the callback instead of walking the symbols twice?

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -627,6 +714,9 @@ gum_symbol_util_deinitialize (void)
   g_hash_table_unref (gum_function_addresses);
   gum_function_addresses = NULL;
 
+  g_hash_table_unref (gum_address_symbols);
+  gum_address_symbols = NULL;
```
oleavr, line 718: Should be moved above `gum_function_addresses`, as these are destroyed in reverse order relative to the order they're declared in.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
WorksButNotTested: We need to NULL initialize closest_address. But I have merged the lines.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
WorksButNotTested: I'm not sure I folllow? The callback *gum_collect_symbol_if_function* now populates both data structures. We use it twice because there are two symbol tables to walk the one in .symtab and the other in .dynamic. We don't need to collect symbols from .dynamic into our reverse lookup as dladdr will work just fine on those. Hence we have the TRUE/FALSE parameter to avoid collecting unnecessary data.

Have I missed something?

WorksButNotTested:
More changes. Just one outstanding question.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
oleavr: Oh! I'm so sorry, I misread the code. That means it should have been two separate callbacks like it was, right? (Unless we want to populate `gum_function_addresses` using both?)

WorksButNotTested: So previously, we walked both .dynamic and .symtab with *gum_collect_symbol_if_function* and then walked .symtab with the new callback. So we walked the .symtab twice.

Now we have a single combined callback with a parameter to indicate whether we want to collect the data into the second structure. So we only walk each of .dynamic and .symtab the once.

So we have the slight added complexity of an additional parameter, and a callback that does two different jobs. But in exchange we should have the performance increase of only processing the tables the once.

I think maybe it's better this new way?

WorksButNotTested:
Note sure if my comment is showing up inline in the discussion, so I've copied it here in case.
"""
So previously, we walked both .dynamic and .symtab with gum_collect_symbol_if_function and then walked .symtab with the new callback. So we walked the .symtab twice.

Now we have a single combined callback with a parameter to indicate whether we want to collect the data into the second structure. So we only walk each of .dynamic and .symtab the once.

So we have the slight added complexity of an additional parameter, and a callback that does two different jobs. But in exchange we should have the performance increase of only processing the tables the once.

I think maybe it's better this new way?
"""

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -344,6 +356,56 @@ gum_find_function (const gchar * name)
...
 }
 
+static gboolean
+gum_find_nearest_symbol_by_address (gpointer address,
+                                    GumNearestSymbolDetails * nearest)
+{
+  GHashTable * table;
+  gchar * name;
+  GList * keys, * cur;
+  gpointer current_address, closest_address;
```
oleavr, line 366: Perhaps we should name it `nearest_address`, so we only use one term?

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -546,10 +620,10 @@ gum_collect_module_functions (const GumModuleDetails * details,
     return TRUE;
 
   gum_elf_module_enumerate_dynamic_symbols (entry->module,
-      gum_collect_symbol_if_function, NULL);
+      gum_collect_symbol_if_function, (gpointer) FALSE);
```
oleavr, line 623:
```suggestion
      gum_collect_symbol_if_function, GINT_TO_POINTER (FALSE));
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -546,10 +620,10 @@ gum_collect_module_functions (const GumModuleDetails * details,
...
 
   gum_elf_module_enumerate_symbols (entry->module,
-      gum_collect_symbol_if_function, NULL);
+      gum_collect_symbol_if_function, (gpointer) TRUE);
```
oleavr, line 626:
```suggestion
      gum_collect_symbol_if_function, GINT_TO_POINTER (TRUE));
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -560,10 +634,12 @@ static gboolean
 gum_collect_symbol_if_function (const GumElfSymbolDetails * details,
                                 gpointer user_data)
 {
+  gboolean include_symbols = (gboolean) user_data;
```
oleavr, line 637:
```suggestion
  gboolean include_symbols = GPOINTER_TO_INT (user_data);
```

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
```diff
@@ -560,10 +634,12 @@ static gboolean
...
   const gchar * name;
   gpointer address;
   GArray * addresses;
   gboolean already_collected;
+  gchar * symbol_name;
```
oleavr, line 642: Should be moved inside the block as it's only used there.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
oleavr: Ahh gotcha. Makes a lot of sense. Realize I was a bit knackered when doing the earlier review, so created quite a bit of confusion here. Sorry about that.

WorksButNotTested: No worries. We got there in the end.

gum/backend-libdwarf/gumsymbolutil-libdwarf.c
WorksButNotTested: Sounds good.

PR #419, opened 2020-04-19
Skip optional undefined args in v8

mrmacete:
In this way v8’s behaviour is consistent with duktape’s one.

### DIFF
--- a/bindings/gumjs/gumv8value.cpp
+++ b/bindings/gumjs/gumv8value.cpp
@@ -120,7 +120,7 @@ _gum_v8_args_parse (const GumV8Args * args,
     if (arg_index < arg_count)
         arg = (*info)[arg_index];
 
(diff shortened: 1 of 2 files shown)


bindings/gumjs/gumv8value.cpp
```diff
@@ -115,7 +115,12 @@ _gum_v8_args_parse (const GumV8Args * args,
       continue;
     }
 
-    if (arg_index >= arg_count)
+    Local<Value> arg;
+
+    if (arg_index < arg_count)
+        arg = (*info)[arg_index];
+
+    if (arg_index >= arg_count || arg->IsUndefined())
```
oleavr, line 123:
```suggestion
    if (arg_index >= arg_count || arg->IsUndefined ())
```

PR #428, opened 2020-05-10
Feature/stalker arm

WorksButNotTested:
Please note additional bug fix to gum_arm_writer_put_ands_reg_reg_imm.

### DIFF
--- a/gum/arch-arm/gumarmwriter.c
+++ b/gum/arch-arm/gumarmwriter.c
@@ -354,7 +354,7 @@ gum_arm_writer_put_call_address_body (GumArmWriter * self,
 
   aligned_address = address & ~GUM_ADDRESS (1);
 
-#ifdef TRUE // CHECK STACK ALIGN
+#if 0 // CHECK STACK ALIGN
   gconstpointer lbl = self->code + 1;
   gum_arm_writer_put_instruction (self, 0xe31d0007); // tst sp, #7
   gum_arm_writer_put_b_cond_label (self, ARM_CC_EQ, lbl);
--- a/gum/arch-arm/gumarmwriter.c
+++ b/gum/arch-arm/gumarmwriter.c
@@ -242,7 +242,7 @@ gum_arm_writer_add_literal_reference_here (GumArmWriter * self,
 }
 
 void
-_gum_arm_writer_put_call_address_with_arguments (GumArmWriter * self,
+gum_arm_writer_put_call_address_with_arguments (GumArmWriter * self,
                                                 GumAddress func,
                                                 guint n_args,
                                                 ...)
@@ -257,16 +257,6 @@ _gum_arm_writer_put_call_address_with_arguments (GumArmWriter * self,
 
   gum_arm_writer_put_argument_list_teardown (self, n_args);
 }
-
-__asm__ (
-  "gum_arm_writer_put_call_address_with_arguments:\n"
-  ".global gum_arm_writer_put_call_address_with_arguments\n"
-  "tst sp, #7\n"
-  "beq ok\n"
-  "udf 0x88\n"
-  "ok: \n"
-  "b _gum_arm_writer_put_call_address_with_arguments\n"
-);
 
 void
 gum_arm_writer_put_call_address_with_arguments_array (GumArmWriter * self,
@@ -288,8 +278,13 @@ gum_arm_writer_put_argument_list_setup (GumArmWriter * self,
 {
   gint arg_index;
 
-  if (n_args > 3)
-    g_error ("Stack passed arguments unsupported");
+  /*
+   * This function does not support the pushing of stack based arguments. If we
+   * add this later, we should note that the stack must be 8 byte aligned. We
+   * should take care not to misalign the stack when pushing an odd number of
+   * stack based arguments.
+   */
+  g_assert (n_args < 4);
 
   for (arg_index = (gint) n_args - 1; arg_index >= 0; arg_index--)
   {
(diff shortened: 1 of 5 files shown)


gum/arch-arm/gumarmwriter.c
```diff
@@ -278,6 +278,14 @@ gum_arm_writer_put_argument_list_setup (GumArmWriter * self,
 {
   gint arg_index;
 
+  /*
+   * This function does not support the pushing of stack based arguments. If we
+   * add this later, we should note that the stack must be 8 byte aligned. We
+   * should take care not to misalign the stack when pushing an odd number of
+   * stack based arguments.
+   */
+  g_assert (n_args < 4);
```
oleavr, line 287:
```suggestion
  g_assert (n_args <= 4);
```

gum/arch-arm/gumarmwriter.c
```diff
@@ -278,6 +278,14 @@ gum_arm_writer_put_argument_list_setup (GumArmWriter * self,
 {
   gint arg_index;
 
+  /*
+   * This function does not support the pushing of stack based arguments. If we
+   * add this later, we should note that the stack must be 8 byte aligned. We
+   * should take care not to misalign the stack when pushing an odd number of
+   * stack based arguments.
+   */
```
oleavr, line 286:
```suggestion
  /*
   * This function does not support the pushing of stack based arguments. If we
   * add this later, we should note that the stack must be 8 byte aligned.
   */
```

gum/arch-arm/gumthumbwriter.c
```diff
@@ -343,6 +343,12 @@ gum_thumb_writer_put_argument_list_setup (GumThumbWriter * self,
 {
   gint arg_index;
 
+  /*
+   * The stack must be 8 byte aligned, if we have an odd number of stack
+   * arguments, we will break this constraint.
+   */
+  g_assert (n_args < 4 || (n_args % 2) == 0);
```
oleavr, line 350:
```suggestion
```

gum/arch-arm/gumthumbwriter.c
```diff
@@ -343,6 +343,12 @@ gum_thumb_writer_put_argument_list_setup (GumThumbWriter * self,
...
+
```
oleavr, line 351:
```suggestion
```

gum/arch-arm/gumthumbwriter.c
```diff
@@ -343,6 +343,12 @@ gum_thumb_writer_put_argument_list_setup (GumThumbWriter * self,
 {
   gint arg_index;
 
+  /*
+   * The stack must be 8 byte aligned, if we have an odd number of stack
+   * arguments, we will break this constraint.
+   */
```
oleavr, line 349:
```suggestion
```

gum/backend-arm/gumstalker-arm-glue.S
```diff
@@ -9,12 +9,21 @@ _gum_stalker_follow_me:
...
-  stmdb sp!, {lr}
+  /*
+   * We need to store the lr onto the stack, so that it can be restored after
+   * returning from __gum_stalker_do_follow_me. However, when calling C code,
+   * the stack is expected to be 8 byte aligned. We therefore push r3 simply as
+   * padding. This is otherwise unnecessary since the calling convention
+   * dictates that r3 is the 4th argument and is therefore caller rather than
+   * callee saved. We don't use r0, as otherwise when restoring the stack, we
+   * would clobber the return value from __gum_stalker_do_follow_me.
+   */
```
oleavr, line 20:
```suggestion
  /*
   * We need to store LR onto the stack, so that it can be restored after
   * returning from do_follow_me(). However, when calling C code, the stack
   * is expected to be 8 byte aligned. We therefore push R3 simply as padding.
   * This is otherwise unnecessary since the calling convention dictates that
   * R3 is the 4th argument and is therefore caller rather than callee saved.
   * We don't use R0, as otherwise when restoring the stack, we would clobber
   * the return value from do_follow_me().
   */
```

gum/backend-arm/gumstalker-arm.c
```diff
@@ -2106,12 +2115,41 @@ gum_exec_ctx_write_thumb_prolog (GumExecCtx * ctx,
...
+   * Note, that unlike the ARM prolog which simply rounds down the stack
+   * pointer, the thumb instruction set often requires wide, or thumb v2
+   * instructions to work with registers other than R0-R7. We therefore retard
+   * the stack by 8, before rounding back up. This works as we know the stack
+   * must be 4 byte aligned since ARM architecture does not support unaligned
+   * data access. e.g. if the stack was already aligned, we simply retard the
+   * pointer by 8 (although wasting a few bytes of stack space, this still
+   * retains alignment), if it was misaligned, we retard the pointer by 8 before
+   * advancing back 4 bytes.
+   */
```
oleavr, line 2134:
```suggestion
  /*
   * Like in the ARM prolog we must now ensure that the stack is 8 byte aligned.
   * Note that unlike the ARM prolog, which simply rounds down the stack
   * pointer, the Thumb instruction set often requires wide, or Thumb v2
   * instructions to work with registers other than R0-R7. We therefore retard
   * the stack by 8, before rounding back up. This works as we know the stack
   * must be 4 byte aligned since ARM architecture does not support unaligned
   * data access. e.g. if the stack was already aligned, we simply retard the
   * pointer by 8 (although wasting a few bytes of stack space, this still
   * retains alignment), if it was misaligned, we retard the pointer by 8 before
   * advancing back 4 bytes.
   */
```

gum/backend-arm/gumstalker-arm.c
```diff
@@ -2131,6 +2169,15 @@ static void
 gum_exec_ctx_write_thumb_epilog (GumExecCtx * ctx,
                                  GumThumbWriter * cw)
 {
+  /*
+   * We know that the context structure was at the top of the stack at the end
+   * of the prolog, before the stack was aligned. Rather than working out how
+   * much alignment was needed, we can simply restore R10 back into SP to
+   * retrieve our stack pointer pre-alignment before we continue restoring the
+   * rest of the context.
+   */
```
oleavr, line 2178:
```suggestion
```

(Already explained in the ARM version.)

PR #430, opened 2020-05-12
Fix ThumbRelocator IT block eoi logic

bet4it:
Just borrow the variable `self->it_block.active` as a flag.

### DIFF
--- a/gum/arch-arm/gumthumbrelocator.c
+++ b/gum/arch-arm/gumthumbrelocator.c
@@ -214,43 +214,46 @@ gum_thumb_relocator_read_one (GumThumbRelocator * self,
   if (!cs_disasm_iter (self->capstone, &code, &size, &address, insn))
     return 0;
 
-  switch (insn->id)
+  if (!self->it_block.active)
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! What does this solve for you?

bet4it:
I add a test about it.

oleavr:
@bet4it Thanks! This breaks the Stalker ARM32 test-suite. You can run it like this: `./tests/run-android-arm.sh -p /Core/Stalker`

bet4it:
I don't have macOS, sad😅
https://github.com/frida/frida-gum/blob/58888455b404aeb5bfe0bf7dad2442097fc473de/tests/run-android-arm.sh#L10

oleavr:
@bet4it Replace that with linux if you're on Linux. It's a (bit too) dumb script 😄

bet4it:
Now all the tests have passed. What do you think about it?

oleavr:
Thanks! (Merged with a few minor style tweaks.)

PR #431, opened 2020-05-13
Feature/stalker arm

WorksButNotTested:


WorksButNotTested:
IT block stalker support

PR #462, opened 2020-07-08
Update gumarm64writer.c to make it compatible with macOS ARM

stevielavern:
PACed ptrs need to be stripped also on macOS arm64e.

### DIFF
--- a/gum/arch-arm64/gumarm64writer.c
+++ b/gum/arch-arm64/gumarm64writer.c
@@ -1429,7 +1429,7 @@ gum_arm64_writer_strip (GumArm64Writer * self,
   if (self->ptrauth_support != GUM_PTRAUTH_SUPPORTED)
     return value;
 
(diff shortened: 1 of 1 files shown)


gum/arch-arm64/gumarm64writer.c
```diff
@@ -1429,7 +1429,7 @@ gum_arm64_writer_strip (GumArm64Writer * self,
   if (self->ptrauth_support != GUM_PTRAUTH_SUPPORTED)
     return value;
 
-  if (self->target_os == GUM_OS_IOS)
+  if (self->target_os == GUM_OS_IOS || self->target_os == GUM_OS_MACOS)
```
oleavr, line 1432:
```suggestion
  if (self->target_os == GUM_OS_MACOS || self->target_os == GUM_OS_IOS)
```

(Nit: for consistency with ordering elsewhere.)

PR #465, opened 2020-07-15
Fix architecture detection on big-endian systems

WorksButNotTested:
Fix for #464. Implementation changed to expose a test seam in lieu of a mocking framework. Logic corrected.

### DIFF
--- a/tests/core/process.c
+++ b/tests/core/process.c
@@ -19,6 +19,11 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined (HAVE_LINUX) && !defined (HAVE_ANDROID)
+#include <sys/auxv.h>
+#include "backend-linux/gumlinux.h"
+#endif
+
 #define TESTCASE(NAME) \
     void test_process_ ## NAME (void)
 #define TESTENTRY(NAME) \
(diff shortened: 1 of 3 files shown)


tests/core/process.c
```diff
@@ -293,6 +302,128 @@ TESTCASE (linux_process_modules)
...
+  cpu32 = GUM_CPU_MIPS;
+#else
+# error Unsupported architecture
+#endif
+  GumCpuType result;
+  guint32 test_data[] =
+  {
+    AT_NULL, 0
+  };
+  result = gum_linux_cpu_type_from_auxv ((guint8 *)test_data,
```
oleavr, line 322: The `result` is only used once and doesn't add any clarity, so perhaps the statement should just be inlined into the assertion?

tests/core/process.c
```diff
@@ -293,6 +302,128 @@ TESTCASE (linux_process_modules)
...
+      sizeof (test_data));
+  g_assert_true (result == cpu32);
+}
+
```
oleavr, line 326: ^ And same as above in the other tests.

tests/core/process.c
```diff
@@ -293,6 +302,120 @@ TESTCASE (linux_process_modules)
...
+  cpu32 = GUM_CPU_MIPS;
+#else
+# error Unsupported architecture
+#endif
+  guint32 test_data[] =
+  {
+    AT_NULL, 0
+  };
+  g_assert_cmpuint (gum_linux_cpu_type_from_auxv ((const guint8 *)test_data,
+      sizeof (test_data)), ==, cpu32);
```
oleavr, line 322:
```suggestion
  g_assert_cmpuint (gum_linux_cpu_type_from_auxv (test_data,
      sizeof (test_data)), ==, cpu32);
```

oleavr: Was missing a space after cast, but with the change to `gconstpointer` there's no need for a cast anymore.

(And same changes below.)

PR #473, opened 2020-08-04
Better discovery of the libc for linux and qnx.

demantz:
The libc is currently detected by checking for the 'exit' symbol.
Whichever shared library contains a function named 'exit' could
falsely be detected to be the libc. This commit adds another check
for the '__libc_start_main' symbol which should be more libc-specific.

As a minimal example which shows the bug, the following program
crashes when frida attaches:

#include <stdio.h>;
void exit(int i) {
    printf("foo");
}
int main(int argc, char** argv) {
    printf("Hi!\n");
    getchar();
    printf("bye!\n");
    return 0;
}

If the libc_start_main symbol is not found, the old check for the exit function
will still happen - so this patch shouldn't make things worse, right?

### DIFF
--- a/gum/backend-linux/gumprocess-linux.c
+++ b/gum/backend-linux/gumprocess-linux.c
@@ -308,10 +308,14 @@ gum_try_init_libc_name (void)
 {
   Dl_info info = { NULL, };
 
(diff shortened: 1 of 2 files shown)


oleavr:
Good point. Thanks a lot!

PR #480, opened 2020-08-16
Support linking against system GLib

meme:
This PR adds support for linking against a system GLib, as well as adds support for building GObject Introspection files (if the user has the required tools installed on their system).

The Frida build system HAS NOT been updated to support this change as more work is planned for that. However, it's straightfoward to test:

```diff
 # We need these legacy paths for dependencies that don't use pkg-config.
 legacy_includes="-I$FRIDA_PREFIX/include"
-legacy_libpaths="-L$FRIDA_PREFIX/lib"
+#legacy_libpaths="-L$FRIDA_PREFIX/lib"
 if [ "$FRIDA_ENV_SDK" != 'none' ]; then
   legacy_includes="$legacy_includes -I$FRIDA_SDKROOT/include"
-  legacy_libpaths="$legacy_libpaths -L$FRIDA_SDKROOT/lib"
+  #legacy_libpaths="$legacy_libpaths -L$FRIDA_SDKROOT/lib"
 fi
 CPPFLAGS="$CPPFLAGS $legacy_includes"
-LDFLAGS="$LDFLAGS $legacy_libpaths"
+#LDFLAGS="$LDFLAGS $legacy_libpaths"
 
 meson_legacy_includes=$(flags_to_args "$legacy_includes")
 meson_legacy_libpaths=$(flags_to_args "$legacy_libpaths")
@@ -870,9 +870,11 @@ chmod 755 "$strip_wrapper"
 
 PKG_CONFIG=$FRIDA_BUILD/${FRIDA_ENV_NAME:-frida}-${host_platform_arch}-pkg-config
 
-pkg_config="$FRIDA_TOOLROOT/bin/pkg-config"
+pkg_config="pkg-config"
+# pkg_config="$FRIDA_TOOLROOT/bin/pkg-config"
 pkg_config_flags=""
-pkg_config_path="$FRIDA_PREFIX_LIB/pkgconfig"
+pkg_config_path="/usr/lib/pkgconfig"
+# pkg_config_path="$FRIDA_PREFIX_LIB/pkgconfig"
 if [ "$FRIDA_ENV_SDK" != 'none' ]; then
   pkg_config_flags=" --define-variable=frida_sdk_prefix=$FRIDA_SDKROOT"
   pkg_config_path="$pkg_config_path:$FRIDA_SDKROOT/lib/pkgconfig"
@@ -880,7 +882,7 @@ fi
 (
   echo "#!/bin/sh"
   echo "export PKG_CONFIG_PATH=\"$pkg_config_path\""
-  echo "exec \"$pkg_config\"$pkg_config_flags --static \"\$@\""
+  echo "exec \"$pkg_config\"$pkg_config_flags \"\$@\""
 ) > "$PKG_CONFIG"
 chmod 755 "$PKG_CONFIG"
```
(This is a crap changeset, but it makes the build system use the system's `pkg-config`, as well as removes the Frida path from `LDFLAGS`.)

The test suite has been modified to disable tests that cease to function on a non-Frida GLib as those tests required the patches applied to Frida's GLib. Tests still pass on the original build configuration as well as this new one.

### DIFF
--- a/gum/gumenumtypes.c.template
+++ b/gum/gumenumtypes.c.template
@@ -1,0 +1,39 @@
+/*** BEGIN file-header ***/
+#include "gumenumtypes.h"
+
+#include <gum/gum.h>
+
+/*** END file-header ***/
+
+/*** BEGIN file-production ***/
+
+/* Enumerations from "@basename@" */
+/*** END file-production ***/
+
+/*** BEGIN value-header ***/
+GType
+@enum_name@_get_type (void)
+{
+  static gsize static_g_define_type_id = 0;
+
+  if (g_once_init_enter (&static_g_define_type_id))
+  {
+    static const G@Type@Value values[] = {
+/*** END value-header ***/
+
+/*** BEGIN value-production ***/
+      { @VALUENAME@, "@VALUENAME@", "@valuenick@" },
+/*** END value-production ***/
+
+/*** BEGIN value-tail ***/
+      { 0, NULL, NULL }
+    };
+    GType g_define_type_id = g_@type@_register_static (
+        g_intern_static_string ("@EnumName@"), values);
+    g_once_init_leave (&static_g_define_type_id, g_define_type_id);
+  }
+
+  return static_g_define_type_id;
+}
+
+/*** END value-tail ***/
--- a/gum/gumexceptor.h
+++ b/gum/gumexceptor.h
@@ -25,7 +25,9 @@ G_DECLARE_FINAL_TYPE (GumExceptor, gum_exceptor, GUM, EXCEPTOR, GObject)
 #else
 # define GUM_NATIVE_SETJMP(env) sigsetjmp (env, TRUE)
 # define GUM_NATIVE_LONGJMP siglongjmp
+# if !defined (GUM_GIR_COMPILATION)
   typedef sigjmp_buf GumExceptorNativeJmpBuf;
+# endif
 #endif
 
 typedef struct _GumExceptionDetails GumExceptionDetails;
--- a/gum/meson.build
+++ b/gum/meson.build
@@ -1,3 +1,5 @@
+gnome = import('gnome')
+
 gum_headers = [
   'gum.h',
   'gumapiresolver.h',
@@ -75,8 +77,10 @@ gum_sources = [
   'arch-mips/gummipsrelocator.c',
 ]
 
+gum_backend_headers = []
+
 if host_os_family == 'windows'
-  gum_headers += [
+  gum_backend_headers += [
     'backend-windows/gumwindows.h',
     'backend-dbghelp/gumdbghelp.h',
     'backend-dbghelp/gumdbghelpbacktracer.h',
@@ -98,7 +102,7 @@ else
 endif
 
 if host_os_family == 'darwin'
-  gum_headers += [
+  gum_backend_headers += [
     'backend-darwin/gumdarwin.h',
     'backend-darwin/gumdarwinbacktracer.h',
     'backend-darwin/gumdarwinsymbolicator.h',
@@ -124,7 +128,7 @@ if host_os_family == 'darwin'
 endif
 
 if host_os_family == 'linux'
-  gum_headers += [
+  gum_backend_headers += [
     'backend-linux/gumlinux.h',
   ]
   gum_sources += [
@@ -135,7 +139,7 @@ if host_os_family == 'linux'
     'backend-posix/gumexceptor-posix.c',
   ]
   if host_os == 'android'
-    gum_headers += [
+    gum_backend_headers += [
       'backend-linux/gumandroid.h',
     ]
     gum_sources += [
@@ -145,7 +149,7 @@ if host_os_family == 'linux'
 endif
 
 if host_os_family == 'qnx'
-  gum_headers += [
+  gum_backend_headers += [
     'backend-qnx/gumqnx.h',
   ]
   gum_sources += [
@@ -158,7 +162,7 @@ if host_os_family == 'qnx'
 endif
 
 if host_os_family == 'linux' or host_os_family == 'qnx'
-  gum_headers += [
+  gum_backend_headers += [
     'backend-elf/gumelfmodule.h',
   ]
   gum_sources += [
@@ -250,7 +254,7 @@ if host_arch == 'mips'
 endif
 
 if unwind_dep.found()
-  gum_headers += [
+  gum_backend_headers += [
     'backend-libunwind/gumunwbacktracer.h',
   ]
   gum_sources += [
@@ -264,7 +268,7 @@ if have_libdwarf
   ]
 endif
 
-install_headers(gum_headers, subdir: install_header_subdir)
+install_headers(gum_headers + gum_backend_headers, subdir: install_header_subdir)
 install_headers(gum_x86_headers, subdir: install_header_subdir + '/arch-x86')
 install_headers(gum_arm_headers, subdir: install_header_subdir + '/arch-arm')
 install_headers(gum_arm64_headers, subdir: install_header_subdir + '/arch-arm64')
@@ -277,6 +281,26 @@ gum = library('frida-gum-' + api_version, gum_sources,
   dependencies: [glib_dep, gobject_dep, gio_dep, ffi_dep, capstone_dep] + extra_deps,
   install: true,
 )
+
+if gir.found()
+  gir_args = [
+    '--quiet',
+    '-DGUM_GIR_COMPILATION',
+  ]
+
+  gnome.generate_gir(gum,
+    sources: gum_sources + gum_headers,
+    namespace: 'Gum',
+    nsversion: api_version,
+    identifier_prefix: 'Gum',
+    symbol_prefix: 'gum',
+    export_packages: 'gum',
+    includes: ['GLib-2.0', 'GObject-2.0', 'Gio-2.0'],
+    header: 'gum/gum.h',
+    extra_args: gir_args + extra_libs_private,
+    install: true,
+  )
+endif
 
 gum_dep = declare_dependency(link_with: gum,
   include_directories: gum_incdirs,
--- a/gum/meson.build
+++ b/gum/meson.build
@@ -274,7 +274,19 @@ install_headers(gum_arm_headers, subdir: install_header_subdir + '/arch-arm')
 install_headers(gum_arm64_headers, subdir: install_header_subdir + '/arch-arm64')
 install_headers(gum_mips_headers, subdir: install_header_subdir + '/arch-mips')
 
-gum = library('frida-gum-' + api_version, gum_sources,
+gum_enums = gnome.mkenums('gum-enum-types',
+  sources: gum_headers,
+  c_template: 'gumenumtypes.c.template',
+  h_template: 'gumenumtypes.h.template',
+  install_dir: install_header_subdir,
+  install_header: true,
+)
+
+gum_enums_dep = declare_dependency(sources: gum_enums)
+
+gum_enum_h = gum_enums[1]
+
+gum = library('frida-gum-' + api_version, gum_sources + gum_enums,
   c_args: frida_component_cflags,
   include_directories: gum_incdirs,
   link_args: extra_libs_private,
@@ -289,7 +301,7 @@ if gir.found()
   ]
 
   gnome.generate_gir(gum,
-    sources: gum_sources + gum_headers,
+    sources: gum_sources + gum_headers + [gum_enum_h],
     namespace: 'Gum',
     nsversion: api_version,
     identifier_prefix: 'Gum',
@@ -305,7 +317,7 @@ endif
 gum_dep = declare_dependency(link_with: gum,
   include_directories: gum_incdirs,
   link_args: extra_libs_private,
-  dependencies: [glib_dep, gobject_dep, gio_dep, capstone_dep],
+  dependencies: [glib_dep, gobject_dep, gio_dep, capstone_dep, gum_enums_dep],
 )
 
 pkg = import('pkgconfig')
--- a/meson.build
+++ b/meson.build
@@ -194,6 +194,14 @@ extra_deps = []
 extra_requires_private = []
 extra_libs_private = []
 
+if cc.has_function('g_thread_set_callbacks', dependencies: [glib_dep])
+  cdata.set('HAVE_FRIDA_GLIB', 1)
+endif
+
+if cc.has_function('ffi_set_mem_callbacks', dependencies: [ffi_dep])
+  cdata.set('HAVE_FRIDA_LIBFFI', 1)
+endif
+
 if host_os == 'android'
   extra_libs_private += ['-llog']
 endif
--- a/meson.build
+++ b/meson.build
@@ -194,6 +194,20 @@ extra_deps = []
 extra_requires_private = []
 extra_libs_private = []
 
+glib_flavor_src = '''
+#include <glib.h>
+
+#ifdef GLIB_STATIC_COMPILATION
+# error GLib is static
+#endif
+'''
+have_shared_glib = cc.compiles(glib_flavor_src, name: 'compiling with shared GLib', dependencies: [glib_dep])
+if have_shared_glib
+  gir = find_program('g-ir-scanner')
+else
+  gir = disabler()
+endif
+
 if cc.has_function('g_thread_set_callbacks', dependencies: [glib_dep])
   cdata.set('HAVE_FRIDA_GLIB', 1)
 endif
--- a/tests/core/interceptor.c
+++ b/tests/core/interceptor.c
@@ -37,7 +37,9 @@ TESTLIST_BEGIN (interceptor)
   TESTENTRY (attach_detach_torture)
 #endif
   TESTENTRY (thread_id)
-#if !(defined (HAVE_ANDROID) && defined (HAVE_ARM64)) && !defined (HAVE_ASAN)
+#if defined (HAVE_FRIDA_GLIB) && \
+    !(defined (HAVE_ANDROID) && defined (HAVE_ARM64)) && \
+    !defined (HAVE_ASAN)
   TESTENTRY (intercepted_free_in_thread_exit)
 #endif
   TESTENTRY (function_arguments)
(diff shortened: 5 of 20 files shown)


meson.build
```diff
@@ -199,6 +199,22 @@ gumjs_extra_deps = []
...
 
+have_gir = false
+
+if cc.has_function('g_thread_set_callbacks', dependencies: [glib_dep])
+  cdata.set('HAVE_FRIDA_GLIB', 1)
+else
+  elf_dep = dependency('libelf')
+  extra_deps += [elf_dep]
+
+  gir = find_program('g-ir-scanner')
```
oleavr, line 210: I think this should be keyed off whether we're building Gum as a shared library; nothing's stopping us from building our GLib as a dynamic library as well.

tests/core/interceptor.c
```diff
@@ -37,7 +37,7 @@ TESTLIST_BEGIN (interceptor)
   TESTENTRY (attach_detach_torture)
 #endif
   TESTENTRY (thread_id)
-#if !(defined (HAVE_ANDROID) && defined (HAVE_ARM64)) && !defined (HAVE_ASAN)
+#if !(defined (HAVE_ANDROID) && defined (HAVE_ARM64)) && !defined (HAVE_ASAN) && HAVE_FRIDA_GLIB
```
oleavr, line 40: Should be checking if it's defined.

tests/core/interceptor.c
```diff
@@ -56,7 +56,9 @@ TESTLIST_BEGIN (interceptor)
   TESTENTRY (already_replaced)
 # ifndef HAVE_ASAN
   TESTENTRY (replace_one)
+#if HAVE_FRIDA_GLIB
```
oleavr, line 59: Should be checking if it's defined.

tests/gumjs/script.c
```diff
@@ -8,6 +8,7 @@
 #include "script-fixture.c"
 
 TESTLIST_BEGIN (script)
+#ifdef HAVE_FRIDA_GLIB
```
oleavr, line 11: Let's disable the whole suite in `main()` instead.

tests/gumtest.c
```diff
@@ -44,7 +44,7 @@ static guint get_number_of_tests_in_suite (GTestSuite * suite);
 gint
 main (gint argc, gchar * argv[])
 {
-#if !DEBUG_HEAP_LEAKS && !defined (HAVE_ASAN)
+#if !DEBUG_HEAP_LEAKS && !defined (HAVE_ASAN) && HAVE_FRIDA_GLIB
```
oleavr, line 47: Should be checking if it's defined.

tests/gumtest.c
```diff
@@ -104,14 +104,18 @@ main (gint argc, gchar * argv[])
   }
   else
   {
+    #ifdef HAVE_FRIDA_GLIB
```
oleavr, line 107: Inconsistent indentation.

tests/prof/sampler.c
```diff
@@ -10,7 +10,7 @@
 TESTLIST_BEGIN (sampler)
   TESTENTRY (cycle)
   TESTENTRY (busy_cycle)
-#ifndef HAVE_ASAN
+#if !defined(HAVE_ASAN) && HAVE_FRIDA_GLIB
```
oleavr, line 13: Should be checking if it's defined.

oleavr:
Oh I almost forgot: we also need to update the Windows build system. 😅  The easiest way to pinpoint where the defines need to go is probably `git grep HAVE_I386`.

meson.build
```diff
@@ -199,6 +199,22 @@ gumjs_extra_deps = []
 gumjs_extra_requires = []
 gumjs_extra_libraries = []
 
+have_gir = false
+
+if cc.has_function('g_thread_set_callbacks', dependencies: [glib_dep])
+  cdata.set('HAVE_FRIDA_GLIB', 1)
+else
+  elf_dep = dependency('libelf')
+  extra_deps += [elf_dep]
```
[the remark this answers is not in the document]
meme: I think since we were using aggressive static linking a peer had the same dependency and it pulled it in. On my system, I was getting linker errors to symbols inside `libelf`. This PR adds it conditionally because the Frida toolchain doesn't provide a `libelf` pkg-config, so adding a `dependency('libelf')` will throw a missing dependency error.

meson.build
```diff
@@ -199,6 +199,33 @@ gumjs_extra_deps = []
...
+
+if has_shared_glib
+  gir = find_program('g-ir-scanner')
+  have_gir = gir.found()
+endif
+
+if cc.has_function('g_thread_set_callbacks', dependencies: [glib_dep])
+  cdata.set('HAVE_FRIDA_GLIB', 1)
+else
+  elf_dep = dependency('libelf')
```
oleavr, line 221: These two lines should be moved into/consolidated with with the code inside:

`if host_os_family == 'linux' or host_os_family == 'qnx'`

tests/core/interceptor.c
```diff
@@ -56,7 +56,9 @@ TESTLIST_BEGIN (interceptor)
   TESTENTRY (already_replaced)
 # ifndef HAVE_ASAN
   TESTENTRY (replace_one)
+# if defined (HAVE_FRIDA_GLIB)
```
oleavr, line 59:
```suggestion
# ifdef HAVE_FRIDA_GLIB
```

To match the surrounding style.

tests/core/interceptor.c
```diff
@@ -56,7 +56,9 @@ TESTLIST_BEGIN (interceptor)
   TESTENTRY (already_replaced)
 # ifndef HAVE_ASAN
```
oleavr, line 57:
```suggestion
#ifndef HAVE_ASAN
```

(Just spotted this existing style bug.)

tests/core/interceptor.c
```diff
@@ -56,7 +56,9 @@ TESTLIST_BEGIN (interceptor)
...
   TESTENTRY (replace_one)
+# if defined (HAVE_FRIDA_GLIB)
   TESTENTRY (replace_two)
+# endif
 # endif
```
oleavr, line 62:
```suggestion
#endif
```

(Just spotted this existing style bug.)

tests/core/interceptor.c
```diff
@@ -233,13 +235,15 @@ TESTCASE (thread_id)
   g_assert_cmpuint (second_thread_id, !=, first_thread_id);
 }
 
+#ifdef HAVE_FRIDA_GLIB
```
oleavr, line 238:
```suggestion
#ifdef HAVE_FRIDA_GLIB

```

tests/core/interceptor.c
```diff
@@ -233,13 +235,15 @@ TESTCASE (thread_id)
...
 TESTCASE (intercepted_free_in_thread_exit)
 {
   interceptor_fixture_attach (fixture, 0, interceptor_fixture_get_libc_free (),
       'a', 'b');
   g_thread_join (g_thread_new ("interceptor-test-thread-exit",
       target_nop_function_a, NULL));
 }
+#endif
```
oleavr, line 246:
```suggestion

#endif
```

tests/core/interceptor.c
```diff
@@ -582,6 +586,7 @@ static gpointer replacement_malloc_calling_malloc_and_replaced_free (
     gsize size);
 static void replacement_free_doing_nothing (gpointer mem);
 
+#ifdef HAVE_FRIDA_GLIB
```
oleavr, line 589:
```suggestion
#ifdef HAVE_FRIDA_GLIB

```

tests/core/interceptor.c
```diff
@@ -612,6 +617,7 @@ TESTCASE (replace_two)
 
   free (ret);
 }
+#endif
```
oleavr, line 620:
```suggestion

#endif
```

tests/gumtest.c
```diff
@@ -44,7 +44,7 @@ static guint get_number_of_tests_in_suite (GTestSuite * suite);
 gint
 main (gint argc, gchar * argv[])
 {
-#if !DEBUG_HEAP_LEAKS && !defined (HAVE_ASAN)
+#if !DEBUG_HEAP_LEAKS && !defined (HAVE_ASAN) && defined (HAVE_FRIDA_GLIB)
```
oleavr, line 47: New check should go first. (See previous comment.)

tests/gumtest.c
```diff
@@ -269,6 +273,7 @@ main (gint argc, gchar * argv[])
   {
     GumScriptBackend * duk_backend, * v8_backend;
 
+# ifdef HAVE_FRIDA_GLIB
```
oleavr, line 276: Should be moved up and merged with the `#if defined (HAVE_GUMJS)` instead, i.e.:

```C
#if defined (HAVE_GUMJS) && defined (HAVE_FRIDA_GLIB)
```

tests/prof/sampler.c
```diff
@@ -10,7 +10,7 @@
 TESTLIST_BEGIN (sampler)
   TESTENTRY (cycle)
   TESTENTRY (busy_cycle)
-#ifndef HAVE_ASAN
+#if !defined(HAVE_ASAN) && defined (HAVE_FRIDA_GLIB)
```
oleavr, line 13: Should go first here. (See earlier comment.)

tests/prof/sampler.c
```diff
@@ -75,6 +75,7 @@ struct _MallocCountHelperContext
   GumSample count;
 };
 
+#ifdef HAVE_FRIDA_GLIB
```
oleavr, line 78:
```suggestion
#ifdef HAVE_FRIDA_GLIB

```

tests/prof/sampler.c
```diff
@@ -123,6 +124,7 @@ TESTCASE (malloc_count)
   g_assert_cmpuint (gum_call_count_sampler_peek_total_count (
       GUM_CALL_COUNT_SAMPLER (fixture->sampler)), >=, 3 + 1);
 }
+#endif
```
oleavr, line 127:
```suggestion

#endif
```

gum/gumexceptor.h
```diff
@@ -24,7 +24,9 @@ G_DECLARE_FINAL_TYPE (GumExceptor, gum_exceptor, GUM, EXCEPTOR, GObject)
 #else
 # define GUM_NATIVE_SETJMP(env) sigsetjmp (env, TRUE)
 # define GUM_NATIVE_LONGJMP siglongjmp
+# if defined (HAVE_FRIDA_GLIB)
```
oleavr, line 27: It's valid to be using a shared build of Frida's GLib with GIR so this is the wrong check – can we get GIR to ignore it instead?

gum/gum-enum-types.c.template
```diff
@@ -0,0 +1,38 @@
+/*** BEGIN file-header ***/
+#include "config.h"
```
oleavr, line 2: We use `-include config.h`, so this include should be omitted.

gum/gum-enum-types.c.template
```diff
@@ -0,0 +1,38 @@
...
+/*** END file-production ***/
+
+/*** BEGIN value-header ***/
+GType
+@enum_name@_get_type (void)
+{
+  static volatile gsize g_define_type_id__volatile = 0;
+
+  if (g_once_init_enter (&g_define_type_id__volatile))
+    {
```
oleavr, line 19: This doesn't match the Gum coding style. (I can't pinpoint all the divergences without seeing the output, but at least the indentation here doesn't match.)

gum/gum-enum-types.c.template
```diff
@@ -0,0 +1,38 @@
...
+
+/*** BEGIN value-production ***/
+        { @VALUENAME@, "@VALUENAME@", "@valuenick@" },
+/*** END value-production ***/
+
+/*** BEGIN value-tail ***/
+        { 0, NULL, NULL }
+      };
+      GType g_define_type_id =
+        g_@type@_register_static (g_intern_static_string ("@EnumName@"), values);
```
oleavr, line 31:
```suggestion
          g_@type@_register_static (g_intern_static_string ("@EnumName@"), values);
```

gum/meson.build
```diff
@@ -274,7 +274,18 @@ install_headers(gum_arm_headers, subdir: install_header_subdir + '/arch-arm')
 install_headers(gum_arm64_headers, subdir: install_header_subdir + '/arch-arm64')
 install_headers(gum_mips_headers, subdir: install_header_subdir + '/arch-mips')
 
-gum = library('frida-gum-' + api_version, gum_sources,
+gum_enums = gnome.mkenums(
+  'gum-enum-types',
+  sources: gum_headers,
+  c_template: 'gum-enum-types.c.template',
+  h_template: 'gum-enum-types.h.template',
```
oleavr, line 281:
```suggestion
  c_template: 'gumenumtypes.c.template',
  h_template: 'gumenumtypes.h.template',
```

For consistency with the style used in this repo.

gum/meson.build
```diff
@@ -289,7 +300,7 @@ if have_gir
 
   gnome.generate_gir(
     gum,
-    sources: gum_sources + gum_headers,
+    sources: gum_sources + gum_headers + [gum_enum_h],
```
oleavr, line 303: I think we have some handwritten _get_type() implementations for some of our enums – won't this conflict? If so we will also need to update the Windows build system to use a custom build step. (I can assist with this part if you don't have access to a Windows dev env.)

meme: Yes. We need to root these out and change them so they are properly formatted enums; is this appropriate for this PR? `GumEventType` being a prime example. And yes, I will need to enlist your help for ensuring this PR works on Windows :smile:.

gum/gumexceptor.h
meme: Using GLib annotations (like you saw for Capstone) just didn't work every single way I twisted it. I will try again, but yeah we need to figure out a better wayt to fix this (GObject Introspection generates an invalid GIR for this type, and I do not know of any other way to fix it... other than to ignore it)

gum/meson.build
oleavr: Cool! Yep, I think we should deal with it in this PR.

I think it's only these three:
```sh
% git grep '_get_type' gum | grep "\.c" | grep "(void)"
gum/gum.c:gum_cpu_type_get_type (void)
gum/gumdarwinmodule.c:gum_darwin_module_flags_get_type (void)
gum/gumprocess.c:gum_code_signing_policy_get_type (void)
%
```

By “properly formatted enums”, do you mean that the `get_type()` is generated, or also that e.g. `GUM_CALL` needs to become `GUM_EVENT_TYPE_CALL`? (Haven't looked into what GIR expects, but I imagine it can be overridden through annotations.)

gum/gumexceptor.h
oleavr: Oh, interesting. Perhaps GIR sets a preprocessor define that we can detect?

gum/gumenumtypes.c.template
```diff
@@ -0,0 +1,36 @@
...
+
+/*** BEGIN value-production ***/
+      { @VALUENAME@, "@VALUENAME@", "@valuenick@" },
+/*** END value-production ***/
+
+/*** BEGIN value-tail ***/
+      { 0, NULL, NULL }
+    };
+    GType g_define_type_id =
+      g_@type@_register_static (g_intern_static_string ("@EnumName@"), values);
```
oleavr, line 29: Hanging indent should be four spaces, not two.

gum/meson.build
meme: `GUM_CALL` needs to be `GUM_EVENT_TYPE_CALL`, yeah. I haven't looked into other enums in the Gum codebase yet, so perhaps you have more insight on that. Regardless, I think we should fix the `get_type()`s in this PR, and then deal with the malformed enum names in another?

meson.build
```diff
@@ -218,6 +244,10 @@ if host_os_family == 'linux' or host_os_family == 'qnx'
   if not have_libelf
     error('libelf is required')
   endif
+  if not have_frida_glib
+    elf_dep = dependency('libelf')
+    extra_deps += [elf_dep]
+  endif
```
oleavr, line 250: This should be done unconditionally inside this branch – but I seem to recall that the reason we check for `libelf.h` here and manually add `-lelf` is that a `.pc` is not available on all systems (e.g. Android and QNX)... So I wonder if we can make the existing logic work for GIR somehow? Like setting `required: false`, and taking the `-lelf` route as a fallback.

(Do we actually need libelf for GIR now that we're skipping backend bits?)

gum/meson.build
oleavr: Ok. Well, there's got to be an annotation to override this, as I recall e.g. GStreamer has several such enums, for instance [GstFlowReturn](https://gstreamer.freedesktop.org/documentation/gstreamer/gstpad.html?gi-language=c#GstFlowReturn) with members such as `GST_FLOW_OK`.

meme: I was wrong: we are not `typedef enum`ing, which is what is causing this problem. So, we'll update all enums to use this format (ABI breaking change, since we currently use `guint` for all enum types).

meson.build
meme: We are only skipping backend bits for GIR generation, not for linking Gum, so libelf is required to link still. The `.pc` for libelf is _not_ provided in the Frida toolchain, which is why we cannot do a `dependency` in here. However, on a non-Frida-GLib build we are targeting the system GLib, so we must use `dependency` to get `libelf`. I am not clear on how using `required: false` and setting a fallback is any better than this?

gum/meson.build
meme: For review purposes, here are the generated files for these enums (and the others): https://gist.github.com/meme/b7eb683a6294c04469bf95cc791290e1

meme:
As a long overdue update: this currently works on Linux, but the `glib-mkenum` changes require some work on Windows. I've got `glib-mkenum` in the toolchain and I have gum-64 building, but the dependent projects like gumjs-64, etc. do not point to the correct header include path when using a Custom Build Step. Development is on-going but slow, because I am mostly unfamiliar with Windows tooling (Visual Studio).

meme:
Rebased against latest `master` and tests still pass in both configurations :construction:

oleavr:
@meme Yay, this evening I finally got around to looking into the Windows build system bits, so it's finally merge o'clock! 🎉

Thanks a lot for this awesome PR and all your patience! ❤️

PR #486, opened 2020-08-18
Add stalker coverage support

WorksButNotTested:


oleavr:
Thanks! ❤️ As we discussed out-of-band, let's solve this in “userspace” instead, i.e. a Frida-specific [module](https://github.com/oleavr/frida-module-example) published to npm.

PR #487, opened 2020-08-24
Support a NULL EventSink in Stalker

meme:
~~This PR is still in the proposal phase as I haven't added ARM64 or MIPS support nor reasonable tests.~~

For background: in my personal project hotwax, my [`event_sink.c`](https://github.com/meme/hotwax/blob/master/src/event_sink.c) simply acts as a stub which does nothing (and returns a query_mask of 0). Now, in another project where I use the Frida Stalker, I've run into the same problem: I do not need an EventSink, but are required to provide one.

So, I propose that we allow EventSink to be `NULL` so that a user can provide ONLY a transformer and nothing else. Or alternatively, they can provide no Transformer or EventSink (not sure what the use-case is for this, but it is now allowed)

So my goals for this PR are _speed_. I want to provide reasonable speed benefits if you do not provide an EventSink. This is why I ask for your feedback @oleavr to see:
1) If this is reasonable
2) How we can design this so that it provides the most speed benefit

### DIFF
--- a/gum/gumeventsink.c
+++ b/gum/gumeventsink.c
@@ -6,7 +6,25 @@
 
 #include "gumeventsink.h"
 
+struct _GumDefaultEventSink
+{
+  GObject parent;
+};
+
+static void gum_default_event_sink_iface_init (gpointer g_iface,
+    gpointer iface_data);
+static GumEventType gum_default_event_sink_query_mask (GumEventSink * sink);
+static void gum_default_event_sink_process (GumEventSink * sink,
+    const GumEvent * ev);
+
 G_DEFINE_INTERFACE (GumEventSink, gum_event_sink, G_TYPE_OBJECT)
+
+G_DEFINE_TYPE_EXTENDED (GumDefaultEventSink,
+                        gum_default_event_sink,
+                        G_TYPE_OBJECT,
+                        0,
+                        G_IMPLEMENT_INTERFACE (GUM_TYPE_EVENT_SINK,
+                            gum_default_event_sink_iface_init))
 
 static void
 gum_event_sink_default_init (GumEventSinkInterface * iface)
@@ -60,3 +78,42 @@ gum_event_sink_stop (GumEventSink * self)
   if (iface->stop != NULL)
     iface->stop (self);
 }
+
+GumEventSink *
+gum_event_sink_make_default (void)
+{
+  return GUM_EVENT_SINK (g_object_new (GUM_TYPE_DEFAULT_EVENT_SINK, NULL));
+}
+
+static void
+gum_default_event_sink_class_init (GumDefaultEventSinkClass * klass)
+{
+}
+
+static void
+gum_default_event_sink_iface_init (gpointer g_iface,
+                                   gpointer iface_data)
+{
+  GumEventSinkInterface * iface = g_iface;
+
+  iface->query_mask = gum_default_event_sink_query_mask;
+  iface->process = gum_default_event_sink_process;
+}
+
+static void
+gum_default_event_sink_init (GumDefaultEventSink * self)
+{
+}
+
+static GumEventType
+gum_default_event_sink_query_mask (GumEventSink * sink)
+{
+  return GUM_NOTHING;
+}
+
+static void
+gum_default_event_sink_process (GumEventSink * sink,
+                                const GumEvent * ev)
+{
+}
+
--- a/gum/gumeventsink.h
+++ b/gum/gumeventsink.h
@@ -16,6 +16,10 @@ G_BEGIN_DECLS
 #define GUM_TYPE_EVENT_SINK (gum_event_sink_get_type ())
 G_DECLARE_INTERFACE (GumEventSink, gum_event_sink, GUM, EVENT_SINK, GObject)
 
+#define GUM_TYPE_DEFAULT_EVENT_SINK (gum_default_event_sink_get_type ())
+G_DECLARE_FINAL_TYPE (GumDefaultEventSink, gum_default_event_sink, GUM,
+    DEFAULT_EVENT_SINK, GObject)
+
 struct _GumEventSinkInterface
 {
   GTypeInterface parent;
@@ -32,6 +36,8 @@ GUM_API void gum_event_sink_start (GumEventSink * self);
 GUM_API void gum_event_sink_process (GumEventSink * self, const GumEvent * ev);
 GUM_API void gum_event_sink_flush (GumEventSink * self);
 GUM_API void gum_event_sink_stop (GumEventSink * self);
+
+GUM_API GumEventSink * gum_event_sink_make_default (void);
 
 G_END_DECLS
 
--- a/tests/core/arch-arm/stalker-arm.c
+++ b/tests/core/arch-arm/stalker-arm.c
@@ -92,6 +92,7 @@ TESTLIST_BEGIN (stalker)
   TESTENTRY (unfollow_should_be_allowed_before_second_transform)
   TESTENTRY (unfollow_should_be_allowed_mid_second_transform)
   TESTENTRY (unfollow_should_be_allowed_after_second_transform)
+  TESTENTRY (follow_me_should_support_nullable_event_sink)
 
   TESTENTRY (follow_syscall)
   TESTENTRY (follow_thread)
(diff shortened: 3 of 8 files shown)


meme:
I agree that if-checks everywhere is not great, I was getting nervous that I may have missed some... I will re-work this PR with NullSink as I agree that is more suitable.

meme:
Integrated a `DefaultEventSink` and I agree it makes much more sense. The question is now: how should we be testing this considering there is a bit of overlap with other Stalker tests?

gum/gumeventsink.h
```diff
@@ -16,6 +16,10 @@ G_BEGIN_DECLS
 #define GUM_TYPE_EVENT_SINK (gum_event_sink_get_type ())
 G_DECLARE_INTERFACE (GumEventSink, gum_event_sink, GUM, EVENT_SINK, GObject)
 
+#define GUM_TYPE_DEFAULT_EVENT_SINK (gum_default_event_sink_get_type ())
+G_DECLARE_FINAL_TYPE (GumDefaultEventSink, gum_default_event_sink, GUM,
+                      DEFAULT_EVENT_SINK, GObject)
```
oleavr, line 21:
```suggestion
    DEFAULT_EVENT_SINK, GObject)
```

Hanging indents should be 4 spaces. (Just realised we're already a bit inconsistent when using these macros, ugh.)

gum/gumeventsink.h
```diff
@@ -33,6 +37,8 @@ GUM_API void gum_event_sink_process (GumEventSink * self, const GumEvent * ev);
 GUM_API void gum_event_sink_flush (GumEventSink * self);
 GUM_API void gum_event_sink_stop (GumEventSink * self);
 
+GUM_API GumEventSink* gum_event_sink_make_default (void);
```
oleavr, line 40:
```suggestion
GUM_API GumEventSink * gum_event_sink_make_default (void);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -25,6 +25,7 @@ TESTLIST_BEGIN (stalker)
   TESTENTRY (unfollow_should_be_allowed_before_second_transform)
   TESTENTRY (unfollow_should_be_allowed_mid_second_transform)
   TESTENTRY (unfollow_should_be_allowed_after_second_transform)
+  TESTENTRY (follow_me_nullable_event_sink)
```
oleavr, line 28:
```suggestion
  TESTENTRY (follow_me_should_support_nullable_event_sink)
```

gum/gumeventsink.c
```diff
@@ -60,3 +79,40 @@ gum_event_sink_stop (GumEventSink * self)
   if (iface->stop != NULL)
     iface->stop (self);
 }
+
+static void gum_default_event_sink_class_init (GumDefaultEventSinkClass* klass)
```
oleavr, line 83:
```suggestion
static void
gum_default_event_sink_class_init (GumDefaultEventSinkClass * klass)
```

(Here and elsewhere.)

oleavr (changes requested):
Yay, thanks! 🤘

meme:
Presumably these tests should be ported to the respective architectures, I can make them but I do not have the hardware to test all platforms currently. Is this something you'll do post-review, or is there a CI I can test it on?

oleavr:
> Presumably these tests should be ported to the respective architectures, I can make them but I do not have the hardware to test all platforms currently. Is this something you'll do post-review, or is there a CI I can test it on?

That would be great! Our CI doesn't yet run tests on those platforms, but it should be fairly safe to transplant the exact same code due to how similar the implementations are.

gum/gumeventsink.c
```diff
@@ -6,8 +6,27 @@
 
 #include "gumeventsink.h"
 
+struct _GumDefaultEventSink
+{
+  GObject parent;
+};
+
+static void gum_default_event_sink_iface_init (gpointer g_iface,
+                                               gpointer iface_data);
```
oleavr, line 15: Hanging indents should be 4 spaces.

gum/gumeventsink.c
```diff
@@ -6,8 +6,27 @@
...
+static void gum_default_event_sink_finalize (GObject * obj);
+static GumEventType gum_default_event_sink_query_mask (GumEventSink * sink);
+static void gum_default_event_sink_process (GumEventSink * sink,
+                                            const GumEvent * ev);
```
oleavr, line 19: Hanging indents should be 4 spaces.

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
...
+
+static void
+gum_default_event_sink_class_init (GumDefaultEventSinkClass * klass)
+{
+  GObjectClass * object_class = G_OBJECT_CLASS (klass);
+  object_class->finalize = gum_default_event_sink_finalize;
+}
+
+static void
+gum_default_event_sink_iface_init (gpointer g_iface, gpointer iface_data)
```
oleavr, line 91: Style violation, should only be one argument per line.

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
   if (iface->stop != NULL)
     iface->stop (self);
 }
+
+static void
+gum_default_event_sink_class_init (GumDefaultEventSinkClass * klass)
+{
+  GObjectClass * object_class = G_OBJECT_CLASS (klass);
+  object_class->finalize = gum_default_event_sink_finalize;
```
oleavr, line 87:
```suggestion

  object_class->finalize = gum_default_event_sink_finalize;
```

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
...
+}
+
+static void
+gum_default_event_sink_iface_init (gpointer g_iface, gpointer iface_data)
+{
+  GumEventSinkInterface * iface = g_iface;
+  iface->query_mask = gum_default_event_sink_query_mask;
```
oleavr, line 94:
```suggestion

  iface->query_mask = gum_default_event_sink_query_mask;
```

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
...
+static void
+gum_default_event_sink_init (GumDefaultEventSink * self)
+{
+}
+
+static void
+gum_default_event_sink_finalize (GObject * object)
+{
+  G_OBJECT_CLASS (gum_default_event_sink_parent_class)->finalize (object);
+}
```
oleavr, line 107: If a `dispose` or `finalize` implementation does nothing but chain up, it should be omitted.

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
...
+
+static GumEventType
+gum_default_event_sink_query_mask(GumEventSink * sink)
```
oleavr, line 110:
```suggestion
gum_default_event_sink_query_mask (GumEventSink * sink)
```

gum/gumeventsink.c
```diff
@@ -60,3 +79,46 @@ gum_event_sink_stop (GumEventSink * self)
...
+{
+  return GUM_NOTHING;
+}
+
+static void
+gum_default_event_sink_process (GumEventSink * sink, const GumEvent * ev)
```
oleavr, line 116: One argument per line (see earlier comment).

oleavr (changes requested):

gum/gumeventsink.c
```diff
@@ -60,3 +78,41 @@ gum_event_sink_stop (GumEventSink * self)
...
+}
+
+static void
+gum_default_event_sink_process (GumEventSink * sink,
+                                const GumEvent * ev)
+{
+}
+
+GumEventSink *
+gum_event_sink_make_default (void)
```
oleavr, line 115: Let's move this right before `gum_default_event_sink_class_init()` as this is higher level so it belongs higher up.

tests/core/arch-arm/stalker-arm.c
```diff
@@ -2505,6 +2506,15 @@ unfollow_during_transform (GumStalkerIterator * iterator,
   ctx->num_blocks_transformed++;
 }
 
+TESTCASE (follow_me_should_support_nullable_event_sink) {
```
oleavr, line 2509:
```suggestion
TESTCASE (follow_me_should_support_nullable_event_sink)
{
```

(Here and the in the corresponding code for the other archs.)

oleavr (changes requested):

oleavr (approved):
Yay! 🙌

PR #489, opened 2020-09-17
Fix arm64e support in relocator

mrmacete:
By checking also for arm64e variants of RET and BR.


oleavr (approved):
Thanks, great catch! 🤘

PR #499, opened 2020-11-05
Fix deadlock when the thread holding the JS lock crashes

mrmacete:
- use the exceptor to detect this condition
- when that happens, the scope mutex is marked as "trapped" in the backend
    - in this condition the mutex remains locked, but `with_lock_held` won't block
    - so that the `ThreadSuspendMonitor` doesn't create deadlocks
    - also, all JS exception handlers are skipped in this state

This will allow crashes of this kind to actually surface instead of causing a deadlock. 

This can be improved over time, but for how this is implemented right now it's not really possible to "catch" and handle this kind of exceptions, because that would require forcibly unlock the scope mutex even if it belongs to a crashed thread.

The test case runs in "slow" mode only

```
gum-tests -m slow -p /GumJS/Script/crash_on_thread_holding_js_lock_should_not_deadlock#V8
gum-tests -m slow -p /GumJS/Script/crash_on_thread_holding_js_lock_should_not_deadlock#QJS
```

### DIFF
--- a/bindings/gumjs/gumquickcore.c
+++ b/bindings/gumjs/gumquickcore.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2020 Ole André Vadla Ravnås <oleavr@nowsecure.com>
+ * Copyright (C) 2020 Francesco Tamagni <mrmacete@protonmail.ch>
  *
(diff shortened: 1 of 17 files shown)


oleavr (approved):
Yay, this is awesome! Thanks!! 🚀

PR #507, opened 2020-11-20
Added support for 'size_t' and 'ssize_t' (as discussed privately)

mame82:
- 'size_t' is converted to UInt64 for JS runtime
- 'ssize_t' is converted to Int64 for JS runtime
- for usage see test cases


oleavr:
Landed in 35918aed65b9aca4d7f1676e8446de028930d035 with some minor tweaks. Thanks!

PR #509, opened 2020-11-23
Bug 508

WorksButNotTested:


WorksButNotTested:
This PR also included a check of the CRC32 of the compressed data when running the pretend_workload.

Issue #510, opened 2020-11-23
GCC support in CModule


mephi42:
If one wants to use CModule on architectures that aren't supported by TCC, one option is to use the GNU toolchain (gcc+ld+objcopy+nm) for code generation and inspection instead. This may also be useful in cases where good optimization is desired.

I have a working prototype at https://github.com/mephi42/frida-gum/tree/gcc, where just a couple tests still fail.
Would it be interesting to have this upstream?

oleavr:
Yeah I have actually been thinking about this occasionally so am super-excited to see your prototype. 🎉 Would love to support this upstream!

PR #511, opened 2020-11-24
GCC support in CModule

mephi42:
This PR makes it possible to build CModules with GCC (issue #510).

It can be activated in two ways:

- Using Frida on architecture that is not supported by TCC.
- Passing `'gcc'` as the third `CModule` constructor parameter.

I've been using the following tweak for testing purposes (all tests pass with it on x86_64 linux):

```
diff --git a/bindings/gumjs/gumcmodule.c b/bindings/gumjs/gumcmodule.c
index f27c1246..1b483df2 100644
--- a/bindings/gumjs/gumcmodule.c
+++ b/bindings/gumjs/gumcmodule.c
@@ -881,6 +881,7 @@ gum_cmodule_gcc_ops = {
 const GumCModuleOps *
 gum_cmodule_get_ops (const gchar * name)
 {
+  return &gum_cmodule_gcc_ops;
 #ifdef HAVE_TINYCC
   if (!name || strcmp (name, "tcc") == 0)
     return &gum_cmodule_tcc_ops;
```

The code works as follows:

- Since switching between TCC and GCC needs to be done during run time, `struct ops` scheme is used.
- The new `gum_cmodule_get_ops` function resolves `struct ops` by name; `gum_cmodule_new` now takes an additional `struct ops` parameter.
- Other `gum_cmodule_*` functions are unchanged, internally they do common processing and call TCC- or GCC-specific code via `struct ops`.
- GCC processing is done in a temporary directory. During compilation it contains user's C code (`module.c`) and Frida headers. The compiler must make sure that calls to Frida API are done directly and not using GOT/PLT scheme. On Intel this is achieved by using `-mcmodel=large`, other architectures might need to add similar options.
- During linking a linker script (`module.lds`) is added. It makes sure that the resulting ELF binary (`a.out`) is self-contained and calls to Frida API are resolved to their runtime addresses.
- Finally, objcopy is used in order to rip raw machine code (`module`) from the ELF binary.
- If the user wants to know what symbols are available, then nm is used.
- At the end the entire directory is removed.


bindings/gumjs/gumcmodule.c
```diff
@@ -8,6 +8,91 @@
...
+  void (*gum_cmodule_add_symbol) (GumCModule *, const gchar *, gconstpointer);
+  gint (*gum_cmodule_link_pre) (GumCModule *, GString **);
+  gint (*gum_cmodule_link) (GumCModule *, gpointer, GString **);
+  void (*gum_cmodule_link_post) (GumCModule *);
+  void (*gum_cmodule_enumerate_symbols) (GumCModule *, GumFoundCSymbolFunc,
+      gpointer);
+  gpointer (*gum_cmodule_find_symbol_by_name) (GumCModule *, const gchar *);
+  void (*gum_cmodule_drop_metadata) (GumCModule *);
+  void (*gum_cmodule_add_define) (GumCModule *, const gchar *, const gchar *);
+};
```
oleavr, line 25: It would be good to define a GObject interface instead, and implement it in two different GObject-classes. For an example of an interface see `gum/gumbacktracer.h` and `gum/gumbacktracer.c`.

oleavr (changes requested):
Thanks, this is awesome!

Once this has been migrated to use a GObject interface, it should be good to merge after tweaking the style a little bit. We have a style-guide [here](https://github.com/frida/frida/blob/master/CONTRIBUTING.md).

bindings/gumjs/gumcmodule.c
mephi42: Should GObject be the new API, or can it live purely inside gumcmodule.c as an implementation detail?

oleavr: I think keeping it inside is fine. Can be split into multiple files later if we end up adding more backends.

oleavr:
Last part of this just landed in 667767974e99e433c7da6c4862dc6349c793bd11 🎉 🙌

PR #513, opened 2020-11-26
Fix ARM Stalker following of threads running Thumb code

WorksButNotTested:


oleavr:
Thanks! 😍 Landed as part of #514.

PR #515, opened 2020-11-28
Fix linking during .gir generation

meme:
During set-up I ran into another bug that I somehow missed in the original PR.

When linking against `libelf` and `libdwarf` using `extra_libs_private` it will only add `-lelf` and `-ldwarf`. However this apparently doesn't do what I thought it did so you'll get a bunch of errors for missing symbols in `libelf`.

It turns out that `g-ir-scanner` expects _instead_ to receive e.g.: `--extra-library=elf`. By using `dependency` and having it link against `gum` (via `extra_deps`) Meson will automatically add `--extra-library=elf`.

So we can either use a `dependency`, or we can add the `--extra-library=` flags manually into `extra_libs_private`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -193,6 +193,7 @@ lzma_dep = dependency('liblzma')
 extra_deps = []
 extra_requires_private = []
 extra_libs_private = []
+extra_gir_args_private = []
 
 glib_flavor_src = '''
 #include <glib.h>
@@ -231,18 +232,17 @@ elif host_os_family == 'linux' or host_os_family == 'qnx'
 endif
 
 if host_os_family == 'linux' or host_os_family == 'qnx'
-  have_libelf = cc.has_header('libelf.h')
-  if not have_libelf
-    error('libelf is required')
-  endif
+  elf_dep = dependency('libelf')
+  extra_deps += [elf_dep]
   have_libdwarf = cc.has_header('libdwarf.h')
   if not have_libdwarf
     error('libdwarf is required')
   endif
   if host_os_family == 'linux'
-    extra_libs_private += ['-ldwarf', '-lelf', '-ldl', '-lz']
+    extra_libs_private += ['-ldwarf', '-ldl', '-lz']
+    extra_gir_args_private += ['--extra-library=dwarf']
   else
-    extra_libs_private += ['-ldwarf', '-lelf', '-lz']
+    extra_libs_private += ['-ldwarf', '-lz']
   endif
 else
   have_libdwarf = false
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -244,6 +244,11 @@ if host_os_family == 'linux' or host_os_family == 'qnx'
   else
     extra_libs_private += ['-ldwarf', '-lelf', '-lz']
   endif
+  if have_shared_glib
+    dwarf_dep = dependency('libdw')
+    elf_dep = dependency('libelf')
```
oleavr, line 249: I wonder if the cleanest solution here is to fix our build recipes (in `releng/deps.mk`) so these two come with `.pc`-files in our SDK -- that way we can drop our ugly hacks here and simply use `dependency()` unconditionally. What do you think?

oleavr (changes requested):
Thanks!

meson.build
meme: That sounds much better~ I'll take a look at this instead

meme:
Unless I'm mistaken, it looks like there are `pkg-config` files:
```
$ find . | grep -E "lib(elf|dw).pc"                                                                                                                        13:04:39
./build/fs-tmp-linux-x86_64/elfutils/config/libelf.pc
./build/fs-tmp-linux-x86_64/elfutils/config/libdw.pc
```

I've pushed up the changes to unconditionally use `pkg-config`. Need to do some more testing locally.

meme:
Hmm... so this works, but I've got some confusion which could be a problem. On Arch Linux when I install `elfutils` it installs a `libdw.pc` file which points to `libdw`, which is NOT `libdwarf`. Frida requires `libdwarf`, however so there are some missing symbols to, e.g.: `dwarf_dealloc`. Should we just make it clear which library is expected?

This is very unfortunate and confusing that these libraries have a naming conflict...

oleavr:
> Should we just make it clear which library is expected?

On Ubuntu the package name is `libdwarf-dev`, but it doesn't bundle a `.pc`. So I think we should assume `-ldwarf` here.

It would be good to consume `libelf` through its `.pc` though, and improve our SDK build recipe to include it. (Like you did in your other PR 👌.)

meme:
https://github.com/frida/frida/pull/1557 will add the `pkg-config` file for `libelf`. And this PR uses _that_ `.pc` file and then uses the original code for checking for `libdwarf` and uses `--extra-library=dwarf`

oleavr:
Thanks! 🤘

PR #517, opened 2020-12-14
fixes Frida issue #1558

ant9000:


oleavr:
Thanks a lot, great catch! 🚀

PR #519, opened 2020-12-15
GCC support in CModule [2/N]

mephi42:
Hi! I finally managed to wrap my head around GObject. This patch is a rewrite of https://github.com/frida/frida-gum/pull/511/commits/8775940ad25bedd48baf2817db645d2bc4a6170d, which enables adding new compiler backends in addition to TCC.

### DIFF
--- a/bindings/gumjs/gumcmodule.c
+++ b/bindings/gumjs/gumcmodule.c
@@ -6,7 +6,24 @@
 
 #include "gumcmodule.h"
 
+#include <glib-object.h>
 #include <gio/gio.h>
+
+struct _GumCModuleClass
+{
+  GObjectClass parent_class;
+
+  const GumMemoryRange * (* get_range) (GumCModule *);
+  void (* add_symbol) (GumCModule *, const gchar *, gconstpointer);
+  gboolean (* link) (GumCModule *, GError **);
+  void (* enumerate_symbols) (GumCModule *, GumFoundCSymbolFunc, gpointer);
+  gpointer (* find_symbol_by_name) (GumCModule *, const gchar *);
+  void (* drop_metadata) (GumCModule *);
+};
+
+#define GUM_TYPE_CMODULE (gum_cmodule_get_type ())
+G_DECLARE_DERIVABLE_TYPE (GumCModule, gum_cmodule, GUM, CMODULE, GObject)
+G_DEFINE_ABSTRACT_TYPE (GumCModule, gum_cmodule, G_TYPE_OBJECT)
 
 #ifdef HAVE_TINYCC
 
@@ -23,8 +40,10 @@ typedef guint GumCModuleHeaderKind;
 typedef void (* GumCModuleInitFunc) (void);
 typedef void (* GumCModuleFinalizeFunc) (void);
 
-struct _GumCModule
+struct _GumTccCModule
 {
+  GumCModule parent;
+
   TCCState * state;
   GumMemoryRange range;
   GumCModuleFinalizeFunc finalize;
@@ -50,6 +69,19 @@ struct _GumEnumerateSymbolsContext
   gpointer user_data;
 };
 
+static GumCModule * gum_tcc_cmodule_new (const gchar * source,
+    GError ** error);
+static void gum_tcc_cmodule_finalize (GObject * object);
+static const GumMemoryRange * gum_tcc_cmodule_get_range (GumCModule * cm);
+static void gum_tcc_cmodule_add_symbol (GumCModule * cm, const gchar * name,
+    gconstpointer value);
+static gboolean gum_tcc_cmodule_link (GumCModule * cm, GError ** error);
+static void gum_tcc_cmodule_enumerate_symbols (GumCModule * cm,
+    GumFoundCSymbolFunc func, gpointer user_data);
+static gpointer gum_tcc_cmodule_find_symbol_by_name (GumCModule * cm,
+    const gchar * name);
+static void gum_tcc_cmodule_drop_metadata (GumCModule * cm);
+static void gum_emit_symbol (void * ctx, const char * name, const void * val);
 static void gum_append_tcc_error (void * opaque, const char * msg);
 static void gum_emit_symbol (void * ctx, const char * name, const void * val);
 static const char * gum_cmodule_load_header (void * opaque, const char * path,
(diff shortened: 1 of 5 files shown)


bindings/gumjs/gumcmodule.h
```diff
@@ -23,7 +23,8 @@ struct _GumCSymbolDetails
 typedef void (* GumFoundCSymbolFunc) (const GumCSymbolDetails * details,
     gpointer user_data);
 
-GUM_API GumCModule * gum_cmodule_new (const gchar * source, GError ** error);
+GUM_API GumCModule * gum_cmodule_new (const gchar * name, const gchar * source,
+    GError ** error);
```
oleavr, line 27: How about:

```suggestion
GUM_API GumCModule * gum_cmodule_new (const gchar * toolchain,
    const gchar * source, GError ** error);
```

bindings/gumjs/gumcmodule.c
```diff
@@ -396,51 +479,62 @@ gum_cmodule_new (const gchar * source,
...
+
+  return GUM_CMODULE_GET_CLASS (self)->find_symbol_by_name (self, name);
 }
 
 void
 gum_cmodule_drop_metadata (GumCModule * self)
 {
-  g_assert_not_reached ();
-}
+  g_assert (self != NULL);
```
oleavr, line 537: I think we can omit assertions like these as passing `NULL` will result in a crash on the next line.

bindings/gumjs/gumcmodule.c
```diff
@@ -382,12 +449,28 @@ gum_undecorate_name (const gchar * name)
...
+}
 
 GumCModule *
-gum_cmodule_new (const gchar * source,
+gum_cmodule_new (const gchar * name,
+                 const gchar * source,
                  GError ** error)
 {
+#ifdef HAVE_TINYCC
+  if (!name || strcmp (name, "tcc") == 0)
```
oleavr, line 470:
```suggestion
  if (name == NULL || strcmp (name, "tcc") == 0)
```

To follow the current style of making all comparisons explicit. (Truthy checks should only be performed on booleans.)

bindings/gumjs/gumcmodule.c
```diff
@@ -173,33 +232,37 @@ gum_cmodule_free (GumCModule * cmodule)
...
-  g_slice_free (GumCModule, cmodule);
+  G_OBJECT_CLASS (gum_tcc_cmodule_parent_class)->dispose (_cmodule);
 }
 
-const GumMemoryRange *
-gum_cmodule_get_range (GumCModule * self)
+static const GumMemoryRange *
+gum_tcc_cmodule_get_range (GumCModule * _self)
 {
+  GumTccCModule * self = (GumTccCModule *) _self;
```
oleavr, line 243:
```suggestion
gum_tcc_cmodule_get_range (GumCModule * cm)
{
  GumTccCModule * self = GUM_TCC_CMODULE (cm);
```

This is the convention I've been using so far. Also good to use cast macros so that type checks can be performed when built with `-Db_ndebug=false` in `config.mk`.

bindings/gumjs/gumcmodule.c
```diff
@@ -140,28 +201,26 @@ gum_cmodule_new (const gchar * source,
...
+    gum_cmodule_free (_cmodule);
 
     return NULL;
   }
 }
 
-void
-gum_cmodule_free (GumCModule * cmodule)
+static void
+gum_tcc_cmodule_dispose (GObject * _cmodule)
```
oleavr, line 219:
```suggestion
gum_tcc_cmodule_dispose (GObject * object)
```

bindings/gumjs/gumcmodule.c
```diff
@@ -63,21 +95,50 @@ static const gchar * gum_undecorate_name (const gchar * name);
...
 
-GumCModule *
-gum_cmodule_new (const gchar * source,
-                 GError ** error)
+static GumCModule *
+gum_tcc_cmodule_new (const gchar * source,
+                     GError ** error)
 {
-  GumCModule * cmodule;
+  GumCModule * _cmodule;
```
oleavr, line 129:
```suggestion
  GumCModule * result;
```

bindings/gumjs/gumcmodule.c
```diff
@@ -63,21 +95,50 @@ static const gchar * gum_undecorate_name (const gchar * name);
...
   gchar * combined_source;
 
-  cmodule = g_slice_new0 (GumCModule);
+  _cmodule = g_object_new (GUM_TYPE_TCC_CMODULE, NULL);
+  cmodule = GUM_TCC_CMODULE (_cmodule);
 
   state = tcc_new ();
   cmodule->state = state;
+  memset (&cmodule->range, 0, sizeof (cmodule->range));
+  cmodule->finalize = NULL;
```
oleavr, line 141: No need to clear to zero – GObject instances are zero-initialized.

bindings/gumjs/gumcmodule.c
```diff
@@ -63,21 +95,50 @@ static const gchar * gum_undecorate_name (const gchar * name);
...
+    GumCModule)
+G_DEFINE_TYPE (GumTccCModule, gum_tcc_cmodule, G_TYPE_OBJECT)
+
+static void
+gum_tcc_cmodule_class_init (GumTccCModuleClass * klass)
+{
+  GObjectClass * object_class = G_OBJECT_CLASS (klass);
+  GumCModuleClass * cmodule_class = GUM_CMODULE_CLASS (klass);
+
+  object_class->dispose = gum_tcc_cmodule_dispose;
```
oleavr, line 109:
```suggestion
  object_class->dispose = gum_tcc_cmodule_dispose;

```

bindings/gumjs/gumcmodule.c
```diff
@@ -63,21 +95,50 @@ static const gchar * gum_undecorate_name (const gchar * name);
 extern void * __va_arg (void * ap, int arg_type, int size, int align);
 #endif
 
+#define GUM_TYPE_TCC_CMODULE (gum_tcc_cmodule_get_type ())
+G_DECLARE_FINAL_TYPE (GumTccCModule, gum_tcc_cmodule, GUM, TCC_CMODULE,
+    GumCModule)
+G_DEFINE_TYPE (GumTccCModule, gum_tcc_cmodule, G_TYPE_OBJECT)
```
oleavr, line 101:
```suggestion
G_DEFINE_TYPE (GumTccCModule, gum_tcc_cmodule, GUM_TYPE_CMODULE)
```

So the parent class is `Gum.CModule` instead of `GLib.Object`.

bindings/gumjs/gumcmodule.c
```diff
@@ -6,8 +6,25 @@
...
+  void (* add_symbol) (GumCModule *, const gchar *, gconstpointer);
+  gboolean (* link) (GumCModule *, GError **);
+  void (* enumerate_symbols) (GumCModule *, GumFoundCSymbolFunc, gpointer);
+  gpointer (* find_symbol_by_name) (GumCModule *, const gchar *);
+  void (* drop_metadata) (GumCModule *);
+};
+
+#define GUM_TYPE_CMODULE (gum_cmodule_get_type ())
+G_DECLARE_DERIVABLE_TYPE (GumCModule, gum_cmodule, GUM, CMODULE, GObject)
+G_DEFINE_ABSTRACT_TYPE (GumCModule, gum_cmodule, G_TYPE_OBJECT)
```
oleavr, line 26: 💥

oleavr (changes requested):
Thanks, this is awesome! 😍  Only some small tweaks needed:

mephi42:
Thanks for the quick and thorough review :+1: I hope I didn't miss anything!

bindings/gumjs/gumcmodule.c
```diff
@@ -396,51 +478,50 @@ gum_cmodule_new (const gchar * source,
 void
 gum_cmodule_free (GumCModule * cmodule)
```
oleavr, line 479: We should delete this and use `g_object_unref ()` instead. (And `g_clear_object (&p)` when needing to discard in nullable cases.)

bindings/gumjs/gumcmodule.c
```diff
@@ -173,33 +231,37 @@ gum_cmodule_free (GumCModule * cmodule)
     gum_memory_free (GSIZE_TO_POINTER (r->base_address), r->size);
   }
```
oleavr, line 232: `::dispose()` needs to handle being run more than once. It should however be used for releasing references to other objects. Memory should be freed in `::finalize()`, and that can only run once. So here we should probably be using `::finalize()`.

oleavr (changes requested):
Yay! 🎉   You didn't miss anything, but I did 😅  But I think these are the last two:

bindings/gumjs/gumcmodule.c
mephi42: I used `g_clear_object` where possible - this might be unnecessary in some locations, but I think the overhead is negligible.

oleavr (approved):
Thanks!

PR #522, opened 2020-12-27
docs: fix simple typo, continguous -> contiguous

timgates42:
There is a small typo in gum/dlmalloc.c.

Should read `contiguous` rather than `continguous`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


oleavr:
Thanks! This should be submitted upstream to dlmalloc. We don't want to maintain any non-essential patches on top of it, so I'm afraid I'll have to reject this PR, even if I like its contents.

PR #524, opened 2020-12-31
Fix mishandled relocation of rip relative indirect call in x86_64

dkw72n:
`FF 15 12 34 56 78` behave differently under x86 and x86_64.

x86:
```
0:  ff 15 12 34 56 78       call   DWORD PTR ds:0x78563412
```
x64:
```
0:  ff 15 12 34 56 78       call   QWORD PTR [rip+0x78563412]        # 0x78563418
```

In x86_64, it's a call to the value of a rip relative address, 
hence should be handled by the rip_relative* routine.

### DIFF
--- a/gum/arch-x86/gumx86relocator.c
+++ b/gum/arch-x86/gumx86relocator.c
@@ -565,6 +565,34 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
   if (!is_rip_relative)
     return FALSE;
 
+  if (insn->id == X86_INS_CALL || insn->id == X86_INS_JMP)
+  {
+    gint32 distance = *((gint32 *) (ctx->end - sizeof (gint32)));
+    guint64 * return_address_placeholder = NULL;
+
+    if (insn->id == X86_INS_CALL)
+    {
+      gum_x86_writer_put_push_reg (cw, GUM_REG_RAX);
+      gum_x86_writer_put_mov_reg_address (cw, GUM_REG_RAX, 0);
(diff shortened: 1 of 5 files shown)


gum/arch-x86/gumx86relocator.c
```diff
@@ -570,6 +577,8 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
       continue;
     else if (insn->id == X86_INS_CMPXCHG && cpu_regs[i] == GUM_REG_RAX)
       continue;
+    else if (insn->id == X86_INS_CALL && !(cpu_regs[i] == GUM_REG_RBX || cpu_regs[i] == GUM_REG_RBP))
```
oleavr, line 580: I'm probably missing something – what's the reason RBX and RBP are special-cased here?

oleavr (commented):
Great catch, thanks a lot! 🚀  Just one question before merging:

gum/arch-x86/gumx86relocator.c
oleavr: Ooh, I see, so it's picking either RBX or RBP as the `rip_reg`. (It's been a while since I wrote this code.) Hmm, I think we need to handle this slightly differently – we cannot move the stack pointer as that may be problematic if the called function expects arguments on the stack. Will try to massage this a bit further.

oleavr:
Tweaked version landed as 28213b457f91b892fbc499b1c91d2036061b8d27. Thanks a lot for this awesome PR! 🙌

PR #530, opened 2021-01-20
Add Stalker prefetch support for x86_64

WorksButNotTested:


tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  compiled_size_run2 = g_hash_table_size (compiled_run2);
+  g_print ("\tcompiled2: %d\n", compiled_size_run2);
+
+  executed_size_run2 = g_hash_table_size (executed_run2);
+  g_print ("\texecuted2: %d\n", executed_size_run2);
+
+  g_assert_cmpuint (compiled_size_run2, ==, 0);
+  g_assert_cmpuint (executed_size_run2, ==, executed_size_run1);
+
+
```
oleavr, line 501:
```suggestion
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  /* Free resources */
+  g_hash_table_destroy (compiled_run2);
+  g_hash_table_destroy (executed_run2);
+  g_hash_table_destroy (compiled_run1);
+  g_hash_table_destroy (executed_run1);
+
+  gum_stalker_unfollow_me (fixture->stalker);
+
+  g_print ("DONE\n");
+
```
oleavr, line 511:
```suggestion
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+}
```
oleavr, line 512: Are pipes leaked? (Any other resource leaks?)

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
+static void
+prefetch_transformer (GumStalkerIterator * iterator,
+                   GumStalkerOutput * output,
+                   gpointer user_data)
```
oleavr, line 517:
```suggestion
prefetch_transformer (GumStalkerIterator * iterator,
                      GumStalkerOutput * output,
                      gpointer user_data)
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  gboolean begin = TRUE;
+  gpointer block_address;
+  gpointer * callout_ctx;
+
+  while (gum_stalker_iterator_next (iterator, &instr))
+  {
+    if (begin)
+    {
+      block_address = GSIZE_TO_POINTER (instr->address);
+      callout_ctx = gum_malloc (sizeof (gpointer));
```
oleavr, line 529:
```suggestion
      callout_ctx = g_new (gpointer);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+      *callout_ctx = block_address;
+
+      gum_stalker_iterator_put_callout (iterator, prefetch_callout,
+          callout_ctx, gum_free);
```
oleavr, line 533:
```suggestion
          callout_ctx, g_free);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  gpointer block_address;
+  gpointer * callout_ctx;
+
+  while (gum_stalker_iterator_next (iterator, &instr))
+  {
+    if (begin)
+    {
+      block_address = GSIZE_TO_POINTER (instr->address);
+      callout_ctx = gum_malloc (sizeof (gpointer));
+      *callout_ctx = block_address;
```
oleavr, line 530: (But why the indirection if it's only one pointer?)

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
+      gum_stalker_iterator_put_callout (iterator, prefetch_callout,
+          callout_ctx, gum_free);
+
+      if (prefetch_compiled != NULL)
+      {
+        g_hash_table_insert (prefetch_compiled, block_address, block_address);
+      }
```
oleavr, line 538:
```suggestion
        g_hash_table_add (prefetch_compiled, block_address);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+      begin = FALSE;
```
oleavr, line 539:
```suggestion

      begin = FALSE;
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+prefetch_callout (GumCpuContext * cpu_context,
+                  gpointer user_data)
+{
+  gpointer * callout_ctx = user_data;
+  gpointer block_address = *callout_ctx;
+
+  if (prefetch_executed != NULL)
+  {
+    g_hash_table_insert (prefetch_executed, block_address, block_address);
+  }
```
oleavr, line 556:
```suggestion
    g_hash_table_add (prefetch_executed, block_address);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -81,6 +91,11 @@ TESTLIST_BEGIN (stalker)
 #endif
 TESTLIST_END ()
 
+#ifdef HAVE_LINUX
+static GHashTable * prefetch_compiled = NULL;
+static GHashTable * prefetch_executed = NULL;
+#endif
```
oleavr, line 97: Variables should go below function prototypes. So these should be moved to the bottom of the `HAVE_LINUX` block introduced below, with a blank line separating this and the function prototypes.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+}
+
+static void
+prefetch_run_child (GumStalker * stalker,
+                    GumMemoryRange * runner_range,
+                    int compile_fd,
+                    int execute_fd)
+{
+  pid_t pid;
+  int ret;
```
oleavr, line 566:
```suggestion
  int res;
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+    prefetch_write_blocks (compile_fd, prefetch_compiled);
+    prefetch_write_blocks (execute_fd, prefetch_executed);
+
+    g_hash_table_destroy (prefetch_compiled);
+    prefetch_compiled = NULL;
+    g_hash_table_destroy (prefetch_executed);
+    prefetch_executed = NULL;
+
+    exit (0);
+    return;
```
oleavr, line 591:
```suggestion
```

Unreachable statement.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
+  /* Wait for the child */
+  ret = waitpid (pid, &status, 0);
+  g_assert_cmpint (ret, ==, pid);
+  g_assert_cmpint (WIFEXITED (status), !=, 0);
+  g_assert_cmpint (WEXITSTATUS (status), ==, 0);
+}
+
+__attribute__((always_inline))
+static inline void
```
oleavr, line 602: Should be updated to match the example after updating it.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
+}
+
+static void
+prefetch_write_blocks (int fd,
+                       GHashTable * table)
+{
+  GHashTableIter iter;
+  gpointer iter_key, iter_value;
+  int ret;
```
oleavr, line 614:
```suggestion
  int res;
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
+  g_hash_table_iter_init (&iter, table);
+  while (g_hash_table_iter_next (&iter, &iter_key, &iter_value))
+  {
+    ret = write (fd, &iter_key, sizeof (gpointer));
+    g_assert_cmpint (ret, ==, sizeof (gpointer));
```
oleavr, line 620:
```suggestion
    g_assert_cmpint (ret, ==, sizeof (gpointer));
```

May fail because of partial writes (if the buffer is full), and theoretically also `EINTR`, so might be good to handle both in case users copy this example and use it in production code.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  }
+}
+
+static void
+prefetch_read_blocks (int fd,
+                      GHashTable * table)
+{
+  gpointer block_address;
+  while (read (fd, &block_address, sizeof (gpointer))
+    == sizeof (gpointer))
```
oleavr, line 630: Partial `read()` is also possible here I suppose, as well as `EINTR`.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  {
+    g_hash_table_insert (table, block_address, block_address);
```
oleavr, line 632:
```suggestion
    g_hash_table_add (table, block_address);
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  }
+}
+#endif
```
oleavr, line 635:
```suggestion

#endif
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+
 TESTCASE (no_events)
```
oleavr, line 637: All of this code should probably be moved down to after the `performance` test (and all its helper functions). Even if we try to follow newspaper ordering, this is an advanced use-case so probably not where we suggest readers start out reading to learn about how Stalker works.

oleavr (changes requested):
Yay, awesome! 🤘

WorksButNotTested:
Note to self. Add parameter for block cycle count.

WorksButNotTested:
Note to self. DON'T inline the activation target.

gum/gumstalker.h
```diff
@@ -110,6 +110,118 @@ GUM_API void gum_stalker_unfollow (GumStalker * self, GumThreadId thread_id);
...
+ *    instances upon completion if required.
+ *
+ * 4. Note that gum_stalker_activate takes a `target` pointer which is used to
+ *    allow stalker to be reactivated whilst executing in an excluded range and
+ *    guranatee that the thread is followed until the 'activation target'
+ *    address is reached. The simplest solution to this is to define an empty
+ *    function and then subsequently call it immediately after activation to
+ *    return stalker to its normal behaviour.
+ *
+ *    __attribute__((always_inline))
```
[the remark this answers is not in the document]
WorksButNotTested: I found this out the hard way earlier!

gum/gumstalker.h
```diff
@@ -110,6 +110,118 @@ GUM_API void gum_stalker_unfollow (GumStalker * self, GumThreadId thread_id);
...
+ * This API is intended for use during fuzzing scenarios such as AFL forkserver.
+ * It allows for the child to feed back instrumented blocks to the parent
+ * to be instrumented so that the next time a child is forked from the parent,
+ * it will already inherit the instrumented block rather than having to
+ * re-instrument every basic block again from scratch.
+ *
+ * This API has the following caveats:
+ *
+ * 1. This API is not marked GUM_API and is thus not available from JavaScript.
+ *    It is unlikely there will be scenarios where this is useful.
```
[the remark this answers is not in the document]
WorksButNotTested: Ok, I’ll add the GUM_API prefix back in

gum/gumstalker.h
```diff
@@ -110,6 +110,118 @@ GUM_API void gum_stalker_unfollow (GumStalker * self, GumThreadId thread_id);
...
+ *    Thus care should also be taken when any callouts are executed to determine
+ *    whether they are running the parent or child context.
+ *
+ * 6. For optimal performance, blocks should be prefetched `n` times, where `n`
+ *    is the trust threshold. Each time a block is executed, a cached copy
+ *    of the original bytes of the block is compared in order to detect
+ *    self-modifying code before a cached instrumented block is executed. The
+ *    trust threshold sets a limit of how many times the block should be
+ *    identical before this comparison step is skipped. Thus by reaching this
+ *    limit during prefetching, we avoid this comparison step in the child.
```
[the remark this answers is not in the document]
WorksButNotTested: Yeah. Let’s parametrise. The child should only perfetch if it has determined the block is not modified. Otherwise, multiple successive children may each partially increment the blocks recycle count to the trust threshold.

gum/gumstalker.h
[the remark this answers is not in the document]
WorksButNotTested: Oh you already have. Thanks.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Ok.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -375,6 +401,239 @@ invoke_flat (TestStalkerFixture * fixture,
...
+  trust = gum_stalker_get_trust_threshold (fixture->stalker);
+
+  /* Create IPC */
+  ret = pipe2 (compile_pipes, O_NONBLOCK);
+  g_assert_cmpint (ret, >=, 0);
+
+  ret = pipe2 (execute_pipes, O_NONBLOCK);
+  g_assert_cmpint (ret, >=, 0);
+
+  // fcntl (execute_pipes[STDIN_FILENO], F_SETPIPE_SZ, 1UL << 20);
```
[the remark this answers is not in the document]
WorksButNotTested: This should probably just be removed. The pipe defaults to 64kb. Which at 8 bytes per block (the address on x64) is enough for 8192 blocks. The test reports approx 1200. We can increase the size, but we must be careful not to hit the rlimit.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Well spotted. I think that's it? The hash tables are freed, and the transformer is set in the fixture which is tidied up when the fixture is torn down.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: I'll take a look at this one.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Will do.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Partial writes can be mitigated by specifying O_DIRECT in pipe2(), but I think it makes each message take up a whole page? man fcntl says:

"""

              When allocating the buffer for the pipe, the kernel may
              use a capacity larger than arg, if that is convenient for
              the implementation.  (In the current implementation, the
              allocation is the next higher power-of-two page-size
              multiple of the requested size.)  The actual capacity (in
              bytes) that is set is returned as the function result.
"""

So it seems unlikely we would hit a partial write from a full buffer? 

We should only get EINTR if the syscall is interrupted by signal delivery, we wouldn't expect this in the test harness. But it should be easy enough to handle, so let's just do that.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: See above. What are your thoughts?

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Yeah, I can move that.

tests/core/arch-x86/stalker-x86.c
WorksButNotTested: Should I add the comment about partial writes to the code?

oleavr: Yeah actually let's just ignore it and not comment it – shouldn't worry about folks expecting too much from tests, as they're allowed to make a lot of assumptions.

tests/core/arch-x86/stalker-x86.c
oleavr: Let's not bother, and keep this simple and without any fuzz :)

oleavr:
(Note that some of the suggestions rename variables but I didn't add suggestions to update references to those variables.)

gum/backend-arm/gumstalker-arm.c
```diff
@@ -928,6 +928,27 @@ _gum_stalker_do_deactivate (GumStalker * self,
...
 }
 
+void
+gum_stalker_prefetch (GumStalker * self,
+                      gpointer address,
+                      gint recycle_count)
+{
+  GumExecCtx * ctx;
+  gpointer code_address;
+  GumExecBlock * block;
```
oleavr, line 938:
```suggestion
  GumExecBlock * block;
  gpointer code_address;
```

oleavr (changes requested):
I might be missing something though:

gum/gumstalker.h
```diff
@@ -110,6 +110,122 @@ GUM_API void gum_stalker_unfollow (GumStalker * self, GumThreadId thread_id);
...
+ * This API is intended for use during fuzzing scenarios such as AFL forkserver.
+ * It allows for the child to feed back the addresses of instrumented blocks to
+ * the parent so that the next time a child is forked from the parent, it will
+ * already inherit the instrumented block rather than having to re-instrument
+ * every basic block again from scratch.
+ *
+ * This API has the following caveats:
+ *
+ * 1. This API is not intended to be used from JavaScript. It is unlikely there
+ *    will be scenarios where this is useful.
```
oleavr, line 123: I don't think we want to mention decisions made in language-bindings in Gum, as we're one layer below here and there should be a clear separation of concerns.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -2159,3 +2183,239 @@ test_window_proc (HWND hwnd, UINT msg, WPARAM wparam, LPARAM lparam)
...
+}
+
+static void
+prefetch_read_blocks (int fd,
+                      GHashTable * table)
+{
+  gpointer block_address;
+
+  while (read (fd, &block_address, sizeof (gpointer))
+    == sizeof (gpointer))
```
oleavr, line 2415: Hanging indents should be 4 spaces.

oleavr (changes requested):

PR #532, opened 2021-01-23
GCC support in CModule [4/N]

mephi42:
Here are the rest of the original patches ported to GObject and tweaked a little bit according to my (hopefully) improved understanding of the code style. What remains is the testing part, right now one can apply

```
--- a/bindings/gumjs/gumcmodule.c
+++ b/bindings/gumjs/gumcmodule.c
@@ -81,6 +81,8 @@ gum_cmodule_new (const gchar * name,
                  const gchar * source,
                  GError ** error)
 {
+  return gum_gcc_cmodule_new (source, error);
+
 #ifdef HAVE_TINYCC
   if (name == NULL || strcmp (name, "tcc") == 0)
     return gum_tcc_cmodule_new (source, error);
```

to run the tests, but I think it should be possible to detect whether GCC is installed and run both TCC and GCC tests in this case.

### DIFF
--- a/bindings/gumjs/gumcmodule.c
+++ b/bindings/gumjs/gumcmodule.c
@@ -603,6 +603,7 @@ G_DECLARE_FINAL_TYPE (GumGccCModule, gum_gcc_cmodule, GUM, GCC_CMODULE,
 struct _GumGccCModule
 {
   GumCModule parent;
+
   gchar * workdir;
   GPtrArray * argv;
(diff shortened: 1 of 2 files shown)


oleavr:
Wow, awesome! :boom: Aiming to land this later today. I think we'll want to revise the JS API a little bit, making the last argument an object of options so it's extensible. (I'll take care of this while merging.)

PR #534, opened 2021-02-09
Fix Instruction operands parsing on x86

muhzii:
the `op->imm` field is a 64 bit value in both variants.

### DIFF
--- a/bindings/gumjs/gumv8instruction.cpp
+++ b/bindings/gumjs/gumv8instruction.cpp
@@ -449,15 +449,8 @@ gum_parse_operands (const cs_insn * insn,
         break;
       case X86_OP_IMM:
         _gum_v8_object_set_ascii (element, type_key, "imm", core);
-        if (op->size <= 4)
-        {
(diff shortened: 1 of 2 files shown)


bindings/gumjs/gumv8instruction.cpp
```diff
@@ -449,15 +449,8 @@ gum_parse_operands (const cs_insn * insn,
...
         _gum_v8_object_set_ascii (element, type_key, "imm", core);
-        if (op->size <= 4)
-        {
-          _gum_v8_object_set_int (element, value_key, op->imm, core);
-        }
-        else
-        {
-          _gum_v8_object_set (element, value_key,
+        _gum_v8_object_set (element, value_key,
               _gum_v8_int64_new (op->imm, core), core);
```
oleavr, line 453:
```suggestion
        _gum_v8_object_set (element, value_key,
            _gum_v8_int64_new (op->imm, core), core);
```

(Hanging indents should be 4 spaces.)

oleavr (changes requested):

oleavr:
Thanks! 🙌

PR #536, opened 2021-02-18
Expose base64 and checksum glib primitives to CModule

mrmacete:
Motivation: this is exactly the kind of common things which are sensibly faster when done from a CModule.


oleavr:
Thanks! 🙌

PR #538, opened 2021-02-26
Stalker/x86: Also emit call probe on jmp insns

fabianfreyer:
Call probes do not work on imported functions called through the GOT, as
these are indirect jumps with the target taken from the PLT.

Also considering direct jumps as candidates for call probes additionally
allows for call probes to be triggered when called as tail-calls.

This has a performance impact at compile-time on the stalker, but should
have no performance impact at run-time.

### DIFF
--- a/gum/backend-x86/gumstalker-x86.c
+++ b/gum/backend-x86/gumstalker-x86.c
@@ -3079,6 +3079,12 @@ gum_exec_block_virtualize_branch_insn (GumExecBlock * block,
           is_false, GUM_NO_HINT);
     }
 
+    if (insn->ci->id == X86_INS_JMP)
+    {
+      if (ctx->stalker->any_probes_attached)
+        gum_exec_block_write_call_probe_code (block, &target, gc);
+    }
+
     if (target.is_indirect)
     {
       regular_entry_func = GUM_ENTRYGATE (jmp_mem);


gum/backend-x86/gumstalker-x86.c
```diff
@@ -3032,6 +3032,10 @@ gum_exec_block_virtualize_branch_insn (GumExecBlock * block,
     gum_x86_relocator_skip_one_no_label (gc->relocator);
     gum_exec_block_write_call_invoke_code (block, &target, gc);
   }
+  else if (insn->ci->id == X86_INS_JMP) {
```
oleavr, line 3035:
```suggestion
  else if (insn->ci->id == X86_INS_JMP)
  {
```

gum/backend-x86/gumstalker-x86.c
```diff
@@ -3032,6 +3032,10 @@ gum_exec_block_virtualize_branch_insn (GumExecBlock * block,
...
+    if (ctx->stalker->any_probes_attached)
+      gum_exec_block_write_call_probe_code (block, &target, gc);
```
oleavr, line 3037: We should move this block down to inside the final `else` [here](https://github.com/frida/frida-gum/blob/24bf115b777c30d8383e7d046ec867e0390d8b25/gum/backend-x86/gumstalker-x86.c#L3081) to avoid breaking instrumentation of `JMP` instructions.

oleavr (changes requested):
Thanks!

fabianfreyer:
Should be updated now

gum/backend-x86/gumstalker-x86.c
```diff
@@ -3064,6 +3064,12 @@ gum_exec_block_virtualize_branch_insn (GumExecBlock * block,
     gpointer is_false;
     GumExecCtxReplaceCurrentBlockFunc regular_entry_func, cond_entry_func;
 
+    if (insn->ci->id == X86_INS_JMP)
+    {
+      if (ctx->stalker->any_probes_attached)
+        gum_exec_block_write_call_probe_code (block, &target, gc);
+    }
```
oleavr, line 3071: We need to move this after the `if (is_conditional)` block below, so that we don't do the probing for conditional branches that are not taken.

oleavr (changes requested):

oleavr (approved):
Thanks!

PR #540, opened 2021-02-27
Stalker: rename and expose cache flushing routine

fabianfreyer:
Rename gum_stalker_invalidate_caches to gum_stalker_flush_caches to make
clear that this will invalidate the complete cache, as opposed to
a single basic block as planned with a future API.

Exposing this API means adding stub functions to the MIPS and ARM32
stalker implementations. On MIPS, Stalker is not supported, so just add
a similar stub implementation there. On ARM32, the invalidation logic
was not ported from ARM64 since it was not needed due to missing call
probe support. There, gum_stalker_flush_caches is also left as a stub
function, pending porting from ARM64.

### DIFF
--- a/gum/backend-arm/gumstalker-arm.c
+++ b/gum/backend-arm/gumstalker-arm.c
@@ -4466,3 +4466,9 @@ gum_is_exclusive_store_insn (const cs_insn * insn)
       return FALSE;
   }
 }
+
+void
+gum_stalker_flush_caches (GumStalker * self)
+{
+    g_warning ("Invalidation unsupported");
+}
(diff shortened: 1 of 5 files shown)


oleavr:
I think this is a great idea, but if we expose this we should also improve the implementation so we can release the slabs we've already allocated, and start fresh. The only caveat I can think of is that we also need to reset the side-stack.

oleavr:
Assuming this is covered by the invalidation feature that landed not that long ago. (Feel free to reopen if you'd like to revisit this at some point.) Cheers! 🙌

PR #544, opened 2021-04-14
Relax android linker path matching

asabil:
In container environments such as anbox, the linker path in /proc/self/maps doesn't always map 1-to-1 with the actual linker path. This commit relaxes the matching of the linker by using a regex.

### DIFF
--- a/gum/backend-linux/gumandroid.c
+++ b/gum/backend-linux/gumandroid.c
@@ -623,8 +623,9 @@ static const GumModuleDetails *
 gum_try_init_linker_details (void)
 {
   const GumModuleDetails * result = NULL;
+  gchar * linker_path;
   GRegex * linker_path_pattern;
-  gchar * linker_path, * maps, ** lines;
+  gchar * maps, ** lines;
   gint num_lines, vdso_index, i;
 
   linker_path = gum_find_linker_path ();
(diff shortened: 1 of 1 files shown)


gum/backend-linux/gumandroid.c
```diff
@@ -622,10 +623,12 @@ static const GumModuleDetails *
 gum_try_init_linker_details (void)
 {
   const GumModuleDetails * result = NULL;
+  GRegex * linker_path_pattern;
   gchar * linker_path, * maps, ** lines;
```
oleavr, line 627:
```suggestion
  gchar * linker_path;
  GRegex * linker_path_pattern;
  gchar * maps, ** lines;
```

(So they're ordered by first-use, except for the variable holding the return value, which goes first.)

gum/backend-linux/gumandroid.c
```diff
@@ -730,9 +734,27 @@ gum_find_linker_path (void)
...
+  const gchar * pattern;
+
+  if (gum_android_get_api_level () >= 29)
+    pattern = (sizeof (gpointer) == 4)
+      ? "/apex/com.android.runtime[^/]*/bin/linker$"
+      : "/apex/com.android.runtime[^/]*/bin/linker64$";
+  else 
+    pattern = (sizeof (gpointer) == 4)
+      ? "/system/bin/linker$"
+      : "/system/bin/linker64$";
```
oleavr, line 749:
```suggestion
  {
    pattern = (sizeof (gpointer) == 4)
        ? "/apex/com.android.runtime[^/]*/bin/linker$"
        : "/apex/com.android.runtime[^/]*/bin/linker64$";
  }
  else
  {
    pattern = (sizeof (gpointer) == 4)
        ? "/system/bin/linker$"
        : "/system/bin/linker64$";
  }
```

oleavr (changes requested):
Thanks! 💪

asabil:
Changes applied

oleavr (approved):
Thanks, looking good!

PR #545, opened 2021-04-20
Extract native libraries on android

P-Sc:
[Since Android Gradle plugin 3.6.0](https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs) `android:extractNativeLibs` is by default set to `false` which causes bundled native libraries (shared objects) to not be extracted to an extra file. Instead, Android extracts and loads them to memory at runtime.

As a consequence, frida cannot find these libraries under the given path and needs to extract them from the APK. Here is an example for such a path:
`/data/app/com.example.hellojni-G-r4G8zOhKSiGhnnvK7F0g==/base.apk!/lib/x86_64/libhello-jni.so`
Notice the apk path before the delimiting exclamation mark and the library path inside the zip after it.

This is related to 
- frida/frida#955
- frida/frida#1471 (though this needs fixing in Vala code)
- and possibly frida/frida#978

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -57,6 +57,9 @@
 /* Define to 1 if OS is macOS. */
 #mesondefine HAVE_MACOS
 
+/* Define to 1 if minizip is available. */
+#mesondefine HAVE_MINIZIP
+
 /* Define to 1 if target is MIPS based. */
 #mesondefine HAVE_MIPS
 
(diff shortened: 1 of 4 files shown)


oleavr:
Thanks a lot, this is awesome! 💥

It looks like the coding style needs some tweaks to follow the surrounding [style](https://github.com/frida/frida/blob/master/CONTRIBUTING.md), but other than that this looks great to me. (I can help out with the style tweaks once I've managed to escape the rabbithole that I'm in – just leaving this note here in case you get a chance to tweak it in the meantime.)

Cheers!

P-Sc:
Sorry, didn't see the style guide.

I fixed it accordingly.

oleavr:
@P-Sc Sorry for the delay here, got stuck in a rabbit-hole for months. Just rebased and made some tweaks. Going to take a stab at testing this next. Please let me know if you see any obvious mistakes in case you get a chance to look at this in the meantime 😊

oleavr:
Works great! Let's go ahead and land this. Cheers! 🙌

PR #546, opened 2021-04-21
Enumerate darwin imports from chained fixups

mrmacete:
In latest arm64e executables there are no binding opcodes, while usually the same information can be reconstructed by walking the metadata indexed by the LC_DYLD_CHAINED_FIXUPS command and stored in linkedit.

The problem in doing so on images which have already been loaded in memory by dyld is that the actual chains are lost because the bitfields in the slots have already been replaced by the actual pointers.

This change reconstructs the correct slots by scanning the pages of the segments which are pointed to by the fixups metadata and looking for pointers to resolved imports.

### DIFF
--- a/gum/gumdarwinmodule.c
+++ b/gum/gumdarwinmodule.c
@@ -131,6 +131,11 @@ struct _GumChainedFixupsHeader
   GumChainedSymbolFormat symbols_format;
 };
 
+#ifdef _MSC_VER
+# pragma warning (push)
+# pragma warning (disable: 4214)
+#endif
+
 enum _GumChainedImportFormat
 {
   GUM_CHAINED_IMPORT          = 1,
@@ -228,6 +233,10 @@ struct _GumChainedPtrArm64eAuthBind
           auth      :  1;
 };
 
+#ifdef _MSC_VER
+# pragma warning (pop)
+#endif
+
 static void gum_darwin_module_initable_iface_init (gpointer g_iface,
     gpointer iface_data);
 static void gum_darwin_module_constructed (GObject * object);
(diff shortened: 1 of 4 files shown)


gum/gumdarwinmodule.c
```diff
@@ -30,6 +33,23 @@ typedef struct _GumEmitTermPointersContext GumEmitTermPointersContext;
 
 typedef struct _GumExportsTrieForeachContext GumExportsTrieForeachContext;
 
+typedef struct _GumChainedFixupsHeader GumChainedFixupsHeader;
+
+typedef struct _GumChainedStartsInImage GumChainedStartsInImage;
+typedef struct _GumChainedStartsInSegment GumChainedStartsInSegment;
+
+typedef uint32_t GumChainedImportFormat;
+typedef uint32_t GumChainedSymbolFormat;
```
oleavr, line 42: For consistency we should be using GLib types here, as this code is also compiled on Windows.

gum/gumdarwinmodule.c
```diff
@@ -99,6 +120,114 @@ struct _GumExportsTrieForeachContext
...
+  GUM_CHAINED_IMPORT          = 1,
+  GUM_CHAINED_IMPORT_ADDEND   = 2,
+  GUM_CHAINED_IMPORT_ADDEND64 = 3,
+};
+
+struct _GumChainedImport
+{
+  uint32_t lib_ordinal :  8,
+           weak_import :  1,
+           name_offset : 23;
```
oleavr, line 145: This syntax may break MSVC compatibility, needs a closer look.

oleavr (changes requested):
Thanks!! 🤘 Just a super-shallow round of review first as I'm deep down in a different context right now:

gum/gumdarwinmodule.c
mrmacete: aww i didn't think about windows at all. i may need some help / advice on this topic.

PR #547, opened 2021-04-26
Fix enumerating modules with spaces on path

suy:
Closes https://github.com/frida/frida/issues/1710


oleavr:
Thanks, great catch! 💥  We also have the same issue in the [Linux injector](https://github.com/frida/frida-core/blob/master/src/linux/frida-helper-backend-glue.c) in frida-core in case you're interested in taking a quick stab at that too! 🤩

PR #550, opened 2021-05-20
Objective-C selectors fallback in DebugSymbol.fromAddress

hot3eed:
This adds ObjC method name resolution as a fallback to the `DebugSymbol.fromAddress` API in the case of a stripped binary, thanks to ObjC metadata functions and `LC_FUNCTION_STARTS`. (thanks for the tip, @mrmacete!)

Two points for discussion:
* Do we need to use a lazily initialized `GumObjCApiResolver` instance in `GumDarwinSymbolicator` for caching and performance purposes?
* I had to remove the `self->name == NULL` check in `gum/gumdarwinmodule.c` because I needed access to the image data in `gum_darwin_symbolicator_objc_details_from_address ()`. Is this bad?

- [x] Get last function's size
- [x] Lazily initialize  `GumObjcApiResolver` and cache it in `GumDarwinSymbolicator `
- [x] Address the `self->name == NULL` check
- [x] ~~Handle ObjC resolver unavailability edge case (see below)~~


gum/backend-darwin/gumdarwinsymbolicator.c
```diff
@@ -457,6 +485,123 @@ gum_darwin_symbolicator_find_functions_matching (GumDarwinSymbolicator * self,
...
+      &op);
+
+  dummy_function.address = address;
+  dummy_function.size = 0;
+
+  matched_function = bsearch (&dummy_function, op.functions->data,
+      op.functions->len, sizeof (GumCollectedFunction),
+      (GCompareFunc)gum_compare_collected_functions);
+
+  resolver = gum_api_resolver_make ("objc");
```
mrmacete, line 525: i think `resolver` could be created once lazily then stored in the `GumDarwinSymbolicator` instance (and disposed in the destructor).

The reason is the objc api resolver does a lot of work during creation.

gum/backend-darwin/gumdarwinsymbolicator.c
```diff
@@ -457,6 +485,123 @@ gum_darwin_symbolicator_find_functions_matching (GumDarwinSymbolicator * self,
...
+    function.address = GUM_ADDRESS (offset + op->linkedit);
+    prev_function = &function;
+
+    g_array_append_val (op->functions, function);
+  }
+
+  /* The very last function won't have a delta, and so no size, which
+   * means we won't be able to detect mid-function pointers there.
+   */
+  prev_function->size = 0;
```
mrmacete, line 588: we could assume the size is the rest of the `__text` section in that case, information which could be collected ahead of time alongside the function starts

gum/backend-darwin/gumdarwinsymbolicator.c
```diff
@@ -314,7 +340,8 @@ gum_darwin_symbolicator_details_from_address (GumDarwinSymbolicator * self,
   symbol = CSSymbolicatorGetSymbolWithAddressAtTime (self->handle, address,
       kCSNow);
   if (CSIsNull (symbol))
-    return FALSE;
+    return gum_darwin_symbolicator_objc_details_from_address (self,
```
mrmacete, line 343: i think this behaviour should be exposed in some way as a configuration option. 

The reason is that the objc api resolver (used under the hood by this feature) can't be used safely if the app is still paused in early instrumentation, and making this fallback by default may introduce stability issues to the existing scripts which use the `DebugSymbol` API during early instrumentation.

gum/backend-darwin/gumdarwinsymbolicator.c
```diff
@@ -326,7 +353,8 @@ gum_darwin_symbolicator_details_from_address (GumDarwinSymbolicator * self,
   {
     g_strlcpy (details->symbol_name, name, sizeof (details->symbol_name));
   }
-  else
+  else if (!gum_darwin_symbolicator_objc_details_from_address (self, address,
```
mrmacete, line 356: same as above (configuration option to enable this fallback)

mrmacete (changes requested):
nice 🤘 ! 

here are some comments related only to the "substance", not the form for now

gum/gumdarwinmodule.c
```diff
@@ -424,8 +424,7 @@ gum_darwin_module_initable_init (GInitable * initable,
     }
   }
 
-  if (self->name == NULL)
-    return gum_darwin_module_ensure_image_loaded (self, error);
+  return gum_darwin_module_ensure_image_loaded (self, error);
```
mrmacete, line 427: not sure of the impact of this change. would it be possible to just make it lazily  when you need it?

gum/backend-darwin/gumdarwinsymbolicator.c
hot3eed: Makes sense, how do you suggest we pass in or read the config data though?

gum/gumdarwinmodule.c
hot3eed: By maybe exposing it as a private API you mean?

mrmacete: yeah that's an option

gum/backend-darwin/gumdarwinsymbolicator.c
mrmacete: Probably with a settings object, i think @oleavr could point us to good examples

hot3eed: Actually, wouldn't resolver be `NULL` in case that `dlopen()` fails for `libobjc.A.dylib`? So we just need to return `FALSE` in that case?

gum/backend-darwin/gumdarwinsymbolicator.c
hot3eed: This is done in b5a1381618d3617376bf4ba9d6b0ee003569613a.

gum/backend-darwin/gumdarwinsymbolicator.c
mrmacete: the real issue isn't about `dlopen()` failing, it's described better here: https://github.com/frida/frida/issues/1167#issuecomment-707284109

oleavr: I wonder if we could somehow avoid these side-effects by detecting that a class isn't initialized, making the ObjC ApiResolver safe in all scenarios. Adding a configuration option would work, but I'm hopeful there's a way we can avoid it.

hot3eed: After discussing with @oleavr, we'll leave this as-is for now since we can't reproduce it at all.

oleavr:
Wow, nicely done! 🤘 Sorry it took me so long to get to this. Since it's been so long I figured I'd go ahead and make the final tweaks. With those out of the way, let's get this landed! 🙌

hot3eed:
@oleavr No worries! Thank you for the review 🙌

PR #553, opened 2021-05-27
Provide access to the underlying JavaScript Engine to users of the fr…

WorksButNotTested:
…ida-gumjs DevKit

The frida-gumjs allows the user to embed the JavaScript engine with its supported FRIDA APIs into their application. However, it doesn't allow sufficient access for the user to add their own functionality to the runtime. There are two major issues:
1. There is no access to the Quick JSContext / v8 Isolate which is required to add items to the global namespace
2. The type information for FRIDAs JavaScript types is not exposed. Therefore access to the argument parsing API has also been added so that the user can provide a method which takes a NativePointer for example. In to add an API which can interoperate with that of FRIDA, it must be able to interpret FRIDA's types.


WorksButNotTested:
This can be achieved by adding to the global object in JavaScript. This change is therefore withdrawn.

PR #561, opened 2021-06-24
Add minimal context to NativeCallback

mrmacete:
With this change, the `NativeCallback` js function is passed an instance of `CallbackContext` as the `this` argument.

This is an object with the `context` and `returnAddress` properties which, similarly to their counterparts found in `InvocationContext` for the `Interceptor` case, provide respectively a (minimal) CPU context and the address of the caller code.

The CPU context is minimal, in that it contains only what's needed to generate accurate backtraces, and it's read-only.

The primary use case for this addition is to provide a means for generating accurate meaningful backtraces from the context of swizzled ObjC methods (by using the `ObjC.implement()` api).

For this reason, this change also extends the Darwin (and Windows) accurate backtracer to take this case into account and properly skip the libffi stack frames. Tests are added to make sure backtraces produced this way are exaclty the same as those produced when using the Interceptor.

Since `CallbackContext` is using the backtrace to derive the `returnAddress` value, the backtracer interface has also been extended to allow for an extra `limit` parameter which can be used to shorten the depth of the backtrace below the maximum allowed in cases like this when not all the result is needed.

This is exposed through a new `gum_backtracer_generate_with_limit ()` api, while the existing `gum_backtracer_generate ()` still works as before, by implicitly using `GUM_MAX_BACKTRACE_DEPTH` as the value for `limit` under the hood.

### DIFF
--- a/bindings/gumjs/gumquickcore.c
+++ b/bindings/gumjs/gumquickcore.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2020 Ole André Vadla Ravnås <oleavr@nowsecure.com>
- * Copyright (C) 2020 Francesco Tamagni <mrmacete@protonmail.ch>
+ * Copyright (C) 2020-2021 Francesco Tamagni <mrmacete@protonmail.ch>
  * Copyright (C) 2020 Marcus Mengs <mame8282@googlemail.com>
  *
  * Licence: wxWindows Library Licence, Version 3.1
@@ -18,11 +18,11 @@
 #include "gumsourcemap.h"
 
 #include <ffi.h>
+#ifdef _MSC_VER
+# include <intrin.h>
+#endif
 #ifdef HAVE_PTRAUTH
 # include <ptrauth.h>
-#endif
-#if defined(_MSC_VER)
-# include <intrin.h>
 #endif
 
 #define GUM_QUICK_FFI_FUNCTION_PARAMS_EMPTY { NULL, }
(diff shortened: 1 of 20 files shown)


bindings/gumjs/gumquickcore.c
```diff
@@ -3622,9 +3660,31 @@ gum_quick_native_callback_invoke (ffi_cif * cif,
...
+  asm ("mov %0, lr" : "=r" (return_address));
+  asm ("mov %0, sp" : "=r" (stack_pointer));
+#elif defined(HAVE_I386)
+# if GLIB_SIZEOF_VOID_P == 8
+  asm ("movq %%rbp, %0" : "=m" (frame_pointer));
+  asm ("movq %%rsp, %0" : "=m" (stack_pointer));
+# else
+  asm ("mov %%ebp, %0" : "=m" (frame_pointer));
+  asm ("mov %%esp, %0" : "=m" (stack_pointer));
+# endif
```
oleavr, line 3686: This needs a special-case for when `_MSC_VER` is defined (Windows). That might be a painful context-switch (unless you happen to have a Windows dev env standing by), and if so I can take a look at it once I've cleared the other items on the Frida 15 backlog. Would be great to get this into the upcoming release! 🔥

oleavr (changes requested):
Wow, awesome! 💥👏

bindings/gumjs/gumquickcore.c
mrmacete: i have no windows here 😞  , so if you could take a look that would be great 🤘 !

mrmacete: This should be fixed in last commit

oleavr:
Thanks! 💥 Landed with some minor tweaks:
- Ordering of architectures should be x86/x86_64/arm/arm64(/mips) for consistency.
- Use `g_clear_object()` instead of `g_clear_pointer()` where applicable.
- `#elif defined(x)` should be `#elif defined (x)`
- Plug two V8 runtime leaks (two persistent handles were `new`ed but not `delete`d).
- Move the arm64 return address masking behind `#ifdef HAVE_DARWIN` as I don't know if we can assume that the Linux kernel sets up virtual memory the same way.

mrmacete:
thanks for the tweaks! ❤️

PR #566, opened 2021-07-09
Added callbacks for stats information

WorksButNotTested:
When fuzzing the stats information is lost each time a new child is forked. Since the parent doesn't run the target code (only prefetches it) it will not generate any transitions for the code under test.

This patch simply adds a handful of weak methods which are called each time stalker handles a transition.


WorksButNotTested:
This has now been converted to a proper glib interface.

PR #576, opened 2021-09-22
Fix x86 relocator to just patch offset for RIP relative instructions …

WorksButNotTested:
…where possible

### DIFF
--- a/gum/arch-x86/gumx86relocator.c
+++ b/gum/arch-x86/gumx86relocator.c
@@ -591,6 +591,8 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
     X86_REG_RSI, X86_REG_RDI
   };
   gint rip_reg_index, i;
+  GumAddress  address;
+  gssize offset;
   GumCpuReg other_reg, rip_reg;
   GumAbiType target_abi = self->output->target_abi;
   guint8 code[16];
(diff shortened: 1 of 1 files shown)


gum/arch-x86/gumx86relocator.c
```diff
@@ -591,6 +591,8 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
     X86_REG_RSI, X86_REG_RDI
   };
   gint rip_reg_index, i;
+  GumAddress  address;
```
oleavr, line 594:
```suggestion
  GumAddress address;
```

gum/arch-x86/gumx86relocator.c
```diff
@@ -591,6 +591,8 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
...
+  gssize offset;
```
oleavr, line 595: These two should be right after `is_rip_relative`, as per [Local variables are listed in chronological order of usage](https://github.com/frida/frida/blob/643acca36d13eb5f33ab687d62ac9fb116722608/CONTRIBUTING.md#rule-21).

gum/arch-x86/gumx86relocator.c
```diff
@@ -606,6 +608,17 @@ gum_x86_relocator_rewrite_if_rip_relative (GumX86Relocator * self,
   if (!is_rip_relative)
     return FALSE;
 
+  address = GUM_ADDRESS(insn->address + insn->size + x86->disp);
+  offset = address - (GUM_ADDRESS(cw->code) + insn->size);
+
+  if (offset >= G_MININT32 && offset <= G_MAXINT32)
+  {
+    gum_memcpy (code, ctx->start, ctx->len);
+    *((gint *) (&code[ctx->len - sizeof (gint)])) = (gint) offset;
```
oleavr, line 617:
```suggestion
    *((gint32 *) &code[ctx->len - sizeof (gint32)]) = (gint32) offset;
```

oleavr (changes requested):
Thanks! 💥

gum/arch-x86/gumx86relocator.c
WorksButNotTested: Yeah. I remembered that rule. But I put them immediately before `other_reg` since it was the next local used. However, it seems that one might be out of order in relation to `cpu_regs`, `cs_regs` and `rip_reg_index`?

oleavr: Ah yes, seems this function has degraded a bit over the years — I probably did a bad job with a previous review. :/

WorksButNotTested: No worries. No code is perfect. Especially not mine!

WorksButNotTested:
Awesome thanks for the merge.

PR #578, opened 2021-09-30
gumjs: ignore Interceptor context in unrelated NativeCallback invokes

mrmacete:
In this way invalid Interceptor contexts from higher up the call stack will be safely ignored in favor of the minimal but correct callback context.

NativeCallback instances invoked as part of `Interceptor.replace()` will continue to behave as usual and get the full context from the Interceptor's thread storage.

The new caveat is that when using the same NativeCallback instance both for `Interceptor.replace()` and an Interceptor-unrelated scenario (for example `ObjC.implement()`), AND both scenarios are invoked as part of the same call stack so that the Interceptor case happens earlier, the Interceptor context will "win" in the second invocation too and there will contain possibly invalid values. To overcome this issue, it is now possible to just use two different NativeCallback instances (which can share the same javascript function body, if needed).

### DIFF
--- a/tests/gumjs/script.c
+++ b/tests/gumjs/script.c
@@ -1931,16 +1931,25 @@ TESTCASE (native_callback_should_get_accurate_backtraces_2)
     "    }"
     "  });"
 
+    "  const interceptHere = detector['- matchesInString:options:range:'];"
+    "  Interceptor.attach(interceptHere.implementation, {"
+    "    onEnter() { }"
+    "  });"
+
     "  detector.matchesInString_options_range_(testString, 0, range);"
 
     "  const origImpl = method.implementation;"
     "  method.implementation = ObjC.implement(method,"
     "    function (handle, selector, ...args) {"
     "      if (sample === null) {"
-    "        sample = Thread.backtrace(this.context, Backtracer.ACCURATE);"
-    "        sampleRet = this.returnAddress;"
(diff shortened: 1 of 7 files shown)


tests/gumjs/script.c
```diff
@@ -1931,16 +1931,25 @@ TESTCASE (native_callback_should_get_accurate_backtraces_2)
     "    }"
     "  });"
 
+    "  const interceptHere = detector['- matchesInString:options:range:'];"
+    "  Interceptor.attach(interceptHere.implementation, {"
+    "    onEnter() { }"
```
oleavr, line 1936:
```suggestion
    "    onEnter() {}"
```

tests/gumjs/script.c
```diff
@@ -5898,7 +5907,8 @@ TESTCASE (replaced_function_should_have_invocation_context)
 {
   COMPILE_AND_LOAD_SCRIPT (
       "Interceptor.replace(" GUM_PTR_CONST ", new NativeCallback(function () {"
-      "  send(this.returnAddress instanceof NativePointer);"
+      "  send(this.returnAddress instanceof NativePointer &&"
+      "     !this.context.pc.isNull());"
```
oleavr, line 5911:
```suggestion
      "      !this.context.pc.isNull());"
```

oleavr (approved):

PR #586, opened 2021-10-20
gumjs: Bump frida-swift-bridge

hot3eed:


oleavr:
Thanks! Unfortunately I totally missed this before making the same change, sorry for the duplication of effort 🤦

PR #591, opened 2021-11-04
Add regex support to `Memory.scan()` and implement `MatchPattern` object

hot3eed:
In this PR:
* Allow `Memory.scan()` to scan for regex patterns, which is triggered automatically when the `pattern` argument is prefixed and suffixed with `/`, e.g. `/Some\s*Pattern/`. 
* Implement the `MatchPattern` object, whose constructor accepts either: a hexstring, `Memory.scan()`-style; or a regex pattern with the format above. This allows the user of the API to re-use a compiled regex, by keeping a reference to the `MatchPattern` object. 
* Make it return a `Promise<void>` instead of `void`.
* Make the `onComplete` callback optional (since we have the settling of the Promise to indicate that we're finished scanning.) 
* The API remains compatible with the older, callback-style one (except for one edge case, see below.)

Points for discussion:
* Is returning `Promise<void>` instead of `void` considered API-breaking? There's one edge case I can think of where code wouldn't take the expected control flow. If a call to `Memory.scan()` throws an exception before starting the scan, due to bad or missing arguments for example, AND the user of the API is catching this exception, then since we now return a Promise, there will be an unhandled rejection instead of a caught exception. Example:

```javascript
try {
  /* should be awaited so that the rejection is thrown as an exception */
  Memory.scan(ptr, size, pattern, {
    onMatch (address, size) {
      ...
    },
    onError (reason) { 
      /* will execute! */
    }
  });
} catch (e) {
  /* will never execute, happens only before initiating the scan */
} 
```
* To keep `Memory.scanSync()` compatible with `Memory.scan()`, we should make it optionally accept a `MatchPattern` for a `pattern` as well. Does this warrant adding a qualifier for `MatchPattern | string` in `xxx_args_parse()` (for V8 and QJS), or should we add a helper method to do that?
* When should we free [this stuff](https://github.com/frida/frida-gum/pull/591/files#diff-66a4ee2d704657707e96bdde6e84adca21eb81ce98e4ad21fedfb70feca139eeR991)?

### DIFF
--- a/bindings/gumjs/gumquickmemory.c
+++ b/bindings/gumjs/gumquickmemory.c
@@ -947,7 +947,8 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
 
   pattern_val = args->elements[2];
 
-  if (JS_IsString (pattern_val)) {
+  if (JS_IsString (pattern_val))
+  {
     const gchar * match_str;
 
     if (!_gum_quick_string_get (ctx, args->elements[2], &match_str))
@@ -955,9 +956,13 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
 
     sc.pattern = gum_match_pattern_new_from_string (match_str);
     JS_FreeCString (ctx, match_str);
-  } else if (JS_IsObject (pattern_val)) {
+  }
+  else if (JS_IsObject (pattern_val))
+  {
     sc.pattern = JS_GetOpaque(pattern_val, core->match_pattern_class);
-  } else {
+  }
+  else
+  {
     return _gum_quick_throw_literal (ctx,
         "expected either a pattern string or a MatchPattern object");
   }
@@ -967,7 +972,8 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
 
   callbacks = args->elements[3];
 
-  if (!JS_IsObject (callbacks)) {
+  if (!JS_IsObject (callbacks))
+  {
     return _gum_quick_throw_literal (ctx,
         "expected an object containing callbacks");
   }
@@ -975,9 +981,8 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
   sc.on_match = JS_GetPropertyStr (ctx, callbacks, "onMatch");
   sc.on_complete = JS_GetPropertyStr (ctx, callbacks, "onComplete");
 
-  if (!JS_IsFunction (ctx, sc.on_match)) {
+  if (!JS_IsFunction (ctx, sc.on_match))
     return _gum_quick_throw_literal (ctx, "expected a callback value");
-  }
 
   sc.on_complete = JS_GetPropertyStr (ctx, callbacks, "onComplete");
   sc.on_error = JS_GetPropertyStr (ctx, callbacks, "onError");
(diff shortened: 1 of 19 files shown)


bindings/gumjs/gumquickmemory.c
```diff
@@ -931,21 +932,63 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
...
+    return JS_EXCEPTION;
 
-  if (!_gum_quick_args_parse (args, "pZsF{onMatch,onError?,onComplete}",
-      &address, &size, &match_str, &sc.on_match, &sc.on_error, &sc.on_complete))
+  if (!_gum_quick_size_get (ctx, args->elements[1], core, &size))
     return JS_EXCEPTION;
+
+  pattern_val = args->elements[2];
+
+  if (JS_IsString (pattern_val)) {
```
oleavr, line 950:
```suggestion
  if (JS_IsString (pattern_val))
  {
```

(Here and elsewhere.)

bindings/gumjs/gumquickmemory.c
```diff
@@ -931,21 +932,63 @@ GUMJS_DEFINE_FUNCTION (gumjs_memory_scan)
...
+    return _gum_quick_throw_literal (ctx,
+        "expected an object containing callbacks");
+  }
+
+  sc.on_match = JS_GetPropertyStr (ctx, callbacks, "onMatch");
+  sc.on_complete = JS_GetPropertyStr (ctx, callbacks, "onComplete");
+
+  if (!JS_IsFunction (ctx, sc.on_match)) {
+    return _gum_quick_throw_literal (ctx, "expected a callback value");
+  }
```
oleavr, line 980: Braces should be omitted for single-line statements. (Unless the line is really long/code is hard to read.)

oleavr (changes requested):
> Is returning Promise<void> instead of void considered API-breaking?

No. But in case of a bad argument we should throw right away, without returning a `Promise`.

> Does this warrant adding a qualifier for MatchPattern | string in xxx_args_parse() (for V8 and QJS), or should we add a helper method to do that?

Let's add a helper method as it's only in two places, and they're part of the same API. (It makes sense to add format specifiers for types that are shared by multiple APIs.)

> When should we free this stuff?

Things should be freed in `gum_memory_scan_context_free()`.

hot3eed:
Sounds good. I've just pushed the changes you requested.

tests/gumjs/script.c
```diff
@@ -6174,6 +6263,36 @@ TESTCASE (memory_can_be_scanned_synchronously)
...
+{
+  guint8 haystack[] = { 0x01, 0x02, 0x13, 0x37, 0x03, 0x13, 0x37 };
+
+  COMPILE_AND_LOAD_SCRIPT (
+      "Promise.resolve(Memory.scan(" GUM_PTR_CONST ", 7, '13 37', {"
+      "  onMatch(address, size) {"
+      "    send('onMatch offset=' + address.sub(" GUM_PTR_CONST ").toInt32()"
+      "      + ' size=' + size);"
+      "  }"
+      "})).catch(e => console.error(e.message)).then(() => send('DONE'));", haystack, haystack);
```
oleavr, line 6276: This exceeds 80 cols. (Would be good to also verify that no other lines exceed 80 cols.)

oleavr (changes requested):
Thanks! Getting there:

tests/gumjs/script.c
hot3eed: Oops, sorry for the rookie mistake! Done.

PR #592, opened 2021-11-05
gumjs: Fix v8 Interceptor context

3vilWind:
There was a regression in the tests starting from #578
Fixed getting wrong variable (v8), explicitly initializing new variables (qjs, v8)

### DIFF
--- a/bindings/gumjs/gumquickcore.c
+++ b/bindings/gumjs/gumquickcore.c
@@ -3537,6 +3537,7 @@ GUMJS_DEFINE_CONSTRUCTOR (gumjs_native_callback_construct)
   cb->wrapper = wrapper;
   cb->func = func;
   cb->core = core;
+  cb->interceptor_replacement_count = 0;
 
   if (!gum_quick_ffi_type_get (ctx, rtype_value, core, &rtype, &cb->data))
     goto propagate_exception;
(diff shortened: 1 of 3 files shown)


bindings/gumjs/gumquickcore.c
```diff
@@ -3537,6 +3537,7 @@ GUMJS_DEFINE_CONSTRUCTOR (gumjs_native_callback_construct)
   cb->wrapper = wrapper;
   cb->func = func;
   cb->core = core;
+  cb->interceptor_replacement_count = 0;
```
oleavr, line 3540: Since we're using `g_slice_new0()` here, all fields will be zero-initialized and we can omit this.

bindings/gumjs/gumv8core.cpp
```diff
@@ -2906,6 +2906,7 @@ GUMJS_DEFINE_CONSTRUCTOR (gumjs_native_callback_construct)
 
   callback = g_slice_new0 (GumV8NativeCallback);
   callback->ref_count = 1;
+  callback->interceptor_replacement_count = 0;
```
oleavr, line 2909: This can also be omitted because of `g_slice_new0()`.

oleavr (changes requested):
Thanks a lot, great catch! 💥  Just a couple of nits:

oleavr:
Actually let me apply these minor tweaks now so we can land this — definitely want this in the upcoming release 👍

PR #593, opened 2021-11-05
gumjs: Expose more capstone api

3vilWind:
- access field in operands
- regsAccess.read/written (through cs_regs_access) - get read/written registers both explicitly and implicitly, unlike existing regsRead/regsWritten


oleavr:
Yay, this is awesome. Thanks a lot! 🙌

PR #596, opened 2021-11-08
kernel: Get base from all_image_info where supported

mrmacete:
Jailbreaks like Unc0ver and checkra1n store the kernel slide in `all_image_info_size`, obtainable via a call to `task_info` on the kernel task port. This change adds support for reconstructing the kernel base address from it.

### DIFF
--- a/gum/backend-darwin/gumkernel-darwin.c
+++ b/gum/backend-darwin/gumkernel-darwin.c
@@ -705,12 +705,12 @@ gum_kernel_do_find_base_address (void)
   if (base == 0)
   {
     version = gum_kernel_get_version ();
-    if (version >= 16.0) // iOS 10.0+
+    if (version >= 16.0) /* iOS 10.0+ */
     {
       base = gum_kernel_bruteforce_base (
           G_GUINT64_CONSTANT (0xfffffff007004000));
     }
-    else if (version >= 15.0) // iOS 9.0+
+    else if (version >= 15.0) /* iOS 9.0+ */
     {
       base = gum_kernel_bruteforce_base (
           G_GUINT64_CONSTANT (0xffffff8004004000));
--- a/gum/backend-darwin/gumkernel-darwin.c
+++ b/gum/backend-darwin/gumkernel-darwin.c
@@ -9,6 +9,7 @@
 #include "gum-init.h"
 #include "gumdarwin.h"
 #include "gummemory-priv.h"
+#include "gumprocess-darwin-priv.h"
 
 #include <mach/mach.h>
 #include <mach-o/loader.h>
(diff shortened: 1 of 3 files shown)


gum/backend-darwin/gumkernel-darwin.c
```diff
@@ -698,13 +700,22 @@ gum_kernel_do_find_base_address (void)
...
 #ifdef HAVE_ARM64
-  if (version >= 16.0) /* iOS 10.0+ */
-    base = gum_kernel_bruteforce_base (G_GUINT64_CONSTANT (0xfffffff007004000));
-  else if (version >= 15.0) /* iOS 9.0+ */
-    base = gum_kernel_bruteforce_base (G_GUINT64_CONSTANT (0xffffff8004004000));
+  base = gum_kernel_get_base_from_all_image_info ();
+  if (base == 0)
+  {
+    version = gum_kernel_get_version ();
+    if (version >= 16.0) // iOS 10.0+
```
oleavr, line 708:
```suggestion
    if (version >= 16.0) /* iOS 10.0+ */
```

Should stick to C-style comments for consistency.

gum/backend-darwin/gumkernel-darwin.c
```diff
@@ -698,13 +700,22 @@ gum_kernel_do_find_base_address (void)
...
+    {
+      base = gum_kernel_bruteforce_base (
+          G_GUINT64_CONSTANT (0xfffffff007004000));
+    }
+    else if (version >= 15.0) // iOS 9.0+
```
oleavr, line 713:
```suggestion
    else if (version >= 15.0) /* iOS 9.0+ */
```

(Here as well.)

gum/backend-darwin/gumprocess-darwin-priv.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __GUM_PROCESS_DARWIN_PRIV_H__
+#define __GUM_PROCESS_DARWIN_PRIV_H__
```
oleavr, line 2: This header isn't self-contained, as it uses GLib types without including needed header(s). Would also be good to add the copyright header. (Though I'm thinking about dropping them at some point — but for now let's add it for consistency.)

gum/backend-darwin/gumprocess-darwin.c
```diff
@@ -6,6 +6,7 @@
  */
 
 #include "gumprocess-priv.h"
+#include "gumprocess-darwin-priv.h"
```
oleavr, line 9: This should be moved to the next group, as the first group should only have one include, which should be the main header for this C source file. (The next group contains internal includes, so this include should go there.)

oleavr (changes requested):
Yay, nice! 🤘

PR #598, opened 2021-11-09
x86-relocator: fix Jccs

0xDC00:
Switch to jcc_short when jcc_near failed (-+2GB).


oleavr (changes requested):
Thanks! Looks good. But this needs test-coverage before we can merge it.

oleavr:
Actually let's not let this block on test coverage, we'll add that later.

PR #600, opened 2021-11-16
Fixing the case of SymWalk being sensitive about XIP.

HonicRoku:
Steps to reproduce the problem that this patch fixes:

1) Run notepad in Frida (x64).
2) Attach hook on combase!CoCreateInstance that prints backtrace onEnter (also load symbols for all modules..).

See truncated or invalid backtraces.

This patch also fixes the case when there are multiple hooks in a cascade, e.g. CreateFileInternal & CreateFileW. See the StackWalk loop.

The problem is that the code tries to skip the first frame by using the return address from the top of stack. SymWalk does not like this -- there is mismatch between XSP and XIP. StackWalk is sensitive about RIP, especially on x64.

It was tested with both 32-bit & 64-bit builds on Windows 10.

### DIFF
--- a/gum/backend-dbghelp/gumdbghelpbacktracer.c
+++ b/gum/backend-dbghelp/gumdbghelpbacktracer.c
@@ -94,6 +94,8 @@ gum_dbghelp_backtracer_generate (GumBacktracer * backtracer,
   /* Get the raw addresses */
   RtlCaptureContext (&context);
 
+  
+
   frame.AddrPC.Mode = AddrModeFlat;
   frame.AddrFrame.Mode = AddrModeFlat;
   frame.AddrStack.Mode = AddrModeFlat;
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! I've just applied some style tweaks and simplifications — could you check if this still works for you?

htgihbj:
Yes, it works. I've also reviewed the code -- looks good to me. Sorry about the style, won't happen again. :) Thank you.

HonicRoku:
Yes, it works (32-bit/64-bit Frida on Windows 10 64-bit). I've also reviewed the code -- looks good to me. Sorry about the style, won't happen again. Thank you!

oleavr:
Yay, thanks for testing! 🙌

oleavr:
(No worries regarding the style tweaks btw! 😊)

PR #601, opened 2021-11-19
Implement `MatchPattern` qualifier and update `Kernel.scan*()`

hot3eed:
This makes the `Kernel.scan*()` API fully compatible with `Memory.scan*()`, which was [just updated](https://github.com/frida/frida-gum/pull/591) to add regex support.

TODO:

- [x] Return a `Promise` from `Kernel.scan()`.
- [x] Add tests

### DIFF
--- a/bindings/gumjs/gumquickvalue.c
+++ b/bindings/gumjs/gumquickvalue.c
@@ -537,6 +537,42 @@ _gum_quick_args_parse (GumQuickArgs * self,
 
         break;
       }
+      case 'M':
+      {
+        GumMatchPattern * pattern;
+
+        if (JS_IsString (arg))
+        {
+          const char * str;
+
+          str = JS_ToCString (ctx, arg);
+          if (str == NULL)
+            goto propagate_exception;
+
+          pattern = gum_match_pattern_new_from_string (str);
+          JS_FreeCString (ctx, str);
+
+          if (pattern == NULL)
+            goto invalid_pattern;
+
+        }
+        else if (JS_IsObject (arg))
+        {
+          pattern = JS_GetOpaque (arg, core->match_pattern_class);
+          if (pattern == NULL)
+            goto expected_pattern;
+        }
+        else
+        {
+          goto expected_pattern;
+        }
+
+        gum_match_pattern_ref (pattern);
+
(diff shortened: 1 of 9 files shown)


bindings/gumjs/gumquickvalue.c
```diff
@@ -537,6 +538,42 @@ _gum_quick_args_parse (GumQuickArgs * self,
...
+          str = JS_ToCString (ctx, arg);
+          if (str == NULL)
+            goto propagate_exception;
+
+          pattern = gum_match_pattern_new_from_string (str);
+          JS_FreeCString (ctx, str);
+
+          if (pattern == NULL)
+            goto invalid_pattern;
+
```
oleavr, line 558: Stray newline.

bindings/gumjs/gumquickvalue.c
```diff
@@ -537,6 +538,42 @@ _gum_quick_args_parse (GumQuickArgs * self,
...
+        if (JS_IsString (arg))
+        {
+          const char * str;
+
+          str = JS_ToCString (ctx, arg);
+          if (str == NULL)
+            goto propagate_exception;
+
+          pattern = gum_match_pattern_new_from_string (str);
+          JS_FreeCString (ctx, str);
```
oleavr, line 554:
```suggestion

          JS_FreeCString (ctx, str);
```

To make it feel more "balanced" visually.

bindings/gumjs/gumquickvalue.c
```diff
@@ -537,6 +538,42 @@ _gum_quick_args_parse (GumQuickArgs * self,
...
+          pattern = JS_GetOpaque (arg, core->match_pattern_class);
+          if (pattern == NULL)
+            goto expected_pattern;
+        }
+        else
+        {
+          goto expected_pattern;
+        }
+
+        gum_match_pattern_ref (pattern);
```
oleavr, line 571: This will create an extra reference for the case where it's constructed from a string. So this `ref()` should only happen for the case when it's constructed from an existing `MatchPattern`.

oleavr (changes requested):

oleavr:
Thanks, this is awesome! 🥳

PR #607, opened 2021-11-29
Excluded3

WorksButNotTested:


oleavr (changes requested):
Thanks, this is awesome! 💥  I've landed the op-code fix, but the other two commits need some work.

Exception handling:
- I've taken care of the style fixes in the implementation, but the tests are still full of style issues.
- Relying on `__gxx_personality_v0` and `_Unwind_Find_FDE` is definitely going to create a regression on Android. Using `gum_module_find_export_by_name (NULL, name)` to look these up dynamically might be enough to solve this.
- The excluded calls lookup table should be a `GumMetalHashTable *` since it will be modified from sensitive code-paths where we cannot use our heap. (We do the same for our `mappings` to avoid this issue.)
- Nothing protects the excluded calls lookup table, so stalking multiple threads is going to result in undefined behaviour.
- gum_stalker_exception_find_fde() has a `+ 1` that looks suspicious, but I'm probably missing something.

PLT:
- From a quick glance this needs a lot of style fixes, most obvious being missing space between function name and opening parentheses. There are also a couple of blank lines that look out of place.

WorksButNotTested:
Hopefully this should take care of things. I have:

- Fixed the style problems I can see. Do let me know if there are any more I have missed, I think it was mostly missing spaces before braces.
- Removed the dependency on the `__gxx_personality_v0` and `_Unwind_Find_FDE` symbols and instead used `gum_find_function` (since if things are statically linked then these functions won't be exports). The functions are only hooked if they are found (which should mean we at least don't break things if we can't find them).
- Switched to use a `GumMetalHashTable` as suggested.
- Moved the `excluded_calls` hashtable to be a member of the `GumExecCtx`. Memory consumption should be minimal and we avoid any thread synchronization issues this way. It also keeps the code clean without a load of lock/unlock logic.
- Commented the functions we are hooking and explained the mysterious `+1`.
- Fixed the style issues in the PLT changes I could find.

s1341:
Amazing work @WorksButNotTested !!! Glad you got to this before I did, as the implementation I had in mind was much more complex and probably fragile.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  void
+  test_set_bit (const char * func,
+                guint32 * val,
+                guint8 bit)
+  {
+    if (g_test_verbose ())
+      g_print ("\tFunc: %s, Set: %d\n", func, bit);
+
+    *val = (*val | ((guint32) 1 << bit));
```
oleavr, line 43:
```suggestion
    *val |= (guint32) 1 << bit;
```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+  {
+    test_try_and_catch_pp (val);
+  }
+
+  void
+  test_try_and_dont_catch (guint32 * val)
+  {
+    test_try_and_dont_catch_pp (val);
+  }
+
```
oleavr, line 57:
```suggestion
```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+class TestException {
```
oleavr, line 60:
```suggestion
class TestException
{
```

Here and elsewhere.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+
```
oleavr, line 61:
```suggestion
```

Here and elsewhere.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+  private:
+    guint32 * val;
+  public:
+    TestResource (guint32 * val) : val (val)
+    {
+      test_set_bit ("TestResource", val, 0);
+    }
+    ~TestResource () {
+      test_set_bit ("TestResource", val, 1);
+    }
```
oleavr, line 78: Indentation is incorrect. See bindings/gumjs/gumv8platform.cpp for some examples. Basically:
- Public goes first, followed by private.
- Visibility label should not be indented.
- Braces always go on the next line.
- Member initializer list should be indented like this:

```
  GumV8PlatformLocker (GumV8Platform * platform)
    : locker (&platform->mutex)
  {
  }
```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+};
+
+static void
+test_try_and_catch_pp (guint32 * val)
+{
+  try {
```
oleavr, line 84:
```suggestion
  try
  {
```

Here and elsewhere.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+
```
oleavr, line 85:
```suggestion
```

Here and elsewhere. Blank lines should never be at the start or end of a block.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+    test_set_bit ("test_try_and_catch_pp", val, 0);
+
+    throw TestException ();
+
+    test_set_bit ("test_try_and_catch_pp", val, 1);
+  } catch (TestException &ex) {
```
oleavr, line 91:
```suggestion
  }
  catch (TestException &ex)
  {
```

Here and elsewhere.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+static void
+test_try_and_dont_catch_pp (guint32 * val)
+{
+    try
+    {
+      test_set_bit ("test_try_and_dont_catch_pp", val, 2);
+
+      test_try_and_dont_catch_pp_2(val);
```
oleavr, line 105:
```suggestion
      test_try_and_dont_catch_pp_2 (val);
```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+    test_set_bit ("test_try_and_catch_pp", val, 2);
+  }
+
+  test_set_bit ("test_try_and_catch_pp", val, 3);
+}
+
+static void
+test_try_and_dont_catch_pp (guint32 * val)
+{
+    try
```
oleavr, line 101: This function should not have 4 space indentation.

tests/core/arch-x86/stalker-x86-exceptions.cpp
oleavr, line 91:
```suggestion
  } catch (TestException & ex) {
```

Space after ampersand, just like asterisk. Here and elsewhere.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+      test_set_bit ("test_try_and_dont_catch_pp", val, 5);
+    }
+
+    test_set_bit ("test_try_and_dont_catch_pp", val, 6);
+}
+
+static void
+test_try_and_dont_catch_pp_2 (guint32 * val)
+{
+
```
oleavr, line 124: Blank line should go after the `TestResource` variable below it. Blocks should never start or end with a blank line.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+{
+  test_set_bit ("test_try_and_dont_catch_pp_3", val, 11);
+  throw TestException ();
+  test_set_bit ("test_try_and_dont_catch_pp_3", val, 12);
+}
+
+
+
+
+
```
oleavr, line 153: Lots of trailing blank lines here.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,153 @@
...
+      test_try_and_dont_catch_pp_3 (val);
+
+      test_set_bit ("test_try_and_dont_catch_pp_2", val, 8);
+    }
+    catch (FakeException&)
+    {
+      test_set_bit ("test_try_and_dont_catch_pp_2", val, 9);
+    }
+
+    test_set_bit ("test_try_and_dont_catch_pp_2", val, 10);
```
oleavr, line 139: This function should not have 4 space indentation.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -105,6 +105,16 @@ TESTLIST_BEGIN (stalker)
 #ifndef HAVE_WINDOWS
   TESTENTRY (ic_var)
 #endif
+
+#ifdef HAVE_LINUX
+  TESTGROUP_BEGIN("ExceptionHandling")
```
oleavr, line 110:
```suggestion
  TESTGROUP_BEGIN ("ExceptionHandling")
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -203,6 +213,18 @@ G_DEFINE_TYPE_EXTENDED (GumTestStalkerObserver,
...
 static GHashTable * prefetch_executed = NULL;
 static PrefetchBackpatchContext bp_ctx;
+
+static void callback_at_end (GumStalkerIterator * iterator,
+    GumStalkerOutput * output, gpointer user_data);
+static void callout_at_end (GumCpuContext * cpu_context, gpointer user_data);
+static void test_check_followed (void);
+
+extern void __cxa_throw (void* thrownException, void* type,
+    void (*destructor)(void*));
```
oleavr, line 223:
```suggestion
extern void __cxa_throw (void * thrownException, void * type,
    void (* destructor)(void *));
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
 }
 
 #endif
+
+#ifdef HAVE_LINUX
```
oleavr, line 3227:
```suggestion
#ifdef HAVE_LINUX

```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+TESTCASE (no_exceptions)
+{
+  guint32 val = 0;
```
oleavr, line 3230:
```suggestion
  guint32 val = 0;

```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+      callback_at_end, &val, NULL);
+
+  gum_stalker_follow_me (fixture->stalker, fixture->transformer,
+      GUM_EVENT_SINK (fixture->sink));
+  test_check_followed ();
+  gum_stalker_unfollow_me (fixture->stalker);
+
+  g_assert_cmpuint (fixture->sink->events->len, >, 0);
+
+  if (g_test_verbose())
```
oleavr, line 3243:
```suggestion
  if (g_test_verbose ())
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  fixture->transformer = gum_stalker_transformer_make_from_callback (
+      callback_at_end, &val, NULL);
+
+  gum_stalker_follow_me (fixture->stalker, fixture->transformer,
+      GUM_EVENT_SINK (fixture->sink));
+  test_try_and_catch (&val);
+  test_check_followed ();
+  gum_stalker_unfollow_me (fixture->stalker);
+
+  if (g_test_verbose())
```
oleavr, line 3263:
```suggestion
  if (g_test_verbose ())
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
```
oleavr, line 3292:
```suggestion
  if (g_test_verbose ())
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  fixture->transformer = gum_stalker_transformer_make_from_callback (
+      callback_at_end, &val, NULL);
+
+  gum_stalker_follow_me (fixture->stalker, fixture->transformer,
+      GUM_EVENT_SINK (fixture->sink));
+  test_try_and_dont_catch (&val);
+  test_check_followed ();
+  gum_stalker_unfollow_me (fixture->stalker);
+
+  if (g_test_verbose())
```
oleavr, line 3315:
```suggestion
  if (g_test_verbose ())
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  test_check_bit (&val, 0);
+  test_check_bit (&val, 1);
+  test_check_bit (&val, 2);
+  test_check_bit (&val, 5);
+  test_check_bit (&val, 6);
+  test_check_bit (&val, 7);
+  test_check_bit (&val, 11);
+  test_check_bit (&val, 31);
+  g_assert_cmpuint (val, ==, 0);
+
```
oleavr, line 3327:
```suggestion
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  fixture->transformer = gum_stalker_transformer_make_from_callback (
+      callback_at_end, &val, NULL);
+
+  gum_stalker_follow_me (fixture->stalker, fixture->transformer,
+      GUM_EVENT_SINK (fixture->sink));
+  test_try_and_dont_catch (&val);
+  test_check_followed ();
+  gum_stalker_unfollow_me (fixture->stalker);
+
+  if (g_test_verbose())
```
oleavr, line 3349:
```suggestion
  if (g_test_verbose ())
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  test_check_bit (&val, 7);
+  test_check_bit (&val, 11);
+  test_check_bit (&val, 31);
+  g_assert_cmpuint (val, ==, 0);
+}
+
+static void
+callback_at_end (GumStalkerIterator * iterator,
+                GumStalkerOutput * output,
+                gpointer user_data)
```
oleavr, line 3366:
```suggestion
                 GumStalkerOutput * output,
                 gpointer user_data)
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+{
+  guint32 * val = (guint32 *) user_data;
```
oleavr, line 3368: Redundant cast. (Since this is C, not C++, `void *` to `any *` does not require a cast.)

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+      gum_stalker_iterator_put_callout (iterator, callout_at_end, val, NULL);
+    }
+  }
+}
+
+static void
+callout_at_end (GumCpuContext * cpu_context,
+                gpointer user_data)
+{
+  guint32 * val = (guint32 *) user_data;
```
oleavr, line 3386: Redundant cast.

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+  *val = *val + (1 << 31);
```
oleavr, line 3387:
```suggestion
  *val += 1 << 31;
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,174 @@ TESTCASE (ic_var)
...
+}
+
+GUM_NOINLINE static void
+test_check_followed (void)
+{
+  /* Avoid calls being optimized out */
+  asm ("");
+}
```
oleavr, line 3395:
```suggestion
}

```

tests/core/meson.build
```diff
@@ -68,9 +68,26 @@ if host_arch == 'arm64'
   endif
 endif
 
+gum_tests_core_link = []
```
oleavr, line 71: Let's rename this to `core_link_with` (or something like that) and move this up right after `core_sources`, with a blank line before it.

tests/core/meson.build
```diff
@@ -68,9 +68,26 @@ if host_arch == 'arm64'
...
+
+if host_os_family == 'linux'
+  if host_arch == 'x86' or host_arch == 'x86_64'
```
oleavr, line 74:
```suggestion
if host_os_family == 'linux' and host_arch in ['x86', 'x86_64']
```

(At least I think this kind of syntax is supported.)

tests/core/meson.build
```diff
@@ -68,9 +68,26 @@ if host_arch == 'arm64'
...
+
+    gum_tests_core_exceptions = static_library('gum-tests-core-exceptions',
+        ['arch-x86/stalker-x86-exceptions.cpp'],
+        include_directories: test_incdirs,
+        dependencies: [gum_dep],
+        override_options: ['cpp_eh=default'],
```
oleavr, line 80: Should be indented by two spaces, not four.

tests/core/meson.build
```diff
@@ -68,9 +68,26 @@ if host_arch == 'arm64'
...
+    )
+    gum_tests_core_link += gum_tests_core_exceptions
+
```
oleavr, line 83:
```suggestion
```

tests/meson.build
```diff
@@ -45,6 +45,14 @@ if have_gumjs
 endif
 
 force_cpp_linking = have_gumpp or have_v8
+
+# We have tests for Stalker C++ exception handling
+if host_os_family == 'linux'
+  if host_arch == 'x86' or host_arch == 'x86_64'
```
oleavr, line 51: Check should be simplified like above.

oleavr (changes requested):
Thanks! 💥 Getting there:

gum/backend-x86/gumstalker-x86.c
```diff
@@ -2030,6 +2092,46 @@ gum_exec_ctx_new (GumStalker * stalker,
...
+
+  {
+    GumInterceptor * interceptor;
+    GumAttachReturn attach_ret;
+
+    interceptor = gum_interceptor_obtain ();
+    stalker->interceptor = interceptor;
+
+    /* This function may be statically linked and therefore not an export */
+    __gxx_personality_v0_impl = gum_find_function ("__gxx_personality_v0");
```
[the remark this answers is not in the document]
WorksButNotTested: This is only called once per thread stalked. So quite infrequently? Is the DebugSymbol performance that much of an overhead it may be significant if it's on a very cold code path? Should I change it to a `dlsym`?

gum/backend-x86/gumstalker-x86.c
```diff
@@ -2080,6 +2182,25 @@ gum_exec_ctx_free (GumExecCtx * ctx)
...
+    GumInterceptor * interceptor;
+
+    interceptor = g_steal_pointer (&stalker->interceptor);
+    if (interceptor != NULL)
+    {
+      if (_Unwind_Find_FDE_impl != NULL)
+        gum_interceptor_revert (interceptor, _Unwind_Find_FDE_impl);
+
+      if (__gxx_personality_v0_impl != NULL)
+        gum_interceptor_revert (interceptor, __gxx_personality_v0_impl);
```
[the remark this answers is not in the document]
WorksButNotTested: Can you point me at a suitable example where this is done so I can copy the approach?

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
oleavr: Hmm well I suppose we could to `gum_module_find_export_by_name (NULL, name)` and fall back to `gum_find_function (name)`. The reason I'm worried about this isn't per call, but per process lifetime, as it tends to be significant when stalking a short-lived process. (E.g. on i/macOS it might take a few seconds for the first call where things are being warmed up.)

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
oleavr: I'd say `bindings/gumjs/gumscriptbackend.c`.

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
WorksButNotTested: I had a problem when I tried this. The `_Unwind_Find_FDE` function although statically linked did have an export. However, the value returned by the DebugSymbol is different. The one returned by the export is never called, I guess maybe it's just a wrapper?

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
WorksButNotTested: I've got a bit of a problem with this. The hook functions need to be able to access the GMetalHashTable containing the lookup information. The only way for these functions to retrieve this information is by gaining a handle to the `GumStalker`. Thus the hooks cannot outlive the Stalker instance itself. I shall move the instantiation accordingly.

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
oleavr: Oh, you mean our own libunwind? If so that sounds like a mistake made by the integrator of the Frida SDK, where they don't limit the exported symbols to the bare minimum.

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
oleavr: But they can retrieve the `ExecCtx` through the thread-local data though?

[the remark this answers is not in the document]
oleavr: Oh, that's currently on the `Stalker` instance. I think we should make it global, like we do for `gum_interceptor_context_private`.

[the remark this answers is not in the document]
WorksButNotTested: I’m not sure I follow? Also if the hash table isn’t attached to the context, then if we don’t purge the related entries from the hashtable, if the same memory pages are reused for another instance slab,  we can end up with collisions?

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
WorksButNotTested: I’d guess so. Unless we somehow end up with two copies of things?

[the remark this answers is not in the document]
oleavr: Well I don't think we should worry about that, we should expect the integrator (who's linking in frida-gum) to ensure they don't export any other symbols than strictly needed, e.g. like frida-server does with `-Wl,--version-script,frida-server.version` containing:

```
FRIDA_SERVER {
  local:
    *;
};
```

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
oleavr: I mean the hash table should be attached to the context, like it is now. But the thread-local key used to find the context, shouldn't be attached to the `Stalker` instance. This is because it doesn't make sense for two different `Stalker` instances to stalk the same thread at the same time.

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
WorksButNotTested: I've put in the change as suggested, but whilst x64 is fine, I am still having a problem on x86. As we can see below,
both contain an implementation of `_Unwind_Find_FDE` of their own, and both depend on `libc`. However, on `x86` whilst `libc` exports `_Unwind_Find_FDE`, on `x64` it does not. Obviously when the tests make use of `_Unwind_Find_FDE` it uses it's in-built copy. But by preferring `gum_module_find_export_by_name` over `gum_find_function` to find the function, on `x86` it has hooked the wrong one.

```bash
$ readelf -sW ../build/tmp-linux-x86/frida-gum/tests/gum-tests | grep _Unwind_Find_FDE
   929: 000642f9     5 FUNC    LOCAL  DEFAULT   15 _Unwind_Find_FDE.cold
 73952: 0101e1f0   628 FUNC    LOCAL  DEFAULT   15 _Unwind_Find_FDE
$ ldd ../build/tmp-linux-x86/frida-gum/tests/gum-tests
        linux-gate.so.1 (0xf7fd0000)
        libdl.so.2 => /lib/i386-linux-gnu/libdl.so.2 (0xf68c5000)
        libresolv.so.2 => /lib/i386-linux-gnu/libresolv.so.2 (0xf68ac000)
        libpthread.so.0 => /lib/i386-linux-gnu/libpthread.so.0 (0xf6889000)
        libm.so.6 => /lib/i386-linux-gnu/libm.so.6 (0xf6784000)
        libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf6595000)
        /lib/ld-linux.so.2 (0xf7fd1000)
$ readelf -sW /lib/i386-linux-gnu/libc.so.6 | grep _Unwind_Find_FDE
  1052: 00146cc0   186 FUNC    GLOBAL DEFAULT   16 _Unwind_Find_FDE@@GCC_3.0
```

```bash
$ readelf -sW ../build/tmp-linux-x86_64/frida-gum/tests/gum-tests | grep _Unwind_Find_FDE
   665: 0000000000116c7f     5 FUNC    LOCAL  DEFAULT   15 _Unwind_Find_FDE.cold
 55018: 00000000010a32b0   595 FUNC    LOCAL  DEFAULT   15 _Unwind_Find_FDE
$ ldd ../build/tmp-linux-x86_64/frida-gum/tests/gum-tests
        linux-vdso.so.1 (0x00007ffff7fce000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ffff67cb000)
        libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007ffff67af000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ffff678c000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ffff663d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ffff644b000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ffff7fcf000)
$ readelf -sW /lib/x86_64-linux-gnu/libc.so.6 | grep _Unwind_Find_FDE
$
```

gum/backend-x86/gumstalker-x86.c
```diff
@@ -905,19 +968,274 @@ gum_stalker_init (GumStalker * self)
...
+#ifdef HAVE_LINUX
+  gum_gxx_personality_v0_impl = gum_exec_ctx_find_gxx_personality_v0 ();
+  gum_unwind_find_fde_impl = gum_exec_ctx_find_unwind_find_fde ();
+#endif
+}
+
+#ifdef HAVE_LINUX
+
+static gum_gxx_personality_v0
+gum_exec_ctx_find_gxx_personality_v0 (void)
```
oleavr, line 981: The name `find_*` makes it sounds rather innocent, like it doesn't go ahead and also hook the function. My gut feeling is that we should merge these three functions into one named something like `gum_stalker_ensure_unwind_apis_instrumented`.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,155 @@
...
+static void test_try_and_dont_catch_pp_3 (guint32 * val);
+
+extern "C"
+{
+  void
+  test_check_bit (guint32 * val,
+                  guint8 bit)
+  {
+    g_assert_true (*val & ((guint32) 1 << bit));
+    *val = *val & ~((guint32) 1 << bit);
```
oleavr, line 34: Can't we use `&=` here to simplify this slightly?

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,155 @@
...
+static void test_try_and_dont_catch_pp_2 (guint32 * val);
+static void test_try_and_dont_catch_pp_3 (guint32 * val);
+
+extern "C"
+{
+  void
+  test_check_bit (guint32 * val,
+                  guint8 bit)
+  {
+    g_assert_true (*val & ((guint32) 1 << bit));
```
oleavr, line 33:
```suggestion
    g_assert_true ((*val & ((guint32) 1 << bit)) != 0);
```

For consistency: we only use truthy/falsey checks for gboolean values.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright (C) 2009-2021 Ole André Vadla Ravnås <oleavr@nowsecure.com>
+ * Copyright (C) 2010-2013 Karl Trygve Kalleberg <karltk@boblycat.org>
```
oleavr, line 3: As far as I can tell there's no code here that's derived from existing code, so these two lines seem confusing.

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,155 @@
...
+};
+
+class TestResource
+{
+public:
+  TestResource (guint32 * val)
+    : val (val)
+  {
+    test_set_bit ("TestResource", val, 0);
+  }
```
oleavr, line 76:
```suggestion
  }

```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  test_set_bit ("test_try_and_dont_catch_pp_2", val, 10);
+}
+
+static void
+test_try_and_dont_catch_pp_3 (guint32 * val)
+{
+  test_set_bit ("test_try_and_dont_catch_pp_3", val, 11);
+  throw TestException ();
+  test_set_bit ("test_try_and_dont_catch_pp_3", val, 12);
```
oleavr, line 154:
```suggestion
  test_set_bit ("test_try_and_dont_catch_pp_3", val, 11);

  throw TestException ();

  test_set_bit ("test_try_and_dont_catch_pp_3", val, 12);
```

For consistency with the above.

tests/core/arch-x86/stalker-x86-fixture.c
```diff
@@ -109,12 +109,12 @@ test_stalker_fixture_dup_code (TestStalkerFixture * fixture,
 {
   GumAddressSpec spec;
 
-  spec.near_address = gum_stalker_follow_me;
+  spec.near_address = (gpointer) gum_stalker_follow_me;
```
oleavr, line 112: Why are these casts needed (this one and the ones below); does this file get included by C++ code?

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,176 @@ TESTCASE (ic_var)
...
+    }
+  }
+}
+
+static void
+callout_at_end (GumCpuContext * cpu_context,
+                gpointer user_data)
+{
+  guint32 * val = user_data;
+  *val += (1 << 31);
```
oleavr, line 3388:
```suggestion
  *val += 1 << 31;
```

Redundant since there's nothing else on the right-hand side.

oleavr (changes requested):
:boom: Getting there:

gum/backend-x86/gumstalker-x86.c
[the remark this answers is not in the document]
WorksButNotTested: I have made my changes. You'll see I haven't squashed the commits this time to make it easier to check the changes. I don't think that on Linux at least I can find the statically linked `_Unwind_Find_FDE` symbol without using the `DebugSymbol` API. So I have reverted back to just providing an `extern` definition for the function in question and then letting the linker give me the address (this should provide me with the statically linked symbol if present, or an imported one if not).

You said this would cause issues on Android. I am not sure whether this same exception handling mechanism is used on Android? So for now I have guarded all the exception handling stuff (including the external function definitions) behind a:
```c
#if defined (HAVE_LINUX) && !defined (HAVE_ANDROID)
```.

The only other option I can think of, is if you think the mechanism should still work the same on Android, but that Android won't be affected by the same duplicate symbol issue, we could retain the exception handling mechanism but use compile time `#if` directives to have Linux use the `extern` function definitions, but have Android use a dynamic approach?

Let me know what you think. Cheers.

File: gum/backend-x86/gumstalker-x86.c [hunk 44]
```diff
@@ -759,8 +818,14 @@ static gboolean gum_store_thread_exit_match (GumAddress address, gsize size,
 
 G_DEFINE_TYPE (GumStalker, gum_stalker, G_TYPE_OBJECT)
 
+static GPrivate gum_stalker_exec_context_private;
```
oleavr, line 821:
```suggestion
static GPrivate gum_stalker_exec_ctx_private;
```

For consistency.

gum/backend-x86/gumstalker-x86.c
```diff
@@ -905,19 +969,192 @@ gum_stalker_init (GumStalker * self)
...
+{
+  static gsize gonce_value;
+
+  if (g_once_init_enter (&gonce_value))
+  {
+    GumAttachReturn attach_ret;
+
+    gum_exec_ctx_interceptor = gum_interceptor_obtain ();
+    if (gum_exec_ctx_interceptor == NULL)
+      return;
```
oleavr, line 991: Returning here will leave the `GOnce` in a dangling state. But this function is guaranteed to not return `NULL`, so this check should be removed.

gum/backend-x86/gumstalker-x86.c
```diff
@@ -905,19 +969,192 @@ gum_stalker_init (GumStalker * self)
...
+
+    g_once_init_leave (&gonce_value, 1);
+  }
+}
+
+static void
+gum_stalker_deinit_unwind_apis_instrumentation (void)
+{
+  if (gum_exec_ctx_interceptor == NULL)
+      return;
```
oleavr, line 1020: Can this happen?

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,157 @@
...
+static void test_try_and_dont_catch_pp_3 (guint32 * val);
+
+extern "C"
+{
+  void
+  test_check_bit (guint32 * val,
+                  guint8 bit)
+  {
+    g_assert_true ((*val & ((guint32) 1 << bit)) != 0);
+    *val &= ~(guint32) 1 << bit;
```
oleavr, line 33:
```suggestion
    *val &= ~1U << bit;
```

Here and below I think we can do this instead of casting.

tests/core/arch-x86/stalker-x86-fixture.c
```diff
@@ -114,7 +114,7 @@ test_stalker_fixture_dup_code (TestStalkerFixture * fixture,
 
   if (fixture->code != NULL)
     gum_free_pages (fixture->code);
-  fixture->code = gum_alloc_n_pages_near (
+  fixture->code = (guint8 *) gum_alloc_n_pages_near (
```
oleavr, line 117: Are these three casts needed? (Here and below.)

tests/core/arch-x86/stalker-x86.c
```diff
@@ -203,6 +213,18 @@ G_DEFINE_TYPE_EXTENDED (GumTestStalkerObserver,
...
 static GHashTable * prefetch_executed = NULL;
 static PrefetchBackpatchContext bp_ctx;
+
+static void callback_at_end (GumStalkerIterator * iterator,
+    GumStalkerOutput * output, gpointer user_data);
+static void callout_at_end (GumCpuContext * cpu_context, gpointer user_data);
+static void test_check_followed (void);
+
+extern void __cxa_throw (void * thrownException, void * type,
+    void (* destructor)(void *));
```
oleavr, line 223:
```suggestion
    void (* destructor) (void *));
```

tests/core/arch-x86/stalker-x86.c
```diff
@@ -3201,3 +3223,176 @@ TESTCASE (ic_var)
...
+    }
+  }
+}
+
+static void
+callout_at_end (GumCpuContext * cpu_context,
+                gpointer user_data)
+{
+  guint32 * val = user_data;
+  *val += 1 << 31;
```
oleavr, line 3388:
```suggestion
  *val += 1U << 31;
```

oleavr (changes requested):

gum/backend-x86/gumstalker-x86.c
oleavr: (Also, indentation is wrong here.)

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,157 @@
...
+static void test_try_and_dont_catch_pp_2 (guint32 * val);
+static void test_try_and_dont_catch_pp_3 (guint32 * val);
+
+extern "C"
+{
+  void
+  test_check_bit (guint32 * val,
+                  guint8 bit)
+  {
+    g_assert_true ((*val & ((guint32) 1 << bit)) != 0);
```
oleavr, line 32:
```suggestion
    g_assert_true ((*val & (1U << bit)) != 0);
```

tests/core/arch-x86/stalker-x86-exceptions.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+  void
+  test_set_bit (const char * func,
+                guint32 * val,
+                guint8 bit)
+  {
+    if (g_test_verbose ())
+      g_print ("\tFunc: %s, Set: %d\n", func, bit);
+
+    *val |= (guint32) 1 << bit;
```
oleavr, line 44:
```suggestion
    *val |= 1U << bit;
```

tests/core/arch-x86/stalker-x86-fixture.c
```diff
@@ -156,7 +156,7 @@ test_stalker_fixture_follow_and_invoke (TestStalkerFixture * fixture,
 #endif
   GCallback invoke_func;
 
-  spec.near_address = gum_stalker_follow_me;
+  spec.near_address = (gpointer) gum_stalker_follow_me;
```
oleavr, line 159: Is this still needed?

tests/core/arch-x86/stalker-x86.c
```diff
@@ -203,6 +213,18 @@ G_DEFINE_TYPE_EXTENDED (GumTestStalkerObserver,
 static GHashTable * prefetch_compiled = NULL;
 static GHashTable * prefetch_executed = NULL;
 static PrefetchBackpatchContext bp_ctx;
+
+static void callback_at_end (GumStalkerIterator * iterator,
+    GumStalkerOutput * output, gpointer user_data);
+static void callout_at_end (GumCpuContext * cpu_context, gpointer user_data);
+static void test_check_followed (void);
+
+extern void __cxa_throw (void * thrownException, void * type,
```
oleavr, line 222:
```suggestion
extern void __cxa_throw (void * thrown_exception, void * type,
```

oleavr (changes requested):

PR #612, opened 2021-12-15
Allow building in isolation

meme:
With these changes it is possible to build frida-gum on Ubuntu outside of the Frida monorepo


.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "subprojects/quickjs"]
+	path = subprojects/quickjs
+	url = https://github.com/frida/quickjs
+[submodule "subprojects/capstone"]
+	path = subprojects/capstone
+	url = https://github.com/frida/capstone
+[submodule "subprojects/tinycc"]
+	path = subprojects/tinycc
+	url = https://github.com/frida/tinycc
```
oleavr, line 9: Nit: Could we put Capstone first, then QuickJS, and finally TinyCC? (Following order of appearance in `meson.build`, which indicates how "important" they are.)

oleavr (changes requested):
Thanks! :bomb:

PR #615, opened 2021-12-23
Optimize Stalker for AARCH64

WorksButNotTested:


gum/backend-arm64/gumstalker-arm64.c
```diff
@@ -2398,6 +2788,16 @@ gum_stalker_invoke_callout (GumCalloutEntry * entry,
   ec->pending_calls--;
 }
 
+/*
```
WorksButNotTested: This comment is old and should be updated.

PR #619, opened 2022-01-18
Fix for debug symbols on windows

mundusnine:
Per the windows [docs](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symloadmoduleex):
```
[in] BaseOfDll

The load address of the module. If the value is zero, the library obtains the load address from the symbol file. The load address contained in the symbol file is not necessarily the actual load address. Debuggers and other applications having an actual load address should use the real load address when calling this function.

If the image is a .pdb file, this parameter cannot be zero. <------------------------------

[in] DllSize

The size of the module, in bytes. If the value is zero, the library obtains the size from the symbol file. The size contained in the symbol file is not necessarily the actual size. Debuggers and other applications having an actual size should use the real size when calling this function.

If the image is a .pdb file, this parameter cannot be zero. <------------------------------
```

### DIFF
--- a/gum/backend-dbghelp/gumsymbolutil-dbghelp.c
+++ b/gum/backend-dbghelp/gumsymbolutil-dbghelp.c
@@ -160,13 +160,20 @@ gum_load_symbols (const gchar * path)
 
   path_utf16 = (WCHAR *) g_utf8_to_utf16 (path, -1, NULL, NULL, NULL);
 
-  base = GPOINTER_TO_SIZE (GetModuleHandleW (path_utf16));
+  HMODULE mod = GetModuleHandleW(path_utf16);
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! Seeing as you closed this PR, I'm curious what you learned since opening it? (I can't remember a whole lot about this part of the codebase.)

mundusnine:
Hi,

I tested with a simple usecase i.e. just dumping the .pdb data from the .exe. From my test I think I don't quite understand yet how it all works on the gum side. I thought the issue was not using the size like I said in the PR but as my [issue](https://github.com/frida/frida-gum/issues/621) states, loading of symbols fails also for dll's. I will make a minimal project to test this.

PR #623, opened 2022-01-29
Add console.count and console.countReset methods

yotamN:
Copied the API from the web standards

### DIFF
--- a/bindings/gumjs/runtime/console.js
+++ b/bindings/gumjs/runtime/console.js
@@ -4,6 +4,12 @@ const engine = global;
 const slice = Array.prototype.slice;
 
 class Console {
+  #counters;
+
+  constructor() {
+    this.#counters = new Map();
+  }
+
   log() {
     sendLogMessage('info', slice.call(arguments));
   }
@@ -14,6 +20,20 @@ class Console {
 
   error() {
     sendLogMessage('error', slice.call(arguments));
+  }
+
+  count(label = 'default') {
+    const newValue = (this.#counters.get(label) ?? 0) + 1;
+    this.#counters.set(label, newValue);
+    this.log(`${label}: ${newValue}`);
+  }
+
+  countReset(label = 'default') {
+    if (this.#counters.has(label)) {
+      this.#counters.delete(label);
+    } else {
+      this.warn(`Counter "${label}" doesn't exist.`);
+    }
   }
 }
 
--- a/bindings/gumjs/runtime/console.js
+++ b/bindings/gumjs/runtime/console.js
@@ -32,7 +32,7 @@ class Console {
     if (this.#counters.has(label)) {
       this.#counters.delete(label);
     } else {
-      this.warn(`Counter "${label}" doesn't exist.`);
+      this.warn(`Count for '${label}' does not exist`);
     }
   }
 }


bindings/gumjs/runtime/console.js
```diff
@@ -4,6 +4,12 @@ const engine = global;
 const slice = Array.prototype.slice;
 
 class Console {
+  #counters;
+
+  constructor() {
+    this.#counters = {};
```
oleavr, line 10: Would be good to use a `Map` here.

oleavr (changes requested):
Thanks, great idea!

bindings/gumjs/runtime/console.js
yotamN: I switched to a `Map`

bindings/gumjs/runtime/console.js
```diff
@@ -15,6 +21,24 @@ class Console {
...
     sendLogMessage('error', slice.call(arguments));
   }
+
+  count(label = 'default') {
+    if (!this.#counters.has(label)) {
+      this.#counters.set(label, 0);
+    }
+
+    const newValue = this.#counters.get(label) + 1;
+    this.#counters.set(label, newValue);
```
oleavr, line 31: This could be simplified to avoid two lookups per label. Perhaps something like:

```suggestion
    const newValue = (this.#counters.get(label) ?? 0) + 1;
    this.#counters.set(label, newValue);
```

bindings/gumjs/runtime/console.js
```diff
@@ -15,6 +21,24 @@ class Console {
...
+    this.log(`${label}: ${newValue}`);
+  }
+
+  countReset(label = 'default') {
+    if (!this.#counters.has(label)) {
+      this.warn(`Counter "${label}" doesn't exist.`);
```
oleavr, line 37: Perhaps we should reverse these two branches so the happy/normal-case is first.

Also it would be good to `.delete()`, so the `Map` doesn't keep growing in case there's a lot of dynamically generated labels over time.

oleavr (changes requested):

yotamN:
Fixed both. I also noticed that the browser implementation prints 0 in countReset but it doesn't really makes sense. Do you want to maintain browser compatibility or not?

oleavr:
> Fixed both. I also noticed that the browser implementation prints 0 in countReset but it doesn't really makes sense. Do you want to maintain browser compatibility or not?

Thanks! I don't think so. At least Chrome 97 doesn't seem to exhibit such behavior:

Screenshot 2022-02-01 at 11 11 46 AM

oleavr:
Ah I see, Firefox behaves that way. Yeah I think we can ignore that 👍

bindings/gumjs/runtime/console.js
```diff
@@ -15,6 +21,20 @@ class Console {
...
+    const newValue = (this.#counters.get(label) ?? 0) + 1;
+    this.#counters.set(label, newValue);
+    this.log(`${label}: ${newValue}`);
+  }
+
+  countReset(label = 'default') {
+    if (this.#counters.has(label)) {
+      this.#counters.delete(label);
+    } else {
+      this.warn(`Counter "${label}" doesn't exist.`);
```
oleavr, line 35:
```suggestion
      this.warn(`Count for '${label}' does not exist`);
```

oleavr: (Figured the Chrome wording is slightly clearer.)

PR #631, opened 2022-02-26
gumjs: Add Process.operatingSystem

vfsfitvnm:
Hi!
So there was a need to add a property (next to `Process.platform`) to determine the operating system (#192).

I added a (unneeded?) macro `GUM_SCRIPT_OS` which is pretty similar to this [one](https://github.com/frida/frida-gum/blob/4233137e80bc60cb00a0623201273f04f2d9e29c/tests/core/interceptor-fixture.c#L39). The alternative consisted in using the already existing [gum_process_get_native_os](https://github.com/frida/frida-gum/blob/4233137e80bc60cb00a0623201273f04f2d9e29c/gum/gumprocess.c#L46) and then retrieve a string from it, but it didn't look as simple as the macro solution.

Two doubts:

1) Should I have added a default case? (e.g. `"unknown"`)
2) Should I have handled darwin-based operating system which aren't MacOs nor iOS? (`defined (HAVE_DARWIN) && !defined (HAVE_MACOS) && !defined (HAVE_IOS)`)

Cheers

### DIFF
--- a/bindings/gumjs/gumquickprocess.c
+++ b/bindings/gumjs/gumquickprocess.c
@@ -38,6 +38,22 @@
 # define GUM_SCRIPT_PLATFORM "qnx"
 #endif
 
+#if defined (HAVE_LINUX) && !defined (HAVE_ANDROID)
+# define GUM_SCRIPT_OS "linux"
+#elif defined (HAVE_ANDROID)
+# define GUM_SCRIPT_OS "android"
+#elif defined (HAVE_MACOS)
+# define GUM_SCRIPT_OS "macos"
+#elif defined (HAVE_IOS)
+# define GUM_SCRIPT_OS "ios"
+#elif defined (HAVE_WINDOWS)
+# define GUM_SCRIPT_OS "windows"
+#elif defined (HAVE_FREEBSD)
+# define GUM_SCRIPT_OS "freebsd"
+#elif defined (HAVE_QNX)
+# define GUM_SCRIPT_OS "qnx"
+#endif
+
 typedef struct _GumQuickMatchContext GumQuickMatchContext;
 typedef struct _GumQuickFindModuleByNameContext GumQuickFindModuleByNameContext;
 typedef struct _GumQuickFindRangeByAddressContext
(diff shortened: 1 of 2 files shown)


oleavr:
Hey, thanks for doing this!

I think if we are going down this road then we should match our existing system parameters' `os` parameter both in name and contents. It is computed here: https://github.com/frida/frida-core/blob/1055b0d494b7123165b9b783715bb1bec60afc63/lib/base/session.vala#L1158-L1218

What do you think?

vfsfitvnm:
Hi, I honestly didn't look at `frida-core` side.

However, I don't think there is a **strict** need to add the other information -  for the sake of keeping the codebase as simple as possible: I couldn't find any issue (on GitHub nor Telegram) about a "platform version".

At the moment, it's very hard to tell if the script is running on MacOS or iOS, but `frida-gum` does always know this piece of information at 0-cost (this is untrue for the others); also, once the user knows what the operating system is, he could determine the other information if he really needs to. But, as of now, the only way to determine the operating system quickly is using dirty tricks (e.g. looking for specific exports/modules) - it feels odd given `frida-gum` has this information out of the box.

You may be wondering if there will be a breaking change when it'll come to adding e.g. the operating system version. Of course it will, but what can we say about this?
```ts
declare namespace Process {

    const platform: Platform;

    const operatingSystem: {
        id: "windows" | "macos" | "ios" | "linux" | "android" | "qnx" | "freebsd" | string,
        name: string?,
        version: string?
    };
}
``` 
Wouldn't it make more sense to move `platform` inside `operatingSystem` (e.g. in a `family` property)? If so, that would be a breaking change as well - it would be odd to have another operating system property outside the `operatingSystem` object in my opinion.
Moreover, #192 has been there for five years, I doubt there will be an issue about retrieving the operating system version anytime soon.

In conclusion, it looks like the effort in matching `compute_system_parameters` would be somewhat unrewarded.

That's why adding `Process.operatingSystem` is, in my opinion, the simplest and quickest way to address the majority (if not the totality) of the operating-system-details issues, without breaking `Process.platform` and without having an odd API.

oleavr:
Hmm yeah, I think you're right. I'm leaning towards naming it `Process.os`, and making that an enumeration that matches the values in frida-core. Does that make sense?

vfsfitvnm:
`Process.os` is fine; however how do you intend matching `frida-core` values? I guess you are referring to `os["id"]` - if so, should we care about a precise Linux distro id?

Another solution consists in adding two properties (`isAndroid` / `isIOS`) to address the two cases where it's difficult to determine whether the script is running in Android or iOS, given `Process.platform` is `"linux"` or `"darwin"`.

In summary I see four solutions:

1) Editing `Process.platform`
    - Simple and clean
    - Breaking
 2) Adding `Process.os`
    - Looks somehow avoidable
    - Creates inconsistency with `frida-core` if we follow the KISS principle
    - Unnecessarily complicated and unneeded if we match `frida-core`
    - Not breaking
3) Adding `Process.isAndroid` and `Process.isIOS`
    - Looks somehow avoidable
    - `Process` may not be the correct namespace to use (maybe one of `Frida`, `Script` or a brand new `Platform` is a better candidate?)
    - Not breaking
  4) Statu quo ante
     - 😢

PR #653, opened 2022-05-06
darwin-module: Handle special lib ordinal values

mrmacete:
In this way, if an import's module can't be resolved by direct ordinal index, the import address is resolved immediately using `dlsym` and the corresponding module path is resolved anyways at emit time (like it's always been) with a new fallback on `dladdr` when `ModuleMap` fails to resolve it.


gum/gumdarwinmodule.c
```diff
@@ -20,6 +20,11 @@
 #define GUM_DARWIN_MODULE_HAS_HEADER_ONLY(self) \
     ((self->flags & GUM_DARWIN_MODULE_FLAGS_HEADER_ONLY) != 0)
 
+#define GUM_SPECIAL_DYLIB_SELF 0
+#define GUM_SPECIAL_DYLIB_MAIN_EXECUTABLE 0xff
+#define GUM_SPECIAL_DYLIB_FLAT_LOOKUP 0xfe
+#define GUM_SPECIAL_DYLIB_WEAK_LOOKUP 0xfd
```
oleavr, line 26: I think we should be using `GUM_BIND_SPECIAL_DYLIB_` instead of these, and make sure we interpret the `lib_ordinal` in the chained imports as a signed 8-bit integer. (Unless they use the bits below these as indices, in which case we'll need to do the special-casing in `gum_emit_chained_imports()`.)

gum/backend-darwin/gumprocess-darwin.c
```diff
@@ -1885,6 +1885,7 @@ gum_emit_import (const GumImportDetails * details,
 {
   GumEnumerateImportsContext * ctx = user_data;
   GumImportDetails d;
+  Dl_info info;
```
oleavr, line 1888: (You should add yourself to the copyright header in this file by the way.)

gum/gumdarwinmodule.c
```diff
@@ -1896,6 +1901,16 @@ const gchar *
...
+  switch (ordinal)
+  {
+    case GUM_SPECIAL_DYLIB_SELF:
+      return self->name;
+    case GUM_SPECIAL_DYLIB_MAIN_EXECUTABLE:
+    case GUM_SPECIAL_DYLIB_FLAT_LOOKUP:
+    case GUM_SPECIAL_DYLIB_WEAK_LOOKUP:
+      return NULL;
+  }
+
```
oleavr, line 1913: (You should add yourself to the copyright header in this file by the way.)

gum/backend-darwin/gumprocess-darwin.c
oleavr, line 1888: It looks like this variable can be moved down to the next scope inside this one.

oleavr (changes requested):
Thanks! 💥

gum/backend-darwin/gumprocess-darwin.c
mrmacete: oh yeah right, because the string we use doesn't actually live on the stack

mrmacete: oh yeah right, because the string we use doesn't actually live on the stack

gum/gumdarwinmodule-priv.h
```diff
@@ -34,6 +34,7 @@
 #define GUM_BIND_SPECIAL_DYLIB_SELF             0
 #define GUM_BIND_SPECIAL_DYLIB_MAIN_EXECUTABLE -1
 #define GUM_BIND_SPECIAL_DYLIB_FLAT_LOOKUP     -2
+#define GUM_BIND_SPECIAL_DYLIB_WEAK_LOOKUP     -3
```
oleavr, line 37: You should add yourself to the copyright header in this file too :)

gum/gumdarwinmodule.c
```diff
@@ -1896,6 +1897,16 @@ const gchar *
...
+  {
+    case GUM_BIND_SPECIAL_DYLIB_SELF:
+      return self->name;
+    case GUM_BIND_SPECIAL_DYLIB_MAIN_EXECUTABLE:
+    case GUM_BIND_SPECIAL_DYLIB_FLAT_LOOKUP:
+    case GUM_BIND_SPECIAL_DYLIB_WEAK_LOOKUP:
+      return NULL;
+  }
+
   gint i = ordinal - 1;
```
oleavr, line 1910: This declaration should be moved up, but leaving the assignment here.

oleavr: (Or I suppose we could move the assignment down to right above where it's first used.)

gum/gumdarwinmodule.c
```diff
@@ -1896,7 +1897,19 @@ const gchar *
...
+    case GUM_BIND_SPECIAL_DYLIB_MAIN_EXECUTABLE:
+    case GUM_BIND_SPECIAL_DYLIB_FLAT_LOOKUP:
+    case GUM_BIND_SPECIAL_DYLIB_WEAK_LOOKUP:
+      return NULL;
+  }
+
+  i = ordinal - 1;
 
   if (!gum_darwin_module_ensure_image_loaded (self, NULL))
     return NULL;
```
oleavr, line 1915: We should move this first, so `self->name` is initialized.

oleavr (changes requested):

gum/gumdarwinmodule.c
mrmacete: oops!

PR #657, opened 2022-05-29
gumjs: add quick cmd support

regomne:
The feature "quick command" in frida REPL, like:

```
[Local::a.exe ]-> .d 401000 10
           0  1  2  3  4  5  6  7  8  9  A  B  C  D  E  F  0123456789ABCDEF
00401000  55 8b ec 81 c4 fc fe ff ff 8b 45 0c 83 f8 01 75  U.........E....u
```

The quick cmd starts with '.' (same as nodejs REPL), and can be customized dynamically by the javascript loaded.

For example, register a new quick cmd 'u' in `agent.js` :

```js
REPL.register_quick_cmd(
    'u', // the quick cmd
    1, // at least 1 argument
    function disasm(addr, inst_cnt = 10) { ... } // target function
)
```

Which can be used in REPL:

```
[Local::a.exe ]-> .u 401000
00401000    push ebp
00401001    mov ebp, esp
00401003    add esp, 0xfffffefc
00401009    mov eax, dword ptr [ebp + 0xc]
0040100C    cmp eax, 1
0040100F    jne 0x401028
00401011    push dword ptr [0x403864]
00401017    push 1
00401019    push 0x466
0040101E    push dword ptr [ebp + 8]
```

### DIFF
--- a/bindings/gumjs/runtime/core.js
+++ b/bindings/gumjs/runtime/core.js
@@ -1,6 +1,7 @@
 const Console = require('./console');
 const hexdump = require('./hexdump');
 const MessageDispatcher = require('./message-dispatcher');
+const REPL = require('./repl');
 
 const engine = global;
 let messageDispatcher;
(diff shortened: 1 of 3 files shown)


oleavr:
Thanks! (Sorry for not noticing this PR before.)

Unfortunately this code does not belong here -- frida-tools' repl.py already has its own agent that adds code to the global scope. So this should be added over there.

PR #667, opened 2022-07-10
Interceptor: Provide original function address back to caller

aviramha:
Followup on: https://github.com/frida/frida-rust/issues/44

### DIFF
--- a/gum/guminterceptor.c
+++ b/gum/guminterceptor.c
@@ -514,7 +514,8 @@ GumReplaceReturn
 gum_interceptor_replace (GumInterceptor * self,
                          gpointer function_address,
                          gpointer replacement_function,
-                         gpointer replacement_data)
+                         gpointer replacement_data,
+                         gpointer * original_function)
 {
   GumReplaceReturn result = GUM_REPLACE_OK;
   GumFunctionContext * function_ctx;
@@ -535,6 +536,9 @@ gum_interceptor_replace (GumInterceptor * self,
 
   function_ctx->replacement_data = replacement_data;
   function_ctx->replacement_function = replacement_function;
+
+  if (original_function != NULL)
+    *original_function = function_ctx->on_invoke_trampoline;
 
   goto beach;
 
--- a/tests/core/interceptor.c
+++ b/tests/core/interceptor.c
@@ -62,6 +62,7 @@ TESTLIST_BEGIN (interceptor)
 # endif
 #endif
   TESTENTRY (replace_then_attach)
+  TESTENTRY (replace_keep_original)
 TESTLIST_END ()
 
 #ifdef HAVE_WINDOWS
@@ -613,7 +614,7 @@ TESTCASE (replace_one)
   malloc_impl = interceptor_fixture_get_libc_malloc ();
 
   g_assert_cmpint (gum_interceptor_replace (fixture->interceptor, malloc_impl,
-      replacement_malloc, &counter), ==, GUM_REPLACE_OK);
+      replacement_malloc, &counter, NULL), ==, GUM_REPLACE_OK);
   ret = malloc_impl (0x42);
 
   /*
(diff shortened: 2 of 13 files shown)


gum/guminterceptor.c
```diff
@@ -535,6 +536,9 @@ gum_interceptor_replace (GumInterceptor * self,
 
   function_ctx->replacement_data = replacement_data;
   function_ctx->replacement_function = replacement_function;
+  if (original_function != NULL) {
+    original_function = function_ctx->on_invoke_trampoline;
```
meme, line 540: This should be `*original_function =` so that the caller's variable is updated. This code changes where `original_function` points rather than the value of the caller's pointer.

meme (changes requested):

gum/guminterceptor.c
aviramha: Yup that's what I did originally then when adding the if I accidentally dropped it. Thanks

gum/guminterceptor.c
```diff
@@ -535,6 +536,9 @@ gum_interceptor_replace (GumInterceptor * self,
 
   function_ctx->replacement_data = replacement_data;
   function_ctx->replacement_function = replacement_function;
+  if (original_function != NULL) {
+    *original_function = function_ctx->on_invoke_trampoline;
+  }
```
oleavr, line 541:
```suggestion

  if (original_function != NULL)
    *original_function = function_ctx->on_invoke_trampoline;
```

tests/core/interceptor.c
```diff
@@ -654,9 +655,9 @@ TESTCASE (replace_two)
   free_impl = interceptor_fixture_get_libc_free ();
 
   gum_interceptor_replace (fixture->interceptor, malloc_impl,
-      replacement_malloc_calling_malloc_and_replaced_free, &malloc_counter);
+      replacement_malloc_calling_malloc_and_replaced_free, &malloc_counter, NULL);
```
oleavr, line 658:
```suggestion
      replacement_malloc_calling_malloc_and_replaced_free, &malloc_counter,
      NULL);
```

tests/core/interceptor.c
```diff
@@ -817,3 +818,37 @@ replacement_malloc (gsize size)
 
   return GSIZE_TO_POINTER (size);
 }
+
+TESTCASE (replace_keep_original)
```
oleavr, line 822: Would be good to move this so it follows right after the `replace_then_attach`-test. (The test list at the top defines the order that the tests run in. This may differ from the order they're defined in below it. When this is the case, it is because we want lower level tests to run first, as those might detect defects that are much harder to debug in the higher level tests that follow. However, the order of the definitions, i.e. the code, should follow newspaper ordering — the reader should see the higher level view and as they read further down, things get increasingly detailed. This is why e.g. `main()` is always first in a typical Frida program.)

tests/core/interceptor.c
```diff
@@ -817,3 +818,37 @@ replacement_malloc (gsize size)
...
+  if (RUNNING_ON_VALGRIND)
+  {
+    g_print ("<skipping, not compatible with Valgrind> ");
+    return;
+  }
+
+  malloc_impl = interceptor_fixture_get_libc_malloc ();
+
+  g_assert_cmpint (gum_interceptor_replace (fixture->interceptor, malloc_impl,
+      replacement_malloc, &counter, (void **)&original_impl), ==, GUM_REPLACE_OK);
```
oleavr, line 838:
```suggestion
      replacement_malloc, &counter, (void **) &original_impl),
      ==, GUM_REPLACE_OK);
```

(Maximum of 80 cols in Gum itself — language bindings may diverge as they should follow a style that makes sense for the specific language/culture.)

tests/core/interceptor.c
```diff
@@ -817,3 +818,37 @@ replacement_malloc (gsize size)
...
+  g_assert_nonnull(original_impl);
```
oleavr, line 839:
```suggestion
  g_assert_nonnull (original_impl);
```

(There's a style guide [here](https://github.com/frida/frida/blob/main/CONTRIBUTING.md) by the way.)

tests/core/interceptor.c
```diff
@@ -817,3 +818,37 @@ replacement_malloc (gsize size)
...
+   * asserts with broken malloc() are quite tricky to debug. :)
+   */
+  g_assert_nonnull (ret);
+
+  gum_interceptor_revert (fixture->interceptor, malloc_impl);
+  g_assert_cmpint (counter, ==, 0);
+  g_assert_cmphex (GPOINTER_TO_SIZE (ret), !=, 0x42);
+
+  free (ret);
+}
```
oleavr, line 854: Missing newline. (Older versions of GCC complain about this.)

oleavr (changes requested):
Thanks! 🙌 (Sorry for the delay, I'm currently on vacation. Will be more responsive when I'm back next week.)

aviramha:
I think I addressed most of the comments - Sorry for not being aware of the style guide. I got too comfortable having `cargo fmt` :)

aviramha:
Thanks for merging this @oleavr ! How does release work? i.e when will it be available as part of a release?

PR #672, opened 2022-08-02
Various ARM32 stalker backend improvements

WorksButNotTested:


oleavr:
Thanks! Just landed this with many style tweaks. (Didn't want to lose any history here in case I screwed up something, so left your branch untouched.)

PR #682, opened 2022-10-14
arm64: update cpu context in glue prolog / epilog

mrmacete:
In this way is consistent with the latest changes where `nzcv` and all vector registers have been added to the arm64 cpu context.

### DIFF
--- a/gum/backend-arm64/guminterceptor-arm64-glue.S
+++ b/gum/backend-arm64/guminterceptor-arm64-glue.S
@@ -1,5 +1,5 @@
-#define GUM_PADDING_SIZE              8
-#define GUM_CPU_CONTEXT_SIZE          ((33 * 8) + (8 * 16))
+#define GUM_PADDING_SIZE              0
+#define GUM_CPU_CONTEXT_SIZE          ((34 * 8) + (32 * 16))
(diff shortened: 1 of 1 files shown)


gum/backend-arm64/guminterceptor-arm64-glue.S
```diff
@@ -23,58 +23,86 @@
   sub sp, sp, GUM_FRAME_SIZE - (2 * 8)
 
   # We want to clobber x1, so let's save it (and x2) first
-  stp x1, x2, [sp, GUM_PADDING_SIZE + (3 * 8)]
+  stp x1, x2, [sp, GUM_PADDING_SIZE + (4 * 8)]
+
+  # sample nzcv
+  mrs x1, nzcv
```
mrmacete, line 29: i'm pretty sure `nzcv` is partially clobbered for the previous `sub`... not sure how to do better 🤔

oleavr: This is fine since `sub` doesn't clobber, but `subs` does. (If this later turns out to be incorrect we could always use `x16` as a scratch register, since it's already saved.)

oleavr (approved):
Thanks! :raised_hands:

PR #687, opened 2022-11-10
gumjs: Add support for 'none' code trap option

mrmacete:
Setting this when constructing a `NativeFunction` will prevent both `Interceptor` and `Stalker` from triggering when it is called.


oleavr:
Thanks! 🙌

PR #696, opened 2022-11-30
darwin-mapper: Fix macro declaration

mrmacete:
The `GUM_ADVANCE_BY()` macro was missing a block scope. When it was used in the context of an `if()` also without scope, a misaligment could have occured between the `pc` and `cursor` offsets.

That happening on the chained symbols array of pointers could have caused the fixup processor to bind invalid pointers, leading to crashes.

When that happened while injecting `launchd`, it caused in turn a kernel panic on arm64e iOS.


oleavr:
Thanks!! 💥 Great catch! 🤘

PR #697, opened 2022-12-05
Fix failure to wait on thread in gum_do_modify_thread

WorksButNotTested:


gum/backend-linux/gumprocess-linux.c
```diff
@@ -551,11 +552,20 @@ gum_do_modify_thread (gpointer data)
...
+  wait_result = gum_libc_waitpid (ctx->thread_id, &status, __WALL);
+
+  if (wait_result != ctx->thread_id)
+    goto failed_to_wait;
+
+  if (!WIFSTOPPED (status))
     goto failed_to_stop;
 
+  /*
+   * Although ptrace injects SIGSTOP into our process, it is possible that our
```
oleavr, line 564:
```suggestion
   * Although ptrace injects SIGSTOP into our process, it is possible that our
```

gum/backend-linux/gumprocess-linux.c
```diff
@@ -551,11 +552,20 @@ gum_do_modify_thread (gpointer data)
     goto failed_to_attach;
   attached = TRUE;
 
-  wait_result = gum_libc_waitpid (ctx->thread_id, &status, 0);
-  if (wait_result != ctx->thread_id || !WIFSTOPPED (status) ||
-      WSTOPSIG (status) != SIGSTOP)
+  wait_result = gum_libc_waitpid (ctx->thread_id, &status, __WALL);
```
oleavr, line 555:
```suggestion
  wait_result = gum_libc_waitpid (ctx->thread_id, &status, __WALL);
```

When is `__WALL` needed here?

gum/backend-linux/gumprocess-linux.c
```diff
@@ -568,7 +578,7 @@ gum_do_modify_thread (gpointer data)
   if (res == -1)
     goto failed_to_write;
 
-  res = gum_libc_ptrace (PTRACE_DETACH, ctx->thread_id, NULL, NULL);
+  res = gum_libc_ptrace (PTRACE_DETACH, ctx->thread_id, NULL, (void *) SIGCONT);
```
oleavr, line 581: Won't this swallow the signal if the process was stopped by a different signal than `SIGSTOP`?

gum/backend-linux/gumprocess-linux.c
```diff
@@ -605,7 +620,7 @@ gum_do_modify_thread (gpointer data)
 beach:
   {
     if (attached)
-      gum_libc_ptrace (PTRACE_DETACH, ctx->thread_id, NULL, NULL);
+      gum_libc_ptrace (PTRACE_DETACH, ctx->thread_id, NULL, (void *) SIGCONT);
```
oleavr, line 623:
```suggestion
      gum_libc_ptrace (PTRACE_DETACH, ctx->thread_id, NULL, (void *) SIGCONT);
```

Should always use GLib's casts for `void *` casts, e.g. `GINT_TO_POINTER()`.

oleavr (changes requested):
Thanks! 🙌

gum/backend-linux/gumprocess-linux.c
WorksButNotTested: The man page states...

```
__WCLONE
    Wait for "clone" children only. If omitted then wait for "non-clone" children only. (A "clone" child is one which
    delivers no signal, or a signal other than SIGCHLD to its parent upon termination.) This option is ignored if __WALL
    is also specified. 
__WALL (since Linux 2.4)
    Wait for all children, regardless of type ("clone" or "non-clone"). 
```

Experimentation has shown that on old 3.x kernels, the wait is not satisfied without `__WALL` being specified. This flag will mainly effect behaviour in the instance that the PID is specified to wait for multiple threads (e.g. <-1, -1, or 0), but since we are waiting for a specified thread, it cannot have any negative consequences.

gum/backend-linux/gumprocess-linux.c
WorksButNotTested: The `ptrace` man page says...

```
PTRACE_ATTACH
              Attach to the process specified in pid, making it a tracee
              of the calling process.  The tracee is sent a SIGSTOP, but
              will not necessarily have stopped by the completion of
              this call; use [waitpid(2)](https://man7.org/linux/man-pages/man2/waitpid.2.html)
              to wait for the tracee to stop.

              See the "Attaching and detaching" subsection for
              additional information.  (addr and data are ignored.)

              Permission to perform a PTRACE_ATTACH is governed by a
              ptrace access mode PTRACE_MODE_ATTACH_REALCREDS check; see
              below.
```

Therefore even if the tracee hasn't stopped, we have already injected SIGSTOP into the process, and so here we would send a corresponding `SIGCONT` to cause it to resume.

gum/backend-linux/gumprocess-linux.c
WorksButNotTested: Done.

gum/backend-linux/gumprocess-linux.c
```diff
@@ -551,11 +552,20 @@ gum_do_modify_thread (gpointer data)
...
+    goto failed_to_wait;
+
+  if (!WIFSTOPPED (status))
     goto failed_to_stop;
 
+  /*
+   * Although ptrace injects SIGSTOP into our process, it is possible that our
+   * target is stopped by another stop signal (e.g. SIGTTIN). The man pages for
+   * ptrace mention the possible race condition. For our purposes, however, we
+   * only require that the target is stopped so that we can read it's registers.
```
oleavr, line 567:
```suggestion
   * only require that the target is stopped so that we can read its registers.
```

oleavr (approved):
💥

PR #702, opened 2022-12-22
Interceptor: Flip page protection on Darwin grafted import activation

mrmacete:
Also add `gum_darwin_query_protection()` as a very fast way to get the current page protection of any address on Darwin.


oleavr:
Thanks! 🤘

PR #703, opened 2022-12-22
Add support for reading thread names

WorksButNotTested:
...on Linux (though adding support for other platforms should be simple).


hexploitable:
Hey @WorksButNotTested I just opened a PR for this too (#706), with Linux + Darwin. Your Linux code is a little neater, so I'll update my PR with your changes for linux, and I'll finalize the changes for Darwin too. Will need to investigate and add Windows/FreeBSD too and then it should be done. Are you cool with that?

WorksButNotTested:
More than happy for you to take this forward. My focus is entirely on Linux, but if you can add support for other operating systems too that would be awesome. Not sure how things work on android? Maybe the same as Linux? But don’t think I catered for that either.

hexploitable:
Yeah so my code was tested on Android and not Linux, and so they are both working. I'll carry it forward then - sweet

PR #705, opened 2022-12-23
Thread time

WorksButNotTested:
Please note that this PR also contains the following:
* https://github.com/frida/frida-gum/pull/762

This is because the RunOnThread functionality is necessary for it to function.
This PR adds support for reading user_time (the time the CPU has spent executing in user-mode in microseconds) on Linux and Android based systems. Similar support for other systems should be relatively straightforward, but aren't included in this PR.

### DIFF
--- a/bindings/gumjs/gumquickthread.c
+++ b/bindings/gumjs/gumquickthread.c
@@ -17,11 +17,13 @@ enum _GumBacktracerType
 
 GUMJS_DECLARE_FUNCTION (gumjs_thread_backtrace)
 GUMJS_DECLARE_FUNCTION (gumjs_thread_sleep)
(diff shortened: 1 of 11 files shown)


WorksButNotTested:
That sounds good. I can add the sampler pretty easily, but wasn't sure where to start with the JS bindings without ending up with a load of duplication?

oleavr:
> That sounds good. I can add the sampler pretty easily, but wasn't sure where to start with the JS bindings without ending up with a load of duplication?

Awesome! I started on JS bindings [here](https://github.com/frida/frida-gum/commits/feature/gumjs-profiler-apis/), though there's not much there it might provide some useful context. Can't believe it's been 7 years since then! 🫠

WorksButNotTested:
Superseded by https://github.com/frida/frida-gum/pull/800

PR #709, opened 2023-01-30
Darwin: Fixup TLV thunk functions during mapping

fabianfreyer:
See-also: https://github.com/frida/frida/issues/2395

### DIFF
--- a/gum/arch-arm64/gumarm64writer.c
+++ b/gum/arch-arm64/gumarm64writer.c
@@ -1808,6 +1808,26 @@ gum_arm64_writer_put_load_store_pair (GumArm64Writer * self,
       (rt2 << 10) | (rn << 5) | rt);
 }
 
+gboolean
+gum_arm64_writer_put_mrs (GumArm64Writer * self,
+                          arm64_reg dst_reg,
+                          guint16 systemreg)
+{
+  GumArm64RegInfo rt;
+
+  gum_arm64_writer_describe_reg (self, dst_reg, &rt);
+
+  if ((rt.width != 64) || (systemreg & 0x8000))
+    return FALSE;
+
+  gum_arm64_writer_put_instruction (self,
+      0xd5300000 |
+      (systemreg << 5) |
+      rt.index);
+
+  return TRUE;
+}
+
 void
 gum_arm64_writer_put_instruction (GumArm64Writer * self,
                                   guint32 insn)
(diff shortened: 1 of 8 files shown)


fabianfreyer:
This still needs some work for 32 bit systems, only tested on 64 bit -- probably missing a couple of `#ifdef`'s.

There's a couple of places where I'm unsure about the code:

https://github.com/frida/frida-gum/blob/2b0eaa1c863ce75b36fe5b772a67082695297c6f/gum/gumdarwinmodule.c#L1148
Here, we need to add file_offset onto the address of the data, but here, that doesn't work and we need to use the `vmaddr`:
https://github.com/frida/frida-gum/blob/2b0eaa1c863ce75b36fe5b772a67082695297c6f/gum/backend-darwin/gumdarwinmapper.c#L670-L679

Additionally, the above was tested on macOS and works there. I'm assuming that this works because `libdyld.dylib`, as part of the DSC, is mapped at the same address both in the frida-helper and the target process. Since the resolver in frida-core's  [`_frida_darwin_helper_backend_inject_into_task `](
https://github.com/frida/frida-core/blob/051cd9738460e83233625b4c3942bf492d1044f6/src/darwin/frida-helper-backend-glue.m#L2012-L2043) is created for the target task,
```c
  resolver = gum_darwin_module_resolver_new (task, &io_error);
```
I'm not sure whether `gum_darwin_module_resolver_find_module` here ensures that the resolved module is copied over into the helper in situations where we're using a device so that `gum_darwin_mapper_find_tlv_get_addr` operates on the data in the target process:
https://github.com/frida/frida-gum/blob/2b0eaa1c863ce75b36fe5b772a67082695297c6f/gum/backend-darwin/gumdarwinmapper.c#L747-L751

fabianfreyer:
Tested on M1 / amd64 macOS

oleavr (changes requested):
Thanks! 🙌

This will need some style fixes. The style is documented [here](https://github.com/frida/frida/blob/main/CONTRIBUTING.md).

(Happy to help with tweaking the style once I have a bit more bandwidth, but leaving this note here in the meantime.)

fabianfreyer:
I tried to fix the style issues -- in case I overlooked some rule, please let me know :)

fabianfreyer:
Rebased, added runtime code to copy over the initial values for the Frida thread, and added somewhat-verbose commit messages describing what's happening.

oleavr:
Landed in 5f814842cbde0ed60c300cb8dccb70de10c3bd41. Thanks! 🙌

fabianfreyer:
Wonderful, thank you so much!

PR #712, opened 2023-02-02
fix detection of libv8

milahu:
fix build with libv8-10.8.25 from nodejs-19.4.0

> Run-time dependency v8-10.0 found: NO (tried pkgconfig and cmake)
>
> meson.build:532:2: ERROR: Dependency "v8-10.0" not found, tried pkgconfig and cmake

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -529,13 +529,13 @@ if have_gumjs
     quickjs_dep_native = dependency('quickjs', native: true)
   endif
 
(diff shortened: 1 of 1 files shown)


milahu:
wrong fix. frida-gum requires https://github.com/frida/v8 which provides /lib/pkgconfig/v8-10.0.pc

```
Name: V8
Description: V8 JavaScript Engine
Version: 10.9.42
```

PR #721, opened 2023-02-11
meson: Require Frida fork of V8

milahu:
part of https://github.com/frida/frida-core/pull/452

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -47,6 +47,7 @@ if host_os_family == 'darwin'
 endif
 
 cc = meson.get_compiler('c')
+cppc = meson.get_compiler('cpp')
 
 frida_component_cflags = []
 ndebug = get_option('b_ndebug')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -529,10 +530,23 @@ if have_gumjs
     quickjs_dep_native = dependency('quickjs', native: true)
   endif
 
-  v8_dep = dependency('v8-10.0',
-    version: '>=10.6.122',
-    required: get_option('v8'),
+  # workaround for https://github.com/mesonbuild/meson/issues/11381
+  if get_option('v8').auto()
+    error('option "v8" cannot be "auto", must be either "enabled" or "disabled"')
+  endif
```
oleavr, line 536: I take it that this is an outdated solution?

oleavr (changes requested):

meson.build
milahu: fixed

meson.build
```diff
@@ -529,14 +530,22 @@ if have_gumjs
...
+    not_found_message: 'please use v8 from https://github.com/frida/v8'
   )
-  cdata.set('HAVE_V8', v8_dep.found())
   if v8_dep.found()
+    assert(
+      cppc.has_header_symbol('v8-platform.h', 'v8::ThreadingBackend', dependencies: [v8_dep]),
+      'please use v8 from https://github.com/frida/v8'
+    )
+    # TODO use the actual name of the found v8_dep
     gumjs_extra_requires += 'v8-10.0'
```
milahu, line 546: currently frida-gum does not work with vanilla v8, so this can be ignored

oleavr: Right! I'm not sure there's much point in looking for `v8` though. But the `not_found_message` change alone does seem like a good idea.

milahu: fixed + check for v8::ThreadingBackend

oleavr (changes requested):
Thanks!

oleavr:
Had to revert this as it caused regressions, e.g. [macos-x86_64](https://github.com/frida/frida/actions/runs/4199451005/jobs/7284379145).

PR #727, opened 2023-02-27
gumjs: Swap interceptor and js locking when crossing scope 

mrmacete:
In this way it's possible to avoid a deadlock happening in Darwin due to `ThreadSuspendMonitor` potentially acquiring the JS lock during calls to `gum_interceptor_transaction_end` from threads without the JS lock held, in case RWX is not supported.

### DIFF
--- a/tests/gumjs/script.c
+++ b/tests/gumjs/script.c
@@ -569,6 +569,8 @@ static void on_outgoing_debug_message (const gchar * message,
 
 #ifdef HAVE_DARWIN
 static gpointer interceptor_attacher_worker (gpointer data);
+void empty_invocation_callback (GumInvocationContext * context,
+    gpointer user_data);
 #endif
 
 static int target_function_int (int arg);
@@ -7142,13 +7144,14 @@ interceptor_attacher_worker (gpointer data)
   int * state = data;
   guint i;
   GumInterceptor * interceptor;
-  TestCallbackListener * listener;
+  GumInvocationListener * listener;
   GumAttachReturn result;
 
   *state = 1;
 
   interceptor = gum_interceptor_obtain ();
-  listener = test_callback_listener_new ();
+  listener = gum_make_call_listener (empty_invocation_callback,
+      empty_invocation_callback, NULL, NULL);
 
   for (i = 0; i != 300; i++)
   {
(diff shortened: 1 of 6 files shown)


tests/gumjs/script.c
```diff
@@ -7083,6 +7090,86 @@ TESTCASE (interceptor_should_refuse_to_attach_without_any_callbacks)
...
+  GumAttachReturn result;
+
+  *state = 1;
+
+  interceptor = gum_interceptor_obtain ();
+  listener = test_callback_listener_new ();
+
+  for (i = 0; i != 300; i++)
+  {
+    result = gum_interceptor_attach (interceptor, target_function_int,
```
oleavr, line 7155: Would be good to use `gum_make_call_listener ()` here if possible.

oleavr (changes requested):
Thanks, great catch! 🔥

tests/gumjs/script.c
mrmacete: yes! way better!! thanks

tests/gumjs/script.c
```diff
@@ -564,6 +567,12 @@ static void on_incoming_debug_message (GumInspectorServer * server,
 static void on_outgoing_debug_message (const gchar * message,
     gpointer user_data);
 
+#ifdef HAVE_DARWIN
+static gpointer interceptor_attacher_worker (gpointer data);
+void empty_invocation_callback (GumInvocationContext * context,
```
oleavr, line 572: This should be static.

tests/gumjs/script.c
```diff
@@ -7083,6 +7092,93 @@ TESTCASE (interceptor_should_refuse_to_attach_without_any_callbacks)
...
+      "  'int', ['int'], { scheduling: 'exclusive' }"
+      ");"
+      "Interceptor.replace(threadSuspend, new NativeCallback((threadId) => {"
+      "  return threadSuspend(threadId);"
+      "}, 'int', ['int']));"
+      "Interceptor.flush();"
+      "setTimeout(() => {"
+      "  for (let i = 0; i !== iterations; i++) {"
+      "    Thread.sleep(0.1);"
+      "  }"
```
oleavr, line 7127:
```suggestion
      "  for (let i = 0; i !== iterations; i++)"
      "    Thread.sleep(0.1);"
```

oleavr (changes requested):

PR #730, opened 2023-03-04
darwin: Fix mapping zero-sized segments

comex:
mach_vm_remap returns KERN_INVALID_ARGUMENT when given a size of 0.  Fix by skipping these segments.


oleavr:
Thanks! 🙌🏻

PR #734, opened 2023-04-18
interceptor: Keep thread port alive between suspend and resume on Darwin

mrmacete:
In this way no thread can be left suspended, regardless of the initial ref count of the send right to its port.


oleavr (approved):
Whoah, nice catch! 🤘 Thanks! 🎉

PR #737, opened 2023-05-03
gum-graft: Add support for chained imports

mrmacete:
Collect graft imports from chained fixups too if present.


mrmacete:
there's still a bunch of CI jobs failing but now doesn't look like it's because of this change

gum/gumdarwingrafter.c
```diff
@@ -666,6 +671,222 @@ gum_darwin_grafter_compute_layout (GumDarwinGrafter * self,
...
+        if (seg_offset == 0)
+          continue;
+
+        seg_starts = (const GumChainedStartsInSegment *)
+            ((const guint8 *) image_starts + seg_offset);
+        format = seg_starts->pointer_format;
+
+        segment = gum_find_segment_by_offset (self,
+            seg_starts->segment_offset);
+
```
oleavr, line 740:
```suggestion
```

gum/gumdarwingrafter.c
```diff
@@ -666,6 +671,222 @@ gum_darwin_grafter_compute_layout (GumDarwinGrafter * self,
...
+          start = seg_starts->page_start[page_index];
+          if (start == GUM_CHAINED_PTR_START_NONE)
+            continue;
+
+          cursor = (const guint8 *) mach_header + seg_starts->segment_offset +
+              (page_index * seg_starts->page_size) +
+              start;
+
+          if (format == GUM_CHAINED_PTR_64 ||
+                format == GUM_CHAINED_PTR_64_OFFSET)
```
oleavr, line 758: It looks like the second line here should line up with the start of `format` above -- i.e. a hanging indent of 4.

gum/gumdarwingrafter.c
```diff
@@ -666,6 +671,222 @@ gum_darwin_grafter_compute_layout (GumDarwinGrafter * self,
   }
 }
 
+static void
+gum_collect_chained_imports (GumDarwinModule * self,
```
oleavr, line 675: Since this isn't a method of the `DarwinModule` class, we shouldn't name it `self`.

```suggestion
gum_collect_chained_imports (GumDarwinModule * module,
```

(Or `mod` in case it requires breaking lines like this.)

gum/gumdarwingrafter.c
```diff
@@ -666,6 +671,222 @@ gum_darwin_grafter_compute_layout (GumDarwinGrafter * self,
...
+        }
+      }
+    }
+
+    command = (const guint8 *) command + lc->cmdsize;
+  }
+}
+
+static const GumDarwinSegment *
+gum_find_segment_by_offset (GumDarwinModule * self,
```
oleavr, line 871: Since this isn't a method of the `DarwinModule` class, we shouldn't name it `self`.

oleavr (changes requested):
Awesome! :boom:

PR #745, opened 2023-07-26
Improve worker

mrmacete:
This change contains two minor fixes:

- Fix typo in worker.js `_dispatchMessage()`
  - so that `console.log` works from within the worker
- Pass `program` to worker's core init
  - so that js exceptions can be generated without segfaulting


oleavr:
Thanks! 🤘 (Just rebased to tweak the commit messages.)

PR #747, opened 2023-08-12
Custom-length backtraces

davinci-tech:
This commit attempts to add a new parameter to the ``Thread.backtrace()`` method called ``limit`` which under-the-hood uses the gum_backtracer_generate_with_limit method to create a custom-length backtrace. This change affects the gumjs bindings for both the v8 and quickjs runtimes.


oleavr:
Thanks a lot! (And apologies for the delay.)

I've applied some style tweaks and started tinkering with a new API for `Thread.backtrace()`, where the second argument is an object with options. In this way we can add further options, like this one, and the code is a bit easier to read.

This feature also needs a test, so will try to do that early next week, and get this landed.

davinci-tech:
This is amazing! Thank you very much!

PR #751, opened 2023-09-25
gumjs: Avoid segfault in file and database when Interceptor is not there

mrmacete:
This fixes the use case where `SQLiteDatabase` and `File` apis are used from within a Worker.

### DIFF
--- a/bindings/gumjs/gumquickmacros.h
+++ b/bindings/gumjs/gumquickmacros.h
@@ -193,7 +193,7 @@
       if (core->interceptor != NULL) \
       { \
         gum_interceptor_ignore_current_thread ( \
-              core->interceptor->interceptor); \
+            core->interceptor->interceptor); \
       } \
     }
 #define GUMJS_INTERCEPTOR_UNIGNORE() \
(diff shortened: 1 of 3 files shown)


bindings/gumjs/gumquickmacros.h
```diff
@@ -188,5 +188,21 @@
               GumQuickArgs * args, \
               int flags, \
               GumQuickCore * core)
+#define GUMJS_INTERCEPTOR_IGNORE() \
+    { \
+      if (core->interceptor != NULL) \
+      { \
+        gum_interceptor_ignore_current_thread ( \
+              core->interceptor->interceptor); \
```
oleavr, line 196: Hanging indent should be 4 spaces.

bindings/gumjs/gumquickmacros.h
```diff
@@ -188,5 +188,21 @@
...
+      } \
+    }
+#define GUMJS_INTERCEPTOR_UNIGNORE() \
+    { \
+      if (core->interceptor != NULL) \
+      { \
+        gum_interceptor_unignore_current_thread ( \
+                core->interceptor->interceptor); \
```
oleavr, line 204: Hanging indent should be 4 spaces.

oleavr (changes requested):
Yay, nice! 🤘

PR #752, opened 2023-10-09
Fix NativePointer from number for 32-bit BE.

forky2:
Fix new NativePointer(number) for 32-bit BE architectures (e.g. ARM BE8).

This union which overlaps gpointer and int64_t fails to allow extraction of
gpointer when architecture is 32-bit big endian. For that combination padding
is required, which this patch adds only for that combination.

### DIFF
--- a/bindings/gumjs/gumquickvalue.c
+++ b/bindings/gumjs/gumquickvalue.c
@@ -1370,11 +1370,12 @@ _gum_quick_native_pointer_parse (JSContext * ctx,
 #else
     union
     {
-        struct {
-            gpointer _pad;
-            gpointer p;
-        };
-        int64_t i;
+      struct
+      {
+        gpointer _pad;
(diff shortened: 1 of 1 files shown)


bindings/gumjs/gumquickvalue.c
```diff
@@ -1361,11 +1361,22 @@ _gum_quick_native_pointer_parse (JSContext * ctx,
...
       int64_t i;
     } v;
+#else
+    union
+    {
+        struct {
+            gpointer _pad;
+            gpointer p;
+        };
+        int64_t i;
```
oleavr, line 1377:
```suggestion
      struct
      {
        gpointer _pad;
        gpointer p;
      };
      int64_t i;
```

oleavr (approved):
Thanks, great catch! 🙌

PR #758, opened 2023-11-23
process: add get_main_module()

mrmacete:
This returns a copy of the details of the module which represents the main executable of the process.

It is exposed to JS via the `Process.mainModule` property (where it's cached because the main module is constant during the process' lifetime).

For darwin this is implemented by getting the first loaded module for which the Mach-O header's `filetype` is set to `MH_EXECUTE`. For all other archs it is currently just the first loaded module (equivalent to `Process.enumerateModules()[0]`).

### DIFF
--- a/bindings/gumjs/gumquickprocess.c
+++ b/bindings/gumjs/gumquickprocess.c
@@ -110,7 +110,7 @@ static void gum_quick_exception_handler_free (
 static gboolean gum_quick_exception_handler_on_exception (
     GumExceptionDetails * details, GumQuickExceptionHandler * handler);
 
-static void gumjs_free_main_module (GumQuickProcess * self);
+static void gumjs_free_main_module_value (GumQuickProcess * self);
 
 static const JSCFunctionListEntry gumjs_process_entries[] =
 {
@@ -146,7 +146,7 @@ _gum_quick_process_init (GumQuickProcess * self,
 
   self->module = module;
   self->core = core;
-  self->main_module = JS_UNINITIALIZED;
+  self->main_module_value = JS_UNINITIALIZED;
 
   _gum_quick_core_store_module_data (core, "process", self);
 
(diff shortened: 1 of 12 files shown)


gum/gumprocess.h
```diff
@@ -209,6 +209,7 @@ GUM_API gboolean gum_process_resolve_module_pointer (gconstpointer ptr,
     gchar ** path, GumMemoryRange * range);
 GUM_API void gum_process_enumerate_modules (GumFoundModuleFunc func,
     gpointer user_data);
+GUM_API GumModuleDetails * gum_process_get_main_module (void);
```
oleavr, line 212:
```suggestion
GUM_API const GumModuleDetails * gum_process_get_main_module (void);
```

- Since the main module won't change, it would be good to return a borrowed pointer to the details. We can use the `g_once*()` API to do this lazily and thread-safe.
- Nit: Perhaps we should move it up, just before `resolve_module_pointer()`, as it's the "main" thing, and looking at other modules is going further into the details.

bindings/gumjs/gumquickprocess.c
```diff
@@ -182,6 +186,18 @@ gumjs_get_parent_module (GumQuickCore * core)
...
+GUMJS_DEFINE_GETTER (gumjs_process_get_main_module)
+{
+  GumQuickProcess * self;
+
+  self = gumjs_get_parent_module (core);
+
+  if (self->main_module == NULL)
+    self->main_module = gum_process_get_main_module ();
+
+  return _gum_quick_module_new (ctx, self->main_module, self->module);
```
oleavr, line 198: Perhaps we should cache the JS wrapper?

oleavr (changes requested):
Yay, this is awesome! :metal:

gum/gumprocess.h
mrmacete: done

bindings/gumjs/gumquickprocess.c
mrmacete: done for quickjs, not super sure how to do it in v8, is there any example around?

oleavr: Yay! For V8, there's an example where we store it [here](https://github.com/frida/frida-gum/blob/ed4f17176f8671af93fbbb9afdbb584fb7825bf6/bindings/gumjs/gumv8interceptor.cpp#L394-L395), and then [use it](https://github.com/frida/frida-gum/blob/ed4f17176f8671af93fbbb9afdbb584fb7825bf6/bindings/gumjs/gumv8interceptor.cpp#L633-L634) at a later point.

PR #765, opened 2023-12-19
Add JS bindings for fast interceptor

WorksButNotTested:


bindings/gumjs/gumquickinterceptor.c
```diff
@@ -719,6 +721,75 @@ gum_quick_replace_entry_free (GumQuickReplaceEntry * entry)
...
+
+    goto propagate_exception;
+  }
+propagate_exception:
+  {
+    gum_quick_replace_entry_free (entry);
+
+    return JS_EXCEPTION;
+  }
+}
```
oleavr, line 791: There's quite some duplicated logic here (and in the V8 counterpart) -- would be good to refactor this. Perhaps `gum_quick_replace_entry_new()`, a function to "register" the entry (that does the counter bump + hash table insertion), and a helper to throw based on the `GumReplaceReturn`.

oleavr (changes requested):
Yay, thanks for doing this! 🙌

bindings/gumjs/gumquickinterceptor.c
WorksButNotTested: Refactored, also for v8.

PR #769, opened 2024-01-10
stalker: copy register branch instructions on arm64

mrmacete:
Instead of replacing them with functionally-equivalent ones, so that any pointer authentication context is used as expected.


oleavr:
Yay! 🚀 Thanks! 🙌

PR #771, opened 2024-01-15
Add Cloak module

mrmacete:
This change adds the `Cloak` module, exposing the add / remove / query apis from `gumcloak.h`. This module is exposed to the Worker too.

There are 2 more loosely related changes:
- add a missing call to `gum_cloak_add_range` in Interceptor to cloak the thunks on arm64
- fix the docs of `gum_cloak_clip_range` to describe the actual semantics of the return value

### DIFF
--- a/bindings/gumjs/gumquickcloak.c
+++ b/bindings/gumjs/gumquickcloak.c
@@ -1,0 +1,217 @@
+/*
+ * Copyright (C) 2024 Francesco Tamagni <mrmacete@protonmail.ch>
+ *
+ * Licence: wxWindows Library Licence, Version 3.1
+ */
+
+#include "gumquickcloak.h"
+
+#include "gumquickmacros.h"
+
+#include <gum/gumcloak.h>
+
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_add_thread)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_remove_thread)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_has_thread)
+
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_add_range)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_remove_range)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_has_range_containing)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_clip_range)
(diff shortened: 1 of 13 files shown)


bindings/gumjs/gumquickcloak.c
```diff
@@ -0,0 +1,218 @@
...
+  return JS_NewBool (ctx, gum_cloak_has_thread (thread_id));
+}
+
+GUMJS_DEFINE_FUNCTION (gumjs_cloak_add_range)
+{
+  gpointer address;
+  gsize size;
+  GumMemoryRange range;
+
+  if (!_gum_quick_args_parse (args, "pZ", &address, &size))
```
oleavr, line 111: Would be good to use a range object at the public API level, for `addRange()` and `removeRange()`, like we do in `Stalker.exclude()`. The C/C++ code can be kept as is, and like `Stalker.exclude()` we can unwrap the object in `runtime/core.js` and call the underscore-prefixed implementation.

bindings/gumjs/gumquickcloak.c
```diff
@@ -0,0 +1,218 @@
...
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_add_range)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_remove_range)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_has_range_containing)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_clip_range)
+
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_add_file_descriptor)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_remove_file_descriptor)
+GUMJS_DECLARE_FUNCTION (gumjs_cloak_has_file_descriptor)
+
+
```
oleavr, line 26: Extra blank line.

oleavr (changes requested):
w00t! :metal:

bindings/gumjs/gumquickcloak.c
mrmacete: done!

bindings/gumjs/gumquickcloak.c
mrmacete: done! thanks

mrmacete:
lol i broke tests again, checking

PR #772, opened 2024-01-15
interceptor: Check lib prefix anyways in claiming grafts on arm64

mrmacete:
This change is needed because since iOS 17, Xcode can load the `/private/preboot/Cryptexes/OS/usr/lib/libLogRedirect.dylib` library (there’s also a macOS counterpart:
`/System/Cryptexes/OS/usr/lib/libLogRedirect.dylib`) which uses interposing for wrapping a bunch of commonly used (and commonly hooked) symbols.

Since those libraries are not in the dyld cache, Frida was flagging them as non-system libraries, making Interceptor fail loudly (instead of silently) if no direct import grafts were found in the loaded binaries for those symbols.

This change adds a check for `/private/preboot` prefix, and now the prefix check is in OR with the dyld cache one instead of being mutually exclusive.

### DIFF
--- a/gum/backend-arm64/guminterceptor-arm64.c
+++ b/gum/backend-arm64/guminterceptor-arm64.c
@@ -586,22 +586,28 @@ gum_compare_grafted_hook (const void * element_a,
 static gboolean
 gum_is_system_module (const gchar * path)
 {
-  static gboolean initialized = FALSE;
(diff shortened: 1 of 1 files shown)


oleavr:
Nice catch!! Thanks! :raised_hands:

PR #784, opened 2024-02-22
Fix: Stalker disrupts exclusive access by emitting a block event.

saicao:
Stalker generates the block event even when those blocks have exclusive access. This may disrupt the exclusive access when a user handles the block event.
It seems that the default stalker event sinker uses spinlocks, which can starve the following thread.

[check this issue](https://github.com/frida/frida/issues/2769)


oleavr:
Thanks, great catch! 🙌🏻

PR #787, opened 2024-03-28
Expose Cloak and Interceptor locks for usage in ThreadSuspendMonitor

mrmacete:
This also adds `gum_spinlock_try_acquire` which has the same semantics as `g_rec_mutex_trylock`.


oleavr:
Thanks!! 🙌

Issue #789, opened 2024-04-01
NativeCallback unable to leave thread's last error unchanged


HexKitchen:
Calling a JavaScript function wrapped as NativeCallback unexpectedly forces the current thread's error status to 0.

JavaScript code can make a change to the thread's error status by assigning to `this.errno` / `this.lastError`, and that works correctly. But if the script wishes instead to preserve the original value, it does not work as expected. Instead, the error status is forced to 0. Even if the script expressly sets `this.errno` / `this.lastError` to the original value, the error status gets set to 0 instead.

Example:

```python
import frida

pid = frida.spawn("test.exe")
session = frida.attach(pid)

script = session.create_script("""
function myCallback() {
    //console.log("Entering myCallback. this.lastError: " + this.lastError);
    //this.lastError = 1337;   // [1]
}

Interceptor.replace(DebugSymbol.fromName("func1").address, new NativeCallback(myCallback, 'void', []))
""")
script.load()

frida.resume(pid)

input()
```

```c
#include <iostream>
#include <Windows.h>

#pragma auto_inline(off)

extern "C" {
    void func1()
    {
        std::cout << "In func1" << std::endl;
    }
}


int main()
{
    SetLastError(1337);

    std::cout << "Last error: " << GetLastError() << std::endl;
    func1();
    std::cout << "Last error: " << GetLastError() << std::endl;
    func1();
    std::cout << "Last error: " << GetLastError() << std::endl;
}
```

Expected output:
```
Last error: 1337
Last error: 1337
Last error: 1337
```

Actual output:
```
Last error: 1337
Last error: 0
Last error: 0
```

Uncommenting the line marked `[1]` produces these results:
```
Last error: 1337
Last error: 0
Last error: 1337
```

Related issue: #405 .

PR #790, opened 2024-04-01
Fix: preserved thread's system error over NativeCallback invocations

HexKitchen:
Fix for #789 : NativeCallback unable to leave thread's last error unchanged 

Though JavaScript function wrapped as a NativeCallback allows both inspecting and altering the current thread's system error, it wasn't possible to leave the system error as-is. Unless replaced by a value unequal to the original, the system error was being forced to 0 upon return from the NativeCallback. Fixed and tests added.


oleavr:
Thanks, great catch! 🙌

PR #795, opened 2024-05-02
Exposed memory_read , memory_write and memory_scan gum functions

hillelpinto:
I've referenced some memory functions (read, write & scan) to the cmodule runtime so that it can be invoked through cmodule scripting


jevinskie:
These need to go inside the `#endif`. Can you also please add tests? Something simple like reading from a stack allocated local variable for the simple case an more complicated case where it tries to read memory that `memcpy` would `SEGFAULT` on while `gum_memory_*` would handle gracefully with either no read bytes or partial read bytes?

@oleavr anything else you’d like to see added from `gummemory.h` while we’re at it? Perhaps some of the scan functionality would be useful?

hillelpinto:
Hey, i updated the code and for the tests yeah of course, where to add them ?

oleavr (changes requested):
Thanks! Could you rebase this on latest main? Just from a quick look it seems the added code references types that aren't present, so I suspect this will break the CModule runtime.

oleavr:
> @oleavr anything else you’d like to see added from `gummemory.h` while we’re at it? Perhaps some of the fancier scan functionalities (masked, etc) would be useful?

I tend to add things as I find myself needing them, but yeah those sound good to me.

oleavr:
(Sorry it took me so long to review -- I was stuck in a rabbit-hole, and have now finally made it out.)

oleavr:
Note that tests should be super-basic -- we already have test coverage for these primitives in `tests/core/`, so the tests only need to ensure that the symbols exist and one simple use-case works. (So we know that the types match the full headers in `gum/`.)

hillelpinto:
@oleavr Code is rebased and test added

oleavr:
Thanks! :raised_hands: Since it took me some time to get to this, I've improved the test to test things inside CModule, and exposed some more APIs while at it. Cheers!

PR #802, opened 2024-05-23
exceptor: Add SA_ONSTACK flag the POSIX backend

asabil:
Add the SA_ONSTACK flag to handle situations where and alternative signal stack has been configured using `sigaltstack`. In the situations where no alternative stack was configured, this behaves as if the flag has not been set.


asabil:
Looks like older QNX versions don't support SA_ONSTACK?

oleavr:
Thanks! 🙌

PR #804, opened 2024-06-09
cpu-features: Ignore CET SS unless actively used

yjugl:
Since #791 we are ensuring a proper call-ret discipline in the x86 interceptor, if we detect that the CPU is compatible with Intel CET shadow stacks. As discussed in #791, this has an unnecessary performance cost if the mitigation is not used by the current process. This extra patch thus ignores CET shadow stacks compatibility on Windows if we detect that the current process is not using the mitigation. It might be cleaner to move this code outside `gum_do_query_cpu_features`, but putting it there makes the patch very simple and atomic.

### DIFF
--- a/gum/gum.c
+++ b/gum/gum.c
@@ -732,6 +732,9 @@ gum_do_query_cpu_features (void)
   gboolean cpu_supports_cet_ss = FALSE;
   gboolean os_enabled_xsave = FALSE;
   guint a, b, c, d;
+#ifdef HAVE_WINDOWS
+  PROCESS_MITIGATION_USER_SHADOW_STACK_POLICY pol;
+#endif
 
   if (gum_get_cpuid (7, &a, &b, &c, &d))
   {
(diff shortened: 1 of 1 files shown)


yjugl:
MinGW builds are failing not finding `ProcessUserShadowStackPolicy` and `PROCESS_MITIGATION_USER_SHADOW_STACK_POLICY`, althouh these two symbols were pushed to mingw-w64 in July 2021 ([integration](https://sourceforge.net/p/mingw-w64/mailman/message/37314758/), [original message](https://sourceforge.net/p/mingw-w64/mailman/message/37314403/)). Could it be that the CI is using an outdated version of mingw-w64? Do you have control over that?

oleavr (changes requested):
Thanks!! (And apologies for the delay.)

We are in control of the MinGW bits, though we're currently piggybacking on the pre-installed components on GitHub's runners to speed up CI, but we could always have it update the components as an up-front step.

However, we still want to retain support for XP -- due to users reversing software on legacy systems -- so we should resolve the function dynamically to retain backwards compatibility.

PR #808, opened 2024-07-12
Fix #807: recv wait can deadlock on an application thread

### LINKED ISSUE #807: recv wait can deadlock on an application thread
### Summary

There is a race condition that can cause `recv().wait()` on an application thread to deadlock.

### Repro steps

For a fairly minimal repro case, see the attached files, which are adapted from the "Blocking recv" example at https://frida.re/docs/messages/ . Running it should cause a deadlock after roughly 1 second.

Attachment: [test1.zip](https://github.com/user-attachments/files/16189017/test1.zip)

I have also seen the same deadlock occur when running the testcase `/GumJS/Script/recv_wait_should_not_leak#V8`.

See also the pull request I am creating to fix this race condition, which includes its own specialized testcase.

If you have any difficulty reproducing it, I'll be glad to assist further.

### Some diagnosis

The implementation of `wait()` starts in `message-dispatcher.js`:

```
  this.wait = function wait() {
    while (!completed)                    // <=== [LINE 1]
      engine._waitForEvent();
  };
```

If at `[LINE 1]` the requested message has already been received, `wait()` will return without needing to call down into `engine._waitForEvent`. Everything is fine in this circumstance.

If the requested message has not yet been received, execution continues in `gumjs_wait_for_event` (code shown for QJS):

```
...
  g_mutex_lock (&self->event_mutex);     // <=== [LINE 2]

  start_count = self->event_count;       // <=== [LINE 3]
  while (self->event_count == start_count && self->event_source_available)
  {
    ...
      g_cond_wait (&self->event_cond, &self->event_mutex);
    ...
  }
```

At `[LINE 2]` the code obtains `event_mutex`, thereby locking in the value of `core->event_count`. Then it waits on `event_cond`. Subsequently, when an event posts, it will wake this thread. Everything is fine in this circumstance as well.

The trouble happens if the event posts after the check on `[LINE 1]` but before the mutex is obtained on `[LINE 2]`. In that case, the `start_count` observed on `[LINE 3]` winds up being the value of `event_count` after the event posts. Furthermore, `g_cond_wait` will hang, potentially forever, unless by luck some additional event (that should not have a bearing on this flow) happens to post.

HexKitchen:
This is a fix for #807 : recv wait can deadlock on an application thread.

See the issue description for a discussion of the root cause.

The proposed fix here operates by introducing a thread-local variable, `event_count_last_seen`. After entering the mutex, which ensures a stable reading of `core->event_count`, the thread enters the wait for `event_cond` only if `core->event_count` is equal to `event_count_last_seen`. In this way, we can guarantee that we won't begin waiting on `event_cond` at a time when the event we need has already been broadcast, which would produce the deadlock.

### DIFF
--- a/bindings/gumjs/gumquickcore.c
+++ b/bindings/gumjs/gumquickcore.c
@@ -3,6 +3,7 @@
  * Copyright (C) 2020-2022 Francesco Tamagni <mrmacete@protonmail.ch>
  * Copyright (C) 2020 Marcus Mengs <mame8282@googlemail.com>
  * Copyright (C) 2021 Abdelrahman Eid <hot3eed@gmail.com>
+ * Copyright (C) 2024 Simon Zuckerbraun <Simon_Zuckerbraun@trendmicro.com>
  *
  * Licence: wxWindows Library Licence, Version 3.1
  */
--- a/bindings/gumjs/gumquickcore.c
+++ b/bindings/gumjs/gumquickcore.c
@@ -2735,7 +2735,8 @@ GUMJS_DEFINE_FUNCTION (gumjs_wait_for_event)
 
   g_mutex_lock (&self->event_mutex);
 
-  while (self->event_count == thread_data->event_count_last_seen && self->event_source_available)
+  while (self->event_count == thread_data->event_count_last_seen
+      && self->event_source_available)
   {
     if (called_from_js_thread)
     {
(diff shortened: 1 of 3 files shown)


oleavr:
Thanks! Great catch. Will try to get this landed on Monday, in time for 16.4.4.

oleavr:
Just pushed some changes to try a slightly simpler approach, without thread-local storage. Also pushed some style fixes/tweaks. I'll have to get some sleep now, so I'll land this first thing tomorrow -- but in case you get a chance to test in the meantime please let me know how it goes :)

oleavr (approved):

HexKitchen:
Super. Agreed this is simpler and solves the deadlock, I was able to confirm in testing as well. Thx!

PR #809, opened 2024-07-14
darwin: Fix sysroot on simulator

ChiChou:
The patch for https://github.com/frida/frida/issues/2763

frida has stoppped working on iOS simulators for months, getting `module not found at "/usr/lib/libSystem.B.dylib"` error. The root cause is the logic for getting sysroot stops working.

Had some discussion with @mrmacete yesterday at 0x41con. We talked about reading env remotely, but surprisingly found the current implementation has a cheat to simply detect the location of `dyld_sim`. However the assumtion that `dyld_sim` is the first loaded image is no longer true. We need one more loop to initialize sysroot before putting them to the lookup table.

Also I believe this is the correct patch for https://github.com/frida/frida-swift-bridge/pull/18

Kudos to Francesco for helping me to set up the latest env and understand the source! 🙌

### DIFF
--- a/gum/backend-darwin/gumdarwinmoduleresolver.c
+++ b/gum/backend-darwin/gumdarwinmoduleresolver.c
@@ -35,6 +35,8 @@ static void gum_darwin_module_resolver_get_property (GObject * object,
 static void gum_darwin_module_resolver_set_property (GObject * object,
     guint property_id, const GValue * value, GParamSpec * pspec);
 
+static gboolean gum_find_sysroot (const GumModuleDetails * details,
+    gpointer user_data);
 static gboolean gum_store_module (const GumModuleDetails * details,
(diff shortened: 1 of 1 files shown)


ChiChou:
This pull request has no test code yet. I wonder if it's possible for the CI to run the simulator..

ChiChou:
Note that the module lookup logic is still problematic under other cases. Here are 2 examples:

1. On macOS, turn off SIP and attach to WhatsApp
2. Some jailbreak users reported that it happens to their iOS too https://github.com/frida/frida/issues/2763#issuecomment-2186212537

They both throw the following exception, but the root cause should be different.

`module not found at "/usr/lib/libSystem.B.dylib"`

oleavr:
Ooh, great catch! Thanks! 🙌

PR #810, opened 2024-07-18
Ensure gum is itself an input to the gumjs devkit

hexploitable:
This addresses a bug that _can_ result in gumjs being built before gum, and thus gumjs being incomplete (missing symbols).
This ensures that when building the gumjs devkit, gum is treated as an input.


oleavr:
Thanks! 🙌

PR #814, opened 2024-08-06
Fix TLV initialization on AArch64

jiska2342:
Fix a number of issues introduced in #709. Injecting a library using TLVs now works on iOS.

Tested on an iPhone 14 running iOS 16.2.1 with Dopamine.


oleavr:
Thanks, this is awesome! 🤩

PR #820, opened 2024-09-25
Optional support for bitcode, symbol prefixing, and cmodule.

hexploitable:
This PR adds 3 new features/options to the build. 

1. **CModule is now a feature** which can be enabled or disabled. It is enabled by default.
2. **Symbol prefixing** is now a feature which can be disabled. Largely due to this causing issues when embedded bitcode (symbols are non-renamed inside the bitcode and therefore the linker gets confused).
3. **Embed bitcode** - support for including bitcode in the builds.

Depends on an update to releng: https://github.com/frida/releng/pull/4

### DIFF
--- a/subprojects/capstone.wrap
+++ b/subprojects/capstone.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/frida/capstone.git
-revision = e98746112da0a40b2ccd0340db0d20cca5f97950
+revision = 27fb5a53bc77baf3c3af7ccd466041418cecca35
 depth = 1
 
 [provide]
--- a/subprojects/glib.wrap
+++ b/subprojects/glib.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/frida/glib.git
-revision = f8e0cd4e9e671a15a69a7323f421f1e8df5c7a10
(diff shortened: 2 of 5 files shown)


meson.build
```diff
@@ -4,10 +4,18 @@ project('frida-gum', 'c', 'cpp',
     check: true).stdout().strip(),
   meson_version: '>=1.1.0',
   default_options: [
-    'c_std=gnu99,c99',
+    'c_std=gnu99,c99'
   ],
 )
 
+enable_bitcode = get_option('bitcode')
```
oleavr, line 11: Meson already has a built-in option for this: `b_bitcode`.

meson.build
```diff
@@ -428,12 +436,18 @@ if get_option('devkits').length() != 0
   endif
   mkdevkit = [python, releng / 'mkdevkit.py']
   uninstalled_dir = meson.global_build_root() / 'meson-uninstalled'
+
+  prefix_syms = get_option('prefixsyms')
```
oleavr, line 440: It would be good to invert the logic, so it's opt-out instead. In that way we won't have to update frida-core, which also uses `mkdevkit.py`.

oleavr (changes requested):
Thanks! So sorry it took me ages to get to this :sweat:

PR #826, opened 2024-09-30
guminterceptor: add alt stack support

ajwerner:
This commit adds a new public API to gum to configure the size of an alternate stack to use for hooking. If the configuration is set, then when an interceptor is triggered, it will switch to this stack after the trampoline but before executing the handler. This is needed to support go or programs which do not have large and mutable thread stacks.

One note is that only x86_64 support has initially been added, and that it only applies for entry invocation listeners. Exit invocation listeners are not supported because they still would not work with go out-of-the-box because go doesn't like junk being added to call stacks.

### DIFF
--- a/gum/guminterceptor.c
+++ b/gum/guminterceptor.c
@@ -27,6 +27,10 @@
 # define GUM_INTERCEPTOR_CODE_SLICE_SIZE 256
 #endif
 
+#if defined(__x86_64__)
+# define GUM_INTERCEPTOR_ALT_STACK_SUPPORTED
+#endif
+
 #define GUM_INTERCEPTOR_LOCK(o) g_rec_mutex_lock (&(o)->mutex)
 #define GUM_INTERCEPTOR_UNLOCK(o) g_rec_mutex_unlock (&(o)->mutex)
 
@@ -115,6 +119,11 @@ struct _InterceptorThreadContext
   GumInvocationStack * stack;
 
   GArray * listener_data_slots;
+
+#if defined(GUM_INTERCEPTOR_ALT_STACK_SUPPORTED)
+  gpointer alt_stack;
+  size_t alt_stack_size;
+#endif
 };
 
 struct _GumInvocationStackEntry
@@ -192,6 +201,8 @@ static void gum_function_context_add_listener (
 static void gum_function_context_remove_listener (
     GumFunctionContext * function_ctx, GumInvocationListener * listener);
 static void listener_entry_free (ListenerEntry * entry);
+static void listener_entry_on_enter ( ListenerEntry * listener_entry,
+    GumInvocationContext * invocation_ctx);
 static gboolean gum_function_context_has_listener (
     GumFunctionContext * function_ctx, GumInvocationListener * listener);
 static ListenerEntry ** gum_function_context_find_listener (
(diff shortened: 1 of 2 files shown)


ajwerner:
One question I have is how best to test this in CI. Locally I changed the default value the stack size to enable the alternate stack switching and I observed the tests passing. Ideally we could run the interceptor suite twice -- once on the local stack and once on the switched stack. Any pointers on how best to do this would be great.

gum/guminterceptor.c
```diff
@@ -1520,8 +1561,76 @@ _gum_function_context_begin_invocation (GumFunctionContext * function_ctx,
...
+  // Check if the stack pointer is within the alt stack range already, and if it is,
+  // we don't need to switch stacks.
+  if (cpu_context->rsp >= (uintptr_t)interceptor_ctx->alt_stack &&
+      cpu_context->rsp < (uintptr_t)new_stack_top)
+    goto no_alt_stack;
+
+  // Switch to the new stack and call the function with arguments
+  asm volatile (
+      "movq %%rsp, %%rbx\n"      // Save the current stack pointer
+      "movq %0, %%rsp\n"         // Set the new stack pointer
```
oleavr, line 1589: I suppose this may be too late in some cases, as we've already used quite some stack space for saving registers?

gum/guminterceptor.c
```diff
@@ -1782,7 +1897,7 @@ gum_function_context_fixup_cpu_context (GumFunctionContext * function_ctx,
 }
 
 static InterceptorThreadContext *
-get_interceptor_thread_context (void)
```
oleavr, line 1785: This is incorrect; in C (unlike C++), `(void)` means "no arguments", whereas `()` means "0 or more arguments".

gum/guminterceptor.h
```diff
@@ -84,6 +84,13 @@ GUM_API void gum_interceptor_with_lock_held (GumInterceptor * self,
     GumInterceptorLockedFunc func, gpointer user_data);
 GUM_API gboolean gum_interceptor_is_locked (GumInterceptor * self);
 
+// Configure the alternate stack for the interceptor. The alternate stack is
```
oleavr, line 87: Style nitpick: Only C-style comments should be used.

gum/guminterceptor.c
```diff
@@ -1939,9 +2054,16 @@ interceptor_thread_context_new (void)
   context->stack = g_array_sized_new (FALSE, TRUE,
       sizeof (GumInvocationStackEntry), GUM_MAX_CALL_DEPTH);
 
+
```
oleavr, line 2057: Style nitpick: For consistency, two or more consecutive blank lines shouldn't be used.

gum/guminterceptor.c
```diff
@@ -1919,7 +2034,7 @@ gum_interceptor_replacement_invocation_backend =
 };
 
 static InterceptorThreadContext *
-interceptor_thread_context_new (void)
```
oleavr, line 1922: (See earlier comment.)

gum/guminterceptor.c
```diff
@@ -27,6 +27,10 @@
 # define GUM_INTERCEPTOR_CODE_SLICE_SIZE 256
 #endif
 
+#if defined(__x86_64__)
```
oleavr, line 30:
```suggestion
#if defined (HAVE_I386) && GLIB_SIZEOF_VOID_P == 8
```

(For consistency.)

Should also check that GCC or Clang is used, as MSVC doesn't support the inline assembly syntax.

oleavr (changes requested):
Thanks for taking a stab at this! This looks promising. (Apologies for the delay! :sweat_smile:)

PR #971, opened 2024-10-08
memory: Make regex patterns raw

mrmacete:
According to the docs of `g_regex_new` about the `G_REGEX_RAW` flag: "Usually strings must be valid UTF-8 strings, using this flag they are considered as a raw sequence of bytes."

The "strings" this refers to are the haystacks we then pass to `g_regex_match_full` when searching.

Without the flag, memory search with regex patterns can get interrupted before the range is over, when "invalid" bytes are encountered, resulting in false negatives.


oleavr:
Thanks! 🙌

PR #985, opened 2025-01-10
Aarch64 be patches

WorksButNotTested:


gum/arch-arm64/gumarm64writer.c
```diff
@@ -1992,15 +1992,21 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
...
 
+    /*
+     * Whilst instructions in aarch64 are always in little endian (even on 
+     * big-endian systems), the data is in native endian. Thus we don't need
+     * to make use of the GINT64_FROM_LE and GINT64_TO_LE when accessing the
+     * "slot" below.
+     */
     for (slot = first_slot; slot != last_slot; slot++)
     {
-      if (GINT64_FROM_LE (*slot) == r->val)
```
oleavr, line 1997: This is incorrect, as the writer could in theory be running on a big-endian system, generating code for an AArch64 system.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -1992,15 +1992,21 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
...
+      if (*slot == r->val)
         break;
     }
 
     if (slot == last_slot)
     {
-      *slot = GINT64_TO_LE (r->val);
```
oleavr, line 2003: (Like above.)

gum/arch-arm64/gumarm64writer.c
```diff
@@ -2024,15 +2030,21 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
...
 
+    /*
+     * Whilst instructions in aarch64 are always in little endian (even on 
+     * big-endian systems), the data is in native endian. Thus we don't need
+     * to make use of the GINT64_FROM_LE and GINT64_TO_LE when accessing the
+     * "slot" below.
+     */
     for (slot = first_slot; slot != last_slot; slot++)
     {
-      if (GINT32_FROM_LE (*slot) == r->val)
```
oleavr, line 2029: (Like above.)

gum/arch-arm64/gumarm64writer.c
```diff
@@ -2024,15 +2030,21 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
...
+      if (*slot == r->val)
         break;
     }
 
     if (slot == last_slot)
     {
-      *slot = GINT32_TO_LE (r->val);
```
oleavr, line 2035: (Like above.)

gum/gumdefs.h
```diff
@@ -119,9 +119,10 @@ typedef GumMipsCpuContext GumCpuContext;
  * The only non-legacy big-endian configuration on 32-bit ARM systems is BE8.
  * In this configuration, whilst the data is in big-endian, the code stream is
  * still in little-endian. Since Capstone is disassembling the code stream, it
- * should work in little-endian even on BE8 systems.
+ * should work in little-endian even on BE8 systems. On big-endian 64-bit ARM
+ * systems, the code stream is likewise in little-endian.
  */
-#if G_BYTE_ORDER == G_LITTLE_ENDIAN || defined (__arm__)
+#if G_BYTE_ORDER == G_LITTLE_ENDIAN || defined (__arm__) || defined (__aarch64__)
```
oleavr, line 125: Exceeds 80 cols.

oleavr (changes requested):
Thanks! :raised_hands:

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: I'm not sure I follow. Is it not the case that the writer should run only on the architecture it is generating code for (e.g. no cross-compiler like support is required)?

If I understand correctly, on aarch64_le, both code and data are little endian. But on aarch64_be, the code remains little endian whilst the data is big endian. So, in both cases when writing code (for example in `gum_arm64_writer_put_ldr_reg_value`) we should use the `GUINTXX_(TO/FROM)_LE` macros. However, on both platforms, data is in native endian and so no conversion should be required when writing `literals` since these are data.

oleavr: > I'm not sure I follow. Is it not the case that the writer should run only on the architecture it is generating code for (e.g. no cross-compiler like support is required)?

That's the main use-case, but we also support other use-cases such as:
- Offline processing of binaries, e.g. [gum-graft](https://github.com/frida/frida-gum/blob/8d9722a81d45699f165a4e90d4cd0e938b4cabac/tools/gumgraft.c), which pre-processes Mach-O binaries for iOS so they contain hooks that can be activated at runtime.
- Local generation of code to be inserted on a remote system, e.g. Frida's [Barebone](https://github.com/frida/frida-core/tree/main/src/barebone) backend.

> If I understand correctly, on aarch64_le, both code and data are little endian. But on aarch64_be, the code remains little endian whilst the data is big endian. So, in both cases when writing code (for example in `gum_arm64_writer_put_ldr_reg_value`) we should use the `GUINTXX_(TO/FROM)_LE` macros. However, on both platforms, data is in native endian and so no conversion should be required when writing `literals` since these are data.

I suppose we'll need a `data_endian` field or something like that, that defaults to the host machine's endian.

WorksButNotTested: Something like...

#ifndef __HOST_DATA_ENDIAN__
# define __HOST_DATA_ENDIAN__ __BYTE_ORDER__
#endif

gum/arch-arm64/gumarm64writer.c
```diff
@@ -1992,15 +1993,37 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
     if (r->width != GUM_LITERAL_64BIT)
       continue;
 
+    /*
+     * Whilst instructions in aarch64 are always in little endian (even on
+     * big-endian systems), the data is in native endian. Thus since we wish to
+     * support writing code for big-endian systems on little-endian targets and
+     * vice versa, we need to check the writer configuration.
+     */
```
oleavr, line 2001: Let's move this comment to where the `data_endian` field is defined. We can then avoid duplicating it further down.

gum/arch-arm64/gumarm64writer.h
```diff
@@ -51,6 +52,15 @@ struct _GumArm64Writer
   GumMetalArray label_refs;
   GumMetalArray literal_refs;
   const guint32 * earliest_literal_insn;
+
+  GumArm64DataEndian data_endian;
```
oleavr, line 56:
```suggestion
  gint data_endian;
```

- Let's avoid introducing a new enum here. Our Vala bindings will be using the `ByteOrder` enum from `glib-2.0.vapi`.
- This field should be moved up. The existing three configuration values are in their own section, starting with `target_os`. I'd say this one should go first, as it's more important/fundamental than `target_os`.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -220,6 +220,7 @@ gum_arm64_writer_init (GumArm64Writer * writer,
   writer->label_defs = NULL;
   writer->label_refs.data = NULL;
   writer->literal_refs.data = NULL;
+  writer->data_endian = GUM_ENDIAN_NATIVE;
```
oleavr, line 223: Assignment should go just before `target_os`.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -1992,15 +1993,37 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
     if (r->width != GUM_LITERAL_64BIT)
       continue;
 
+    /*
+     * Whilst instructions in aarch64 are always in little endian (even on
```
oleavr, line 1997:
```suggestion
     * Whilst instructions in AArch64 are always in little endian (even on
```

(Here and elsewhere.) For consistency with our existing comments.

gum/arch-arm64/gumarm64writer.c
```diff
@@ -1992,15 +1993,37 @@ gum_arm64_writer_commit_literals (GumArm64Writer * self)
...
+     * big-endian systems), the data is in native endian. Thus since we wish to
+     * support writing code for big-endian systems on little-endian targets and
+     * vice versa, we need to check the writer configuration.
+     */
     for (slot = first_slot; slot != last_slot; slot++)
     {
-      if (GINT64_FROM_LE (*slot) == r->val)
-        break;
+      if (self->data_endian == GUM_ENDIAN_LITTLE)
```
oleavr, line 2004: Is the optimizer able to avoid repeatedly reading and checking this field's value, or should we introduce a local variable used throughout the function?

oleavr (changes requested):

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: Done.

gum/arch-arm64/gumarm64writer.h
WorksButNotTested: Done. I kept the `typedef guint GumArm64DataEndian;`, was that right? Also added a comment to it for what the values should be (since the type is otherwise not very constrained).

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: Done.

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: Done

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: I would be very surprised if the optimizer didn't take care of it anyway, but the actual code generation part isn't hugely performance critical (to the point where a few cycles matter), its the performance of the generated code which is most important (for applications such as fuzzing).

gum/arch-arm64/gumarm64writer.c
```diff
@@ -213,6 +213,7 @@ gum_arm64_writer_init (GumArm64Writer * writer,
   writer->ref_count = 1;
   writer->flush_on_destroy = TRUE;
 
+  writer->data_endian = __BYTE_ORDER__;
```
oleavr, line 216: Should use `G_BYTE_ORDER` and friends here and elsewhere.

oleavr (changes requested):

gum/arch-arm64/gumarm64writer.c
WorksButNotTested: Fixed. Apologies for the delay.

oleavr:
Thanks, merged! :raised_hands: Apologies for the delay.

Issue #996, opened 2025-02-20
MIniDebugInfo Causes SIGBUS Crash on ARMv7a When Using DebugSymbol.findFunctionsMatching() (frida-server >= 16.6.0)


TsaiHao:
Commit: 8ed32c4

## Reproduction Steps:
1. Run frida-server with a version later than 16.6.0.
2. Attach to a process on an Armv7a Android device (tested on an Amazon FireTV).
3. Execute any debug symbol command, for example:
```js
DebugSymbol.findFunctionsMatching('print*')
```

## Observed Behavior:

The process crashes with a SIGBUS (BUS_ADRALN) error. Here’s the crash log:
```
[AFTAN69D6D::PID::13901 ]-> DebugSymbol.findFunctionsMatching('print*')
Process crashed: SIGBUS BUS_ADRALN

***
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'xxx'
Revision: '0'
ABI: 'arm'
Timestamp: 2025-02-20 14:45:52.961456989+0100
Process uptime: 12s
Cmdline: /data/local/tmp/wait
pid: 13901, tid: 13935, name: gum-js-loop  >>> /data/local/tmp/wait <<<
uid: 0
signal 7 (SIGBUS), code 1 (BUS_ADRALN), fault addr 0xa6b80059
    r0  00000001  r1  00000000  r2  00000000  r3  ffffffff
    r4  a6f889f0  r5  a6f889f0  r6  a6b80055  r7  a6e1b978
    r8  a6f88a00  r9  a6f889f8  r10 a8e2d644  r11 a6f88998
    ip  a8e1c2c4  sp  a6f88978  lr  a8df9b0f  pc  a787a960
1 total frames
backtrace:
      #00 pc 004ed960  /memfd:frida-agent-32.so (deleted)
***
[AFTAN69D6D::PID::13901 ]->

Thank you for using Frida!
```

## Analysis:
It looks like the crash occurs when reading data from the cached debug info. The stack trace points to the following functions:
• gum_elf_module_enumerate_symbols
• gum_elf_module_enumerate_symbols_in_section (twice)
• gum_elf_module_read_symbol

Notably, the second parameter passed to `gum_elf_module_read_symbol` is `cursor=0xa6b80055`, 0x4 less than the fault address, which suggests that the fault might be happening when reading the value field in: https://github.com/frida/frida-gum/blob/main/gum/gumelfmodule.c#L1816.

[...]

PR #997, opened 2025-02-24
elf-module: Fix memory alignment issue in ELF symbol enumeration

TsaiHao:
- Add memory alignment check and handling for symbol section start address
- Allocate aligned memory buffer when needed to avoid SIGBUS on certain architectures
- Properly free allocated memory buffer in all code paths


TsaiHao:
Trying to fix https://github.com/frida/frida-gum/issues/996

oleavr:
Thank you! I ended up going for a slightly different approach in a9aba9178ab84680fc5d1745fc3dbc603687e9f0, to avoid surprises in the future.

PR #1009, opened 2025-03-26
Fix thread communication deadlock when ptrace operations fail

DoranekoSystems:
The thread modification code wasn't properly handling error cases during ptrace operations, causing deadlocks. 

This fixes the communication protocol in `gum_linux_handle_modify_thread_comms` to properly receive and handle
error acknowledgments from the child process.

See: https://github.com/frida/frida-gum/blob/0671c27d941d77490fff4ca6dbb9ca10a0f872bb/gum/backend-linux/gumprocess-linux.c#L664-L682


oleavr (changes requested):
Hmm, I might be missing something, but to me it looks like the behavior with these proposed changes is identical to before. This is because `gum_await_ack()` reads one byte, and then returns whether that byte is the expected one. It doesn't loop waiting for one specific one. So with the proposed changes you are duplicating that logic in two places instead.

DoranekoSystems:
I think your point is correct. I was mistaken.
However, I believe the original implementation doesn't seem to properly propagate errors. What do you think? 

Also, is the following an implementation mistake?
https://github.com/frida/frida-gum/blob/0671c27d941d77490fff4ca6dbb9ca10a0f872bb/gum/backend-linux/gumprocess-linux.c#L1572-L1587

oleavr:
> I think your point is correct. I was mistaken. However, I believe the original implementation doesn't seem to properly propagate errors. What do you think?

It's possible. Appreciate your help reviewing it. I'm too busy to context-switch to that right now.

> Also, is the following an implementation mistake?
> 
> https://github.com/frida/frida-gum/blob/0671c27d941d77490fff4ca6dbb9ca10a0f872bb/gum/backend-linux/gumprocess-linux.c#L1572-L1587

Ooh oops, that looks like a refactoring gone wrong. No reason to do it twice.

PR #1018, opened 2025-04-04
darwin: Fix double free in find_module_by_address

mrmacete:
An extra call to `g_ptr_array_unref` caused crashes due to memory corruption.


oleavr:
Oops! 🙊 Thanks! 🙌🏻

PR #1021, opened 2025-04-07
darwin: Fix module type in find_module_by_address

mrmacete:
This was a type confusion causing subtle crashes.


oleavr:
Whoah, thanks, great catch! 🤘🙌🏻

PR #1022, opened 2025-04-09
darwin: Enumerate threads faster without flags

mrmacete:
In this way any heap allocation is avoided, so there's no risk of deadlock in use cases like the Interceptor, where thread enumeration is used to suspend and resume threads by id.


oleavr:
Thanks! 🙌

PR #1024, opened 2025-04-16
interceptor: Avoid heap allocation during suspend

mrmacete:
By storing the suspended thread ids into a `GumMetalArray` instead of a `GQueue` we prevent deadlocks when a suspended thread is holding the `dlmalloc` lock.


oleavr:
Thanks! :raised_hands:

PR #1031, opened 2025-05-20
darwin: Fix iOS Simulator support

ChiChou:
This is intended to fix the case that user want to directly attach a process inside Simulator, without frida-gadget

https://github.com/frida/frida/issues/2763#issuecomment-2828670124

The current implementation `gum_darwin_query_sysroot` only supports in process, and the method it depends can't detect `dyld_sim`'s existence. After some experiments I found that the only way to detect the dyld for simulator is to iterate mapped files. Neither `_dyld` api nor `TASK_DYLD_ALL_IMAGE_INFO` work.

This is a draft PR because:

* There is another bug that I haven't located the root cause, that the dynamic dylib injected into the simulator process will jump to a random address and crash
* At the moment we do not have a proper unit test for simulator (do we?)
* My patch includes breaking changes for certain GUM_API interface


gum/backend-darwin/gumdarwinmoduleresolver.c
```diff
@@ -318,7 +318,7 @@ gum_darwin_module_resolver_rebuild_indexes (GumDarwinModuleResolver * self,
     mod = g_ptr_array_index (latest_modules, i);
     path = gum_module_get_path (mod);
 
-    if (g_str_has_suffix (path, "/usr/lib/dyld_sim"))
```
ChiChou, line 321: Note: `dyld_sim` will not be included unless `gum_collect_modules` falls back to `gum_collect_modules_forensically` branch. So this branch was dead code

PR #1033, opened 2025-05-28
Don't attempt to patch inline cache entries if the block has none

WorksButNotTested:


oleavr:
Thanks! Should `gum_exec_block_backpatch_inline_cache()` even be called for such blocks? I'm wondering if the root cause here is that we shouldn't be emitting a call to that function.

WorksButNotTested:
I think the root cause is a difference between the trust threshold at the point the block is prefetched between the parent and child in the forkserver. So the child creates the block with an online cache whilst the parent does not (at least initially).

Trying to track this externally seemed like exposing the consumer of the library to the internal implementation. Further having an external API which would perform a null dereference if called with invalid parameters seemed like a bad idea anyway.

PR #1034, opened 2025-05-29
darwin: Dispose module resolver indexes

mrmacete:
To avoid leaking native modules in scenarios where short-lived resolvers are used (for example on different tasks).


oleavr:
Thanks! 🙌

PR #1045, opened 2025-06-10
Fix issue with test failures not being propagated in emulator

WorksButNotTested:


oleavr:
Thanks! :raised_hands:

PR #1051, opened 2025-07-25
cpu-context: implement neon on linux/arm64

londek:
Hi,
This PR adds support to reading vector registers on aarch64 linux. I've tested it on Android 15 and works great.
It can be easily extended in future in order to implement other registers or bigger vector variants.

I've moved elf.h include from gumandroid.c to gumandroid.h as my IDE had problems with reading it via gumandroid.c, however the final effect should be the same.

I've omitted fpsr and fpcr as they don't seem to be used on other platforms.


oleavr:
Thanks! 🙌

PR #1057, opened 2025-08-06
darwin-module: Swap-hook lldb image notifier

mrmacete:
Since on modern iOS devices the `lldb_image_notifier` function is only 1-instruction long, Interceptor currently fails to attach it. When this happens, the module registry is left non-functional.

This change swaps the pointer in the dyld "all image infos" structure with our own, so there's no need to use Interceptor in order to get image load/unload events.

### DIFF
--- a/gum/backend-darwin/gummoduleregistry-darwin.c
+++ b/gum/backend-darwin/gummoduleregistry-darwin.c
@@ -144,8 +144,6 @@ gum_lldb_image_notifier (enum dyld_image_mode mode,
       gum_remove_image (info[i].imageLoadAddress);
   }
 
-  g_assert (gum_dyld_notifier_context != NULL);
-
   return gum_dyld_notifier_context->original (mode, info_count, info);
 }
 


gum/backend-darwin/gummoduleregistry-darwin.c
```diff
@@ -130,29 +130,22 @@ gum_module_registry_on_image_removed (const struct mach_header * mh,
...
-
   for (i = 0; i != info_count; i++)
   {
     if (mode == dyld_image_adding)
       gum_add_image (info[i].imageLoadAddress, NULL);
     else
       gum_remove_image (info[i].imageLoadAddress);
   }
+
+  if (gum_dyld_notifier_context != NULL)
```
oleavr, line 147: Is this check needed? I guess this is racy regardless, since we're not holding the loader lock.

oleavr (approved):

gum/backend-darwin/gummoduleregistry-darwin.c
mrmacete: let's remove it for simplicity

mrmacete: turn into an assertion, so hopefully if we loose a race we'd know

mrmacete: nah, racy is racy, removed the assertion too

oleavr: Sounds good 👍

oleavr:
Thanks! 🙌

PR #1058, opened 2025-08-08
gumjs-interceptor: Ensure consistent onLeave call

mrmacete:
Avoid calling `onLeave` when `onEnter` is present but hasn't been called yet. This can happen when attaching the listeners mid-call.

In that scenario, calling `onLeave` with a fresh invocation context could result in unexpectedly uninitialised properties leading to unpredictable behaviour.

This change guarantees that when both `onEnter` and `onLeave` are present, `onLeave` is only called back after `onEnter` actually happened.


oleavr:
Thanks! 🙌 I think eventually it would be good to push this down into Interceptor itself.

mrmacete:
yeah i think it makes sense to have in both places, from my early tests fixing this at interceptor level doesn't guarantee it's fixed in gumjs - but i will dig more

PR #1069, opened 2025-10-21
darwin: Optimise apiresolver find_method_by_address and remove symbolutil cache invalidator

mrmacete:
This change brings down the complexity of looking up Objective-C methods by address, by ignoring classes whose containing module is not the same of the address being looked up.

The other factor, even if less impactful, is relying on cached class and method handles, making sure the cache gets refreshed in case modules are added.

All this makes `DebugSymbol.fromAddress()` faster by about 10 times, at the cost of not synthesising the symbols belonging to methods implemented in different modules than the classes', like via Categories or dynamically by swizzling.

**EDIT: rewrote to lookup classes defined in the same module as the address to lookup, or classes which that module defines categories on. With this, categories are fully supported. Only dynamically swizzled methods are currently left out.**

### DIFF
--- a/gum/backend-darwin/gumsymbolutil-darwin.c
+++ b/gum/backend-darwin/gumsymbolutil-darwin.c
@@ -20,48 +20,19 @@
 # include "gumarm64reader.h"
 #endif
 
-#define GUM_TYPE_SYMBOL_CACHE_INVALIDATOR \
-    (gum_symbol_cache_invalidator_get_type ())
-G_DECLARE_FINAL_TYPE (GumSymbolCacheInvalidator,
-                      gum_symbol_cache_invalidator,
-                      GUM, SYMBOL_CACHE_INVALIDATOR,
-                      GObject)
-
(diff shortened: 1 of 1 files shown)


gum/backend-darwin/gumsymbolutil-darwin.c
```diff
@@ -100,10 +81,21 @@ do_deinit (void)
...
+    registry = gum_module_registry_obtain ();
+
+    g_signal_handler_disconnect (registry, invalidator_added_handler);
+    invalidator_added_handler = 0;
+
+    g_signal_handler_disconnect (registry, invalidator_removed_handler);
+    invalidator_removed_handler = 0;
 
-  invalidator_initialized = FALSE;
+    gum_module_registry_unlock (registry);
```
oleavr: Is there a corresponding `lock()` that this balances?

oleavr (changes requested):
w00t! 🤘

PR #1071, opened 2025-12-01
riscv support

wyxhjr:
Initial support, draft stage, some issues with the interceptor need to be resolved.


oleavr (changes requested):
Thanks! 🙌  What is the current state of this? From a quick look it seems it won't compile cleanly (at least not without warnings).

PR #1073, opened 2025-12-10
gum-graft: Add --force option to allow re-grafting

TaleStarsTeam:
This PR introduces a new `-f` / `--force` command-line argument to the `gum-graft` tool.

**Changes:**
- Added `GUM_DARWIN_GRAFTER_FLAGS_FORCE` to skip the check for existing `__FRIDA` segments.
- Users can now force the grafting process even if the binary is already instrumented. This is useful for iterative testing without needing to restore the original binary every time.
- Fixed an uninitialized `GError` variable in `tools/gumgraft.c` that could cause segmentation faults.

### DIFF
--- a/tools/gumgraft.c
+++ b/tools/gumgraft.c
@@ -11,6 +11,7 @@ static gchar ** code_offsets = NULL;
 static gboolean ingest_function_starts = FALSE;
 static gboolean ingest_imports = FALSE;
 static gboolean transform_lazy_binds = FALSE;
+static gboolean force = FALSE;
 
 static GOptionEntry options[] =
 {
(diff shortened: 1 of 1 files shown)


gum/gumdarwingrafter.c
```diff
@@ -301,7 +301,12 @@ gum_darwin_grafter_graft (GumDarwinGrafter * self,
         &g_array_index (module->segments, GumDarwinSegment, i);
 
     if (g_str_has_prefix (segment->name, "__FRIDA_"))
+    {
+      if ((self->flags & GUM_DARWIN_GRAFTER_FLAGS_FORCE) != 0)
```
oleavr: We should skip the entire loop instead of doing this check inside it.

oleavr (changes requested):
Thanks! 🙌  After giving this some more thought, I don't think we should add a flag here. But what we should do instead is improve the logic so we don't end up with duplicate segment names: like we now would with __FRIDA_TEXT0 and __FRIDA_DATA0.

Issue #1074, opened 2025-12-11
`gum_native_module_enumerate_exports` for windows does not distinguish between variables and functions


Ninja3047:
https://github.com/frida/frida-gum/blob/5ee7ae3f1719db72b0b621f660d3a2fff4d045a4/gum/backend-windows/gummodule-windows.c#L272-L283

This is documented as a comment here, but creating an issue here as well for visibility.

One way to check this would be to check if the address is in an executable page, if it is, then it's likely a function, otherwise it's a data export.

PR #1075, opened 2025-12-12
Support data exports on Windows

Ninja3047:
This is determined by checking the page permission of the export address. If it is pointing to an executable page, then it's a function, otherwise it is a variable. This should prevent unintentional patching of variables. 

Closes: https://github.com/frida/frida-gum/issues/1074


oleavr:
Thanks! 🙌

PR #1076, opened 2025-12-19
fix/arm64: Handle fast interceptor patching limitations

Jiay1C:
Added a check in gum_interceptor_backend_prepare_trampoline to return FALSE for fast interceptors when there is insufficient space for a jump instruction to the replacement function. This ensures that patching is only attempted when it is feasible.


oleavr (changes requested):
Thanks! And apologies for the delay.

I'm afraid this breaks support for the deflector code-path though (i.e. jumping through a nearby Mach-O or ELF-header, on OSes where we support this). See this logic:

```C
  if (ctx->type == GUM_INTERCEPTOR_TYPE_FAST)
  {
    deflector_target = ctx->replacement_function;
  }
  ```

PR #1077, opened 2025-12-19
gum-graft: simple & naive implementation of instruction relocations

daniillnull:
This PR introduces simple implementation of instructions relocation in gum-graft (gumdarwingrafter.c).

Currently, instructions are always copied in the trampoline entry "as is". Instrumentation on B or BL instructions will generate invalid code without any warning.

### Changes:
* Grafter will try to relocate instrumented instruction if required
* If instruction should be relocated, but it does not know how to relocate it properly, warning will be shown
* Fixed minor issue in gum_darwin_grafter_transform_load_commands(), so it can process some (probably obfuscated?) binaries properly

### Implementation notes:
* Of course, using Arm64Relocator is more preferable (as it was stated in TODO comment). But it can generate multiple instructions from single jump instruction, so "trampoline" entry becomes variable sized. It comes to be more complex and difficult solution.
* Not sure if printing warnings via g_printerr() is a good way, but I suppose it's better than silently generate invalid code.

Please let me know if any changes need to be made.

### DIFF
--- a/gum/gumdarwingrafter.c
+++ b/gum/gumdarwingrafter.c
@@ -994,7 +994,8 @@ gum_darwin_grafter_transform_load_commands (gconstpointer commands_in,
     {
       const GumSegmentCommand64 * sc = command_in;
 
-      is_linkedit_command = sc->fileoff == layout->linkedit_offset_in;
+      /* XXX: It is known that some files in the wild have segments of 0KB at the same position as __LINKEDIT, so filesize check is needed */
+      is_linkedit_command = (sc->filesize > 0 && sc->fileoff == layout->linkedit_offset_in);
     }
 
     if (is_linkedit_command)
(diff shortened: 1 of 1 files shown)


oleavr (changes requested):
Thanks! And apologies for the delay here.

I think we should support variable-sized trampolines instead, so we won't have to duplicate relocator logic.

oleavr:
I've taken this forward and reworked it on top of the real `Arm64Relocator`, so grafting handles arbitrary instructions — including PC-relative ones — not just position-independent prologues (the direction the on_invoke path's use Arm64Relocator note was pointing at).

Two supporting commits teach the relocator to emit direct / PC-relative forms (branches, BL, ADR/ADRP, LDR) when the target is in range, so the grafted code stays slide-invariant; the grafter then sizes each hook's on_invoke tail individually and packs the trampolines back to back.

I've kept you as author of the grafter commit and added myself as co-author. Verified with the test suite and end-to-end on a real device: the grafted hook fires under required code-signing, with no debugger attached.

Issue #392, opened 2020-01-03
gum_interceptor_attach fails to attach to functions that start with a single uncoditional jump on arm/arm64


JaviMerino:
The interceptor fails to attach to srand() on arm64. On a Pixel 3:

```
$ frida-trace -U -i 'libc.so!srand' com.google.android.calculator
Instrumenting functions...                                              
srand: Loaded handler at "/home/jm/src/frida/__handlers__/libc.so/srand.js"
Error: Skipping "srand": unable to intercept function at 0x75d27ec464; please file a bug
```

The `srand()` function is a single-instruction trampoline to `srandom()`:

```
000000000008d464 <srand>:
   8d464:       140160c3        b       e5770 <srandom@plt>
```

Other single instruction functions fail as well. For example, I can't trace `libcrypto.so!BN_pseudo_rand`. Using this script:

``` python
import frida

session = frida.get_usb_device().attach('com.google.android.calculator')
script = session.create_script("""'use strict';
var addr = Module.findExportByName("libcrypto.so", "BN_pseudo_rand");
send('arch ' + Process.arch);
send('opcode', addr.readByteArray(4));

Interceptor.attach(addr, {
    onEnter: function (args) {}
});""")

def on_message(msg, data):
    if msg['type'] == 'send':
        if msg['payload'] == 'opcode':
            print('[i] First opcode: 0x{}'.format(bytes(reversed(data)).hex()))
        else:
            print('[i] {}'.format(msg['payload']))
    elif msg['type'] == 'error':
        print('[!] {}'.format(msg['stack']))

script.on('message', on_message)
script.load()
```

I get this output:
```
[i] arch arm64
[i] First opcode: 0x1401ae6d
[!] Error: unable to intercept function at 0x75d47b8d0c; please file a bug
    at frida/runtime/core.js:387
    at /script1.js:8
```
Note that the opcode 0x1401ae6d is "b  11b6c0", again an unconditional branch.

Same for arm (32-bit). This script:

```python
import frida

session = frida.get_usb_device().attach('com.android.chrome')
script = session.create_script("""'use strict';
var addr = Module.findExportByName("libcrypto.so", "RAND_set_rand_method");
send('arch ' + Process.arch);
send('opcode', addr.readByteArray(2));

Interceptor.attach(addr, {
    onEnter: function (args) {}
});""")

def on_message(msg, data):
    if msg['type'] == 'send':
        if msg['payload'] == 'opcode':
[...]

PR #1079, opened 2025-12-21
Fix `Process.enumerateModules()` under x86(-64) arch.

Redbeanw44602:
Fix: https://github.com/frida/frida/issues/3582; May be related? https://github.com/frida/frida-gum/issues/392

---

In short, the [rtld_db_dlactivity](https://github.com/LineageOS/android_bionic/blob/5e704b240138776d1b6a54892d3619424ff4016a/linker/linker_gdb_support.cpp#L38) on x64 is shorter, only one byte.

```
(lldb) disassemble -s 0x7f2b2d374140 -c 10
0x7f2b2d374140: retq
0x7f2b2d374141: int3   
0x7f2b2d374142: int3   
0x7f2b2d374143: int3   
0x7f2b2d374144: int3   
0x7f2b2d374145: int3   
0x7f2b2d374145: int3   
0x7f2b2d374146: int3   
0x7f2b2d374147: int3   
0x7f2b2d374148: int3   
```

Unfortunately, the interceptor has issues handling short functions; investigations suggest it requires [at least 5 bytes](https://github.com/frida/frida-gum/blob/742d69e4be49ac76acc5ac7aa4a3c0bb57351c08/gum/backend-x86/guminterceptor-x86.c#L150).

I'm unsure if the interceptor‘s arm64 backend also has this problem. I disassembled the linker64 on the phone and found the function is 8 bytes long, potentially avoiding the issue.

```
.text:00000000000F06E4                   ; void _dl_rtld_db_dlactivity()
.text:00000000000F06E4                                   EXPORT __dl_rtld_db_dlactivity
.text:00000000000F06E4                   __dl_rtld_db_dlactivity                 ; CODE XREF: __dl_notify_gdb_of_load+38↑p
.text:00000000000F06E4                                                           ; __dl_notify_gdb_of_load+60↑p ...
.text:00000000000F06E4                   ; __unwind {
.text:00000000000F06E4 5F 24 03 D5                       BTI             c
.text:00000000000F06E8 C0 03 5F D6                       RET
.text:00000000000F06E8                   ; } // starts at F06E4
```

---

The patch attempts to resolve the problem using the aligned bytes after the function; generally, overwriting them is safe.

Additionally, I added error logging to the `gum_hook_rtld_notifier`; I believe errors shouldn't be ignored here, as they complicate troubleshooting.

### DIFF
--- a/gum/backend-elf/gummoduleregistry-elf.c
+++ b/gum/backend-elf/gummoduleregistry-elf.c
@@ -133,8 +133,11 @@ gum_hook_rtld_notifier (const GumRtldNotifierDetails * details,
     }
   }
 
-  gum_interceptor_attach (gum_rtld_interceptor, impl + offset,
-      *handler, NULL, GUM_ATTACH_FLAGS_UNIGNORABLE);
+  if (gum_interceptor_attach (gum_rtld_interceptor, impl + offset, *handler,
+        NULL, GUM_ATTACH_FLAGS_UNIGNORABLE) != GUM_ATTACH_OK)
+  {
+    gum_panic ("Unable to hook RTLD stub; please file a bug");
+  }
 }
 
 static void


gum/backend-elf/gummoduleregistry-elf.c
```diff
@@ -133,8 +134,11 @@ gum_hook_rtld_notifier (const GumRtldNotifierDetails * details,
     }
   }
 
-  gum_interceptor_attach (gum_rtld_interceptor, impl + offset,
+  attach_result = gum_interceptor_attach (gum_rtld_interceptor, impl + offset,
       *handler, NULL, GUM_ATTACH_FLAGS_UNIGNORABLE);
+  if (attach_result != GUM_ATTACH_OK) {
+    g_error("Failed to attach to rtld stub, error = %d.", attach_result);
+  }
```
oleavr:
```suggestion
  attach_result = gum_interceptor_attach (gum_rtld_interceptor, impl + offset,
      *handler, NULL, GUM_ATTACH_FLAGS_UNIGNORABLE);
  if (attach_result != GUM_ATTACH_OK)
    g_error ("Failed to attach to RTLD stub, error = %d", attach_result);
```

PR #1093, opened 2026-02-09
gumjs: add column names, etc to SQLite API

ChiChou:
This change copy the API style from Bun js to support getting 

* column names
* column types
* declared type in the schema
* number of sql parameters

This is useful to build a SQLite database viewer GUI when we cannot predict the columns in user inputs like "select *"

### DIFF
--- a/bindings/gumjs/gumquickdatabase.c
+++ b/bindings/gumjs/gumquickdatabase.c
@@ -44,6 +44,10 @@ static void gum_database_free (GumDatabase * self);
 static void gum_database_close (GumDatabase * self);
 
 GUMJS_DECLARE_FINALIZER (gumjs_statement_finalize)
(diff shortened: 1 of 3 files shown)


ChiChou:
> Thanks! Could you also open a PR over at DefinitelyTyped to update the typings? That way we can land this and cut a release with up-to-date typings (in Frida.Compiler).

I see the commit messaegs in DefinitelyTyped are always with version numbers, for example:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73323/changes/75880c74ead3caf62d7026e28faed6ea64a2b653

Which version is it going to ship? So I can send the PR. Thanks for reviewing!

ChiChou:
Just sent it
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74539

oleavr:
Thanks! :raised_hands:

PR #1094, opened 2026-02-09
Fix SIGSEGV when enumerating exports from APK-embedded libraries

monkeywave:
This PR fixes a SIGSEGV when calling `Module.enumerateExports()` on Android for APK-embedded shared libraries loaded via direct mmap (e.g. Chrome/Trichrome `libmonochrome_64.so` with `android:extractNativeLibs="false"`).

## Reproduction

**Environment**
- Android 14
- App: Chrome / Trichrome (APK-embedded `.so`)

**Steps**
```js
Process.getModuleByName("libmonochrome_64.so").enumerateExports();
```

**Observed (before)**
- Process crash: `SIGSEGV` / `SEGV_ACCERR`

**Expected**
- No crash; return exports (or a safe fallback if the relevant ELF tables are not readable)

## Root Cause

On Android when native libraries are **not extracted** (`android:extractNativeLibs="false"`), the dynamic linker
maps the `.so` directly from the APK and may introduce **PROT_NONE gaps** between `PT_LOAD` segments.
Current ELF parsing may dereference pointers that land inside these unreadable gaps (e.g. during access to
`STRTAB`, `SYMTAB`, or hash tables), causing `SEGV_ACCERR`.

## Changes

1. **ONLINE mode:** skip APK extraction for embedded libraries (data is already available at the module base).
2. **Hardening:** validate memory protections before reading ELF metadata (string table, symbol table, hash tables).
3. **Performance:** cache per-module page protection results to avoid repeated queries.
4. **Safety:** add null/invalid checks and return a graceful fallback instead of crashing.

## Behavior After

- No crash in `enumerateExports()`.
- If ELF sections/tables are readable: exports are returned normally.
- If some tables are unreadable due to `PROT_NONE` gaps: returns an empty (or partial) result instead of faulting.

## Testing

Manually verified on Android 14 with Chrome/Trichrome:
```js
const m = Process.getModuleByName("libmonochrome_64.so");
m.enumerateExports();
```

- Before: SIGSEGV
- After: works reliably (exports returned when accessible; otherwise safe fallback)

Please let me know if any changes need to be made or if I should test it with certain Android versions as well :-)

### DIFF
--- a/gum/gumelfmodule.c
+++ b/gum/gumelfmodule.c
@@ -98,6 +98,8 @@ struct _GumElfModule
   GumElfDynamicAddressState dynamic_address_state;
   const gchar * dynamic_strings;
 
+  GHashTable * page_is_readable_cache;
+
   GMutex mutex;
 
   GumElfModule * fallback_elf_module;
@@ -344,6 +346,8 @@ gum_elf_module_init (GumElfModule * self)
   self->dynamic_address_state = GUM_ELF_DYNAMIC_ADDRESS_PRISTINE;
 
   g_mutex_init (&self->mutex);
+
+  self->page_is_readable_cache = g_hash_table_new (NULL, NULL);
 }
 
 static void
@@ -364,6 +368,8 @@ gum_elf_module_finalize (GObject * object)
   GumElfModule * self = GUM_ELF_MODULE (object);
 
   g_mutex_clear (&self->mutex);
+
+  g_hash_table_unref (self->page_is_readable_cache);
 
   g_ptr_array_unref (self->sections);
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! I was hoping to get this into 17.7.0, but ran out of time. Will get to this first thing over the weekend.

oleavr:
Simplified a bit, and changed `gum_elf_module_parse_symbol()` back to not expecting `strings` to ever be `NULL`. Looking at the callers, this should be impossible.

I wasn't able to reproduce this issue w/ Chrome on Android 16 though, so please let me know in case I broke something :smile:

PR #1099, opened 2026-03-06
Fix SIGSEGV when enumerating exports from APK-embedded libraries (ONLINE mode)

monkeywave:
Hey @oleavr,

first of all: thank you for your amazing work on Frida 🙏 — it’s hugely appreciated :-)

I tried the recent changes from **#1094**, but on my side the crash still happens (now consistently in a different code-path). I also left details here, in case it helps for cross-reference:  
https://github.com/frida/frida-gum/pull/1094#issuecomment-3999398128

## Summary

On **Android 14 + Android 16 AVD** (Google APIs, arm64) with `com.android.chrome`, this still crashes:

```js
var m = Process.getModuleByName("libmonochrome_64.so");
m.enumerateExports();
```

It only reproduces when the module is **APK-embedded** (path contains `!`), e.g.:

```
.../TrichromeLibrary.apk!/lib/arm64-v8a/libmonochrome_64.so
```

The tombstone’s top frame points at `__fseeko64()`, which suggests we’re crashing in the **minizip APK extraction path**, i.e., not the in-memory ELF parsing path.

## Proposed fix in this PR

If we’re in **ONLINE** mode and the path contains `!`, we can **skip `gum_maybe_extract_from_apk()`** and use the already-mapped bytes at `base_address` directly:

```c
else if (self->source_mode == GUM_ELF_SOURCE_MODE_ONLINE &&
         strstr (self->source_path, "!") != NULL)
{
  self->file_bytes = g_bytes_new_static (
      GSIZE_TO_POINTER (self->base_address),
      G_MAXSIZE - self->base_address);
}
```

Rationale: for ONLINE mode, the in-memory module image is already available and sufficient for export enumeration; avoiding APK I/O also avoids the minizip `FILE*` crash entirely.

## Result

With the patch applied, `enumerateExports()` works reliably and returns a normal export list (example):

```js
[Android Emulator 5554::Chrome ]-> m
{
  "base": "0x71d6807000",
  "name": "libmonochrome_64.so",
  "path": ".../TrichromeLibrary.apk!/lib/arm64-v8a/libmonochrome_64.so",
  "size": 148017152,
    "version": null
}

[Android Emulator 5554::Chrome ]-> m.enumerateExports();
[
    {
        "address": "0x71d90d4888",
        "name": "Java_J_N__1V_1IJJO",
        "type": "function"
    },
    {
        "address": "0x71d90de080",
        "name": "Java_J_N__1V_1IOOOOOOO",
        "type": "function"
    },
    {
        "address": "0x71d90bc430",
        "name": "Java_J_N__1I_1J",
        "type": "function"
    },
  ...
]
```

<details>
<summary>Crash excerpt (before patch)</summary>

```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0000000000000060
Cause: null pointer dereference
#00 pc ... /apex/com.android.runtime/lib64/bionic/libc.so (__fseeko64(...)+36)
#01 pc ... /memfd:frida-agent-64.so (deleted)
...
```

</details>

Thanks again :-)

All the best

Daniel

### DIFF
--- a/gum/gumelfmodule.c
+++ b/gum/gumelfmodule.c
@@ -550,6 +550,13 @@ gum_elf_module_load (GumElfModule * self,
       self->file_bytes = g_bytes_new_static (
           GSIZE_TO_POINTER (self->base_address), gum_query_page_size ());
     }
+    else if (self->source_mode == GUM_ELF_SOURCE_MODE_ONLINE &&
+             strstr (self->source_path, "!") != NULL)
+    {
+      self->file_bytes = g_bytes_new_static (
+          GSIZE_TO_POINTER (self->base_address),
+          G_MAXSIZE - self->base_address);
+    }
     else
 #endif
     if (!gum_maybe_extract_from_apk (self->source_path, &self->file_bytes))
--- a/gum/gumelfmodule.c
+++ b/gum/gumelfmodule.c
@@ -207,6 +207,8 @@ static gboolean gum_store_symtab_params (
     const GumElfDynamicEntryDetails * details, gpointer user_data);
 static gboolean gum_adjust_symtab_params (const GumElfSectionDetails * details,
     gpointer user_data);
+static gboolean gum_count_and_forward_symbol (
+    const GumElfSymbolDetails * details, gpointer user_data);
 static void gum_elf_module_enumerate_symbols_in_section (GumElfModule * self,
     GumElfSectionType section, GumFoundElfSymbolFunc func, gpointer user_data);
 static gboolean gum_emit_each_needed (const GumElfDynamicEntryDetails * details,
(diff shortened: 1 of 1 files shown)


monkeywave:
Hey @oleavr ,

thanks for looking into this :-) I built frida-server with the latest frida-core (including your `fseeko64`/`ftello64` libc-shim additions from f0fd7aef) and can confirm it resolves the SIGSEGV -- the type confusion between `FridaFile*` and bionic's `FILE*` seems to be the direct cause of the crash.

That said, I think the changes in this PR are still valuable as a complementary fix at the frida-gum layer:

- **Avoiding the APK extraction path entirely**: Even with the libc-shim fix, `gum_maybe_extract_from_apk()` still fires for APK-embedded libraries. It works now (no crash), but it's extracting a large library from the APK via minizip just to read section headers -- when the ELF data is already mapped in process memory. Skipping this for APK paths avoids that unnecessary work.

- **`.dynsym` fallback for `enumerateSymbols()`**: APK-embedded `.so` files are typically stripped (no `.symtab`). So even after a successful APK extraction, `enumerateSymbols()` would return `[]`. The `.dynsym` fallback ensures we still return meaningful symbol data from the dynamic symbol table, which is always available in memory via `PT_DYNAMIC`.

Your libc-shim fix is the right solution at the stdio layer -- it benefits all code in the agent that touches shimmed file handles. This PR just makes sure frida-gum takes the optimal code path for APK-embedded ELFs in the first place.

Happy to hear your thoughts :-)

oleavr:
Thanks for testing, that's good to hear! :)

> * Avoiding the APK extraction path entirely

Is it safe to assume that this is always the case, and there aren't APKs out there with unstripped .so files? I haven't looked into this and lack data, so I'm hesitant to assume so.

> * `.dynsym` fallback for `enumerateSymbols()`

Makes sense. Let's land this part now.

oleavr:
Just landed the .dynsym fallback in 01eadbff559a7f787c7cc71a2b6537055b2a89fb. Thanks! :raised_hands:

oleavr:
Closing due to inactivity. Feel free to pick this back up at a later point 🙌🏻

PR #1104, opened 2026-04-10
Free slow slabs on Stalker context destroy

v-p-b:
This fixes a memory leak. 

When using Stalker on Windows (x86-64) I observed commited memory bytes to raise linearly without actual physical memory consumption, eventually leading to "0xc000012d - Out of Virtual Memory" errors. With this change the commited memory remains stable in the stalked process. 

I didn't check the behavior with other backends.

### DIFF
--- a/gum/backend-x86/gumstalker-x86.c
+++ b/gum/backend-x86/gumstalker-x86.c
@@ -2524,7 +2524,6 @@ gum_exec_ctx_free (GumExecCtx * ctx)
     slow_slab = next;
   }
 
-
   g_object_unref (ctx->sink);
   g_object_unref (ctx->transformer);
   g_clear_object (&ctx->observer);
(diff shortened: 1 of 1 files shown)


gum/backend-x86/gumstalker-x86.c
```diff
@@ -2508,6 +2509,22 @@ gum_exec_ctx_free (GumExecCtx * ctx)
...
+
+    is_initial = next == NULL;
+    if (is_initial)
+      break;
+
+    gum_code_slab_free (&slow_slab->slab);
+
+    slow_slab = next;
+  }
+
```
oleavr, line 2526:
```suggestion
```

oleavr (approved):
Thanks, great catch! 🙌

v-p-b:
Thanks for the quick merge! I think you should check the implementations for other platforms too, e.g.:

https://github.com/frida/frida-gum/blob/81c22932f00aba01e8109b332f0b4f30d0a5414c/gum/backend-arm64/gumstalker-arm64.c#L2224

... this implementation looks very similar to the leaky one but I don't have a test system at hand to verify it.

PR #1106, opened 2026-05-03
stalker-x86: Skip AVX2 ymm save on Win7 WoW64

Fitblip:
Hi @oleavr! Long time no chat :) I've been working on a project that leverages the stalker infrastructure and found a bug with how stalker's AVX2 instructions work on win7 64 bit/WoW64. 

On Windows 7 running a 32-bit (WoW64) process, executing the AVX2 ymm save/restore vextracti128/vinserti128) emitted into Stalker's prolog/epilog from a JIT RWX page corrupts wow64cpu state. TEB64 fields are zeroed (NtTib.Self at +0x30, PEB pointer at +0x60, Win32ThreadInfo at +0x78, and TlsSlots[0] / WOW64_TLS_STACKPTR64 at +0x1480), causing the next wow64 syscall to fault inside wow64cpu!CpupReturnFromSimulatedCode while reading `mov rsp, [r12+0x1480]` as zero.

The bug is wow64cpu-specific to NT 6.1: Win7 32-bit native, Win10/11 WoW64, and 64-bit processes are unaffected as far as I can tell. fxsave already preserves the lower 128 bits of YMM, and the Windows x86 ABI does not preserve YMM upper halves across calls, so omitting the save is safe (I think).

All that said I'm not really sure the right way to test this. I was thinking we could ensure that those bytes aren't emitted in the stalker's memory pages but there's not an easy way to see the pages the stalker uses and byte scanning feels flakey at best. I'm open to ideas! 

Anecdotally this fix seems to be exactly what fixes my stalking pipeline when running 32 bit programs on win 7 64 bit. 

LMK what you think! 🚀 Thank you!

### DIFF
--- a/gum/backend-windows/gumprocess-windows.c
+++ b/gum/backend-windows/gumprocess-windows.c
@@ -1018,6 +1018,52 @@ beach:
   }
 }
 
+gboolean
+_gum_windows_is_win7_wow64 (void)
+{
+  static gsize initialized = FALSE;
+  static gboolean is_win7_wow64;
+
+  if (g_once_init_enter (&initialized))
+  {
+    GumIsWow64ProcessFunc is_wow64_process;
+    HMODULE kernel32 = GetModuleHandleW (L"kernel32.dll");
+
+    is_win7_wow64 = FALSE;
+
+    is_wow64_process = (GumIsWow64ProcessFunc)
+        GetProcAddress (kernel32, "IsWow64Process");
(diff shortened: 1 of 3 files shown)


oleavr:
Hey!

Yeah, time flies! :) Hope you're doing well.

Thanks for this. Great catch! Just landed it with some tweaks as a6d2a1ff5316af0df8258eb52b0e5a6d88d7ff85.

Cheers!

PR #1107, opened 2026-05-08
darwin: Ensure page plan on allocated code pages

mrmacete:
Before this change, `gum_memory_is_codesign_out_of_the_way` always returned TRUE for freshly allocated memory, and those pages didn't end up in the page plan when debugger mappings are enforced.

On older devices this is not a problem, but on more recent SoCs page plans are needed for these too in order to avoid failing for missing codesign at execution time. This left `Interceptor` and `Memory.patchCode` non-functional on iOS 26+ for more recent apple devices.

This change introduces the `gum_page_is_freshly_allocated` function to explicitly detect pages in such state (by using `mincore`) and ensure they get included in the page plan.

### DIFF
--- a/gum/backend-darwin/gummemory-darwin.c
+++ b/gum/backend-darwin/gummemory-darwin.c
@@ -50,6 +50,7 @@ static gboolean gum_try_alloc_in_range_if_near_enough (
 static gboolean gum_try_suggest_allocation_base (const GumMemoryRange * range,
     const GumAllocNearContext * ctx, gpointer * allocation_base);
 static gint gum_page_protection_to_bsd (GumPageProtection prot);
+static gboolean gum_page_is_freshly_allocated (gpointer page, gsize size);
 
 void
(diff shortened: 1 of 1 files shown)


gum/backend-darwin/gummemory-darwin.c
```diff
@@ -922,6 +927,21 @@ gum_memory_is_codesign_out_of_the_way (gpointer page)
...
 }
 
+static gboolean
+gum_page_is_freshly_allocated (gpointer page,
+                               gsize size)
+{
+  guint8 vec = 0;
+
+  /* TODO: Implement shim. */
+  if (mincore (page, size, (char *)&vec) != 0)
```
oleavr:
```suggestion
  if (mincore (page, size, (char *) &vec) != 0)
```

oleavr (approved):
Yay, great catch! 🤘

PR #1109, opened 2026-05-11
gumjs: Extend Memory.alloc() to specify protection

mrmacete:
Add the optional `protection` field to the existing options object. If not specified, it falls back to `"rw"` to be consistent with the current behaviour.

In this way it's possible to allocate page-aligned executable memory even in scenarios where memory can't be flipped from rw to rx (like on non-jailbroken iOS 26+).


oleavr:
Thanks! 🙌

Issue #1095, opened 2026-02-18
gum_interceptor_replace trampoline breaks ARM PC-relative conditional branches (`beq`) on ARMv7


AkosBuhler:
### Summary

On ARMv7 (ARM mode), `gum_interceptor_replace()` sometimes produces a broken trampoline when the overwritten prologue contains a **PC-relative conditional branch** (e.g. `beq`). The branch instruction appears to be copied into the trampoline without relocating/fixing up its target, so the branch target becomes incorrect when executed from the trampoline address.

```c
GumReplaceReturn gum_interceptor_replace (GumInterceptor * self,
    gpointer function_address, gpointer replacement_function,
    gpointer replacement_data, gpointer * original_function);
```

If the original function’s prologue includes a **PC-relative `beq`**, the trampoline’s copy of that instruction branches to the wrong target (because it’s now running from a different address). This can lead to incorrect control-flow / crashes.

### Reproduction notes

- Hook created with `gum_interceptor_replace()`
- `function_address` redirects correctly to `replacement_function`
- Calling `original_function` executes the trampoline
- Trampoline does not correctly handle relocation of PC-relative conditional branches in the copied prologue

### Original function (ARM, disassembly)

```asm
000002d8 <BEF_RIO_EBD1318_ReadDin>:
 2d8:	e3500000 	cmp	r0, #0
 2dc:	0a000012 	beq	32c <BEF_RIO_EBD1318_ReadDin+0x54>
 2e0:	e92d4070 	push	{r4, r5, r6, lr}
 2e4:	e1a05002 	mov	r5, r2
 2e8:	e5902004 	ldr	r2, [r0, #4]
 2ec:	e3a03001 	mov	r3, #1
 2f0:	e1a03113 	lsl	r3, r3, r1
 2f4:	e1a04001 	mov	r4, r1
 2f8:	e1d33002 	bics	r3, r3, r2
 2fc:	0a000007 	beq	320 <BEF_RIO_EBD1318_ReadDin+0x48>
 300:	e5900000 	ldr	r0, [r0]
 304:	ebfffffe 	bl	0 <ebd1318_get_input>
 308:	e7eb3050 	ubfx	r3, r0, #0, #12
 30c:	e3a00000 	mov	r0, #0
 310:	e1a03453 	asr	r3, r3, r4
 314:	e2033001 	and	r3, r3, #1
 318:	e5c53000 	strb	r3, [r5]
[...]

PR #1112, opened 2026-05-18
arm32-relocator: Fix B<cond> silently dropping condition code

Lich4rr:
# arm: fix conditional branch relocation

## Problem

`gumthumbrelocator` and `gumarm64relocator` both implement `rewrite_b_cond`
to correctly relocate conditional branches. 

`gumarmrelocator` has no equivalent: `rewrite_b` handles all `ARM_INS_B`
regardless of condition code and silently emits an unconditional `LDR PC, =target`, 
corrupting control flow.

## Fix

`write_one` now dispatches `ARM_INS_B` to a new
`gum_arm_relocator_rewrite_b_cond` when the condition code is not `AL`,
mirroring the approach used by `gumthumbrelocator` and `gumarm64relocator`:

Two test cases `b_cond_imm_a1_positive_should_be_rewritten` and
`b_cond_imm_a1_negative_should_be_rewritten` are added to
`tests/core/arch-arm/armrelocator.c`.

Fixes **#1095**


oleavr:
Thanks! 🙌

PR #1113, opened 2026-05-25
fix(stalker): account for pending literal pool size in is_out_of_space()

### LINKED ISSUE #1115: Stalker (ARM64): access violation when putCallout() is called on every instruction in a large basic block (>~1100 insns)
## Environment

| | |
|---|---|
| **Frida version** | 17.9.1 (client + server, matched) |
| **frida-server** | 17.9.1 |
| **Device** | Pixel 7, Android 16, kernel 6.1, arm64-v8a |
| **Host** | macOS, frida-tools 14.8.1 |
| **Runtime** | V8 |

## Summary

When `Stalker.follow()` is used with a `transform` callback that calls `iter.putCallout()` on **every** instruction in a basic block containing ~1100+ instructions, the JIT-compiled code crashes with an access violation at a nonsensical address. The crash is **deterministic** and **reproducible with a minimal synthetic SO** (no proprietary code).

## Observed Behaviour

```
[3] Stalker transform, putCallout on EVERY instruction  <- expect crash
[script-error] Error: access violation accessing 0xa0a2b5d400000212
```

The crash address changes between runs but is always clearly invalid — the low 32 bits look like a small integer (byte offset from slab start), not a real pointer.

| Callout density | Result |
|---|---|
| Every 2nd instruction (stride=2, ~827 callouts) | ✅ PASS |
| Every instruction (stride=1, ~1671 callouts) | 💥 CRASH |
| Empirical threshold | ~1100–1200 callouts in one basic block |

## Root Cause

`gum_stalker_iterator_is_out_of_space()` in `gumstalker-arm64.c` decides when to split a large basic block across slab boundaries:

```c
capacity = (guint8 *) gum_slab_end (slab) -
    (guint8 *) gum_arm64_writer_cur (cw);
snapshot_size = gum_stalker_snapshot_space_needed_for (...);
return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size;
```

It reserves `GUM_EXEC_BLOCK_MIN_CAPACITY (2048)` + snapshot space, but **does not account for the pending ARM64 literal pool** that `gum_arm64_writer_flush()` → `commit_literals()` will write at the current code pointer when the block is finalized.

Each `putCallout()` adds 2 LDR-literal references (for `entry_address` and `gum_stalker_invoke_callout`), contributing 2 × 8 = 16 bytes of pending literal pool data per callout. With ~1500 callouts in a single block, that's **~24 KB** of literal pool.

Additionally, `gum_arm64_writer_maybe_commit_literals()` has a threshold of 1 MB (`space_used <= 1048572`), which never triggers inside a 128 KB code slab — so the pool grows unbounded until flush.

When `is_out_of_space` fires (slab has < 2048 bytes left), `flush()` writes the ~24 KB literal pool **past `slab_end`**, corrupting the adjacent slow/data slab. The corrupted memory is later executed as JIT code, causing the access violation at garbage addresses like `0x75c635d400000213`.

## Proposed Fix

One-function patch to `gum/backend-arm64/gumstalker-arm64.c`:

```diff
 static gboolean
 gum_stalker_iterator_is_out_of_space (GumStalkerIterator * self)
 {
   GumExecBlock * block = self->exec_block;
   GumSlab * slab = &block->code_slab->slab;
-  gsize capacity, snapshot_size;
+  GumArm64Writer * cw = self->generator_context->code_writer;
+  gsize capacity, snapshot_size, pending_pool_size;
 
   capacity = (guint8 *) gum_slab_end (slab) -
-      (guint8 *) gum_arm64_writer_cur (self->generator_context->code_writer);
+      (guint8 *) gum_arm64_writer_cur (cw);
 
   snapshot_size = gum_stalker_snapshot_space_needed_for (
       self->exec_context->stalker,
       self->generator_context->instruction->end - block->real_start);
 
-  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size;
+  /* Account for pending literal references that commit_literals() will write
+   * at the current code pointer when the block is flushed. Each 64-bit ref
+   * occupies 8 bytes in the pool (worst case: no deduplication). We also
+   * reserve 4 bytes for the B instruction that skips over the pool. */
+  pending_pool_size = 0;
+  if (cw->literal_refs.data != NULL)
+  {
+    pending_pool_size = cw->literal_refs.length * sizeof (guint64) + 4;
+  }
+
+  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size
+      + pending_pool_size;
 }
```

**Why this works:** When the pending literal pool grows large enough, `is_out_of_space` returns TRUE earlier, causing `iterator_next()` to split the block. The next block starts with a fresh writer (literal_refs = 0), so the pool never overflows the slab. For normal transforms without heavy callout usage, `pending_pool_size` is effectively 0 — behavior unchanged.

## Verification

Built frida-gum from source (current main + this patch), cross-compiled frida-server for android-arm64, pushed to device, ran the reproducer:

```
[1] Baseline (no Stalker)                               PASS
[2] Stalker transform, putCallout every 2nd instruction  PASS  callouts_fired=827
[3] Stalker transform, putCallout on EVERY instruction   PASS  callouts_fired=1671
```

All 1671 instructions traced with per-instruction callouts — no crash.

## Workaround (until the fix lands)

Limit callouts per block to ~500 in your transform callback:

```javascript
Stalker.follow(threadId, {
  transform(iter) {
    let ins, calloutCount = 0;
    while ((ins = iter.next()) !== null) {
      if (calloutCount < 500) {
        iter.putCallout(() => { /* your callout */ });
        calloutCount++;
      }
      iter.keep();
    }
  }
});
```

### Step 1 — build a synthetic SO with a ~1500-instruction branchless block

```python
# build_repro_so.py
import subprocess, sys, os, tempfile

NUM_INSNS = 1500
NDK_CLANG = "/path/to/ndk/aarch64-linux-android28-clang"  # adjust
OUTPUT_SO = "librepro_stalker.so"

def generate_asm(n):
    lines = [".text", ".global big_block", ".type big_block, %function", "big_block:"]
    lines += ["    ldr x1, [x0]", "    ldr x2, [x0, frida/frida#8]", "    ldr x3, [x0, frida/frida#16]",
              "    ldr x4, [x0, frida/frida#24]", "    mov x5, #0", "    mov x6, frida/frida#1",
              "    mov x7, #0x41", "    mov x8, #0xff"]
    regs = ["x%d" % i for i in range(1, 16)]
    count = 8
    while count < n - 5:
        i, j, k = count % 15, (count*7+3) % 15, (count*13+5) % 15
        op = count % 6
        if   op == 0: lines.append("    add %s, %s, %s" % (regs[i], regs[j], regs[k]))
        elif op == 1: lines.append("    eor %s, %s, %s" % (regs[i], regs[j], regs[k]))
        elif op == 2: lines.append("    lsl %s, %s, #%d" % (regs[i], regs[j], (count%4)+1))
        elif op == 3: lines.append("    sub %s, %s, %s" % (regs[i], regs[j], regs[k]))
        elif op == 4: lines.append("    and %s, %s, %s" % (regs[i], regs[j], regs[k]))
        elif op == 5: lines.append("    orr %s, %s, %s" % (regs[i], regs[j], regs[k]))
        count += 1
    lines += ["    str x1, [x0]", "    str x2, [x0, frida/frida#8]",
              "    str x3, [x0, frida/frida#16]", "    str x4, [x0, frida/frida#24]", "    ret"]
    lines.append(".size big_block, .-big_block")
    return "\n".join(lines)

with tempfile.NamedTemporaryFile(mode='w', suffix='.S', delete=False) as f:
    f.write(generate_asm(NUM_INSNS)); asm = f.name
subprocess.run([NDK_CLANG, "-shared", "-O0", "-nostdlib", "-o", OUTPUT_SO, asm], check=True)
os.unlink(asm)
print("Built", OUTPUT_SO)
```

### Step 2 — push to device and run

```bash
adb push librepro_stalker.so /data/local/tmp/
frida -U -f /system/bin/sleep --runtime=v8 -l repro_official.js
```

### Step 3 — Frida script (`repro_official.js`)

```javascript
'use strict';
const SO_PATH = '/data/local/tmp/librepro_stalker.so';

function run() {
    const mod    = Module.load(SO_PATH);
    const fnAddr = mod.findExportByName('big_block');
    const buf    = Memory.alloc(256);
    for (let i = 0; i < 32; i++) buf.add(i * 8).writeU64(i + 1);
    const fn = new NativeFunction(fnAddr, 'void', ['pointer']);

    // Test 1: baseline
    fn(buf);
    console.log('[1] baseline PASS');

    // Test 2: stride=2 (should pass)
    {
        let callouts = 0;
        Interceptor.attach(fnAddr, {
            onEnter() {
                Stalker.follow(this.threadId, {
                    events: { compile: false },
                    transform(iter) {
                        let ins, n = 0;
                        while ((ins = iter.next()) !== null) {
                            if (++n % 2 === 0) iter.putCallout(() => { callouts++; });
                            iter.keep();
                        }
                    }
                });
            },
            onLeave() { Stalker.unfollow(this.threadId); Stalker.flush(); Stalker.garbageCollect(); }
        });
        fn(buf);
        Interceptor.detachAll(); Stalker.garbageCollect();
        console.log('[2] stride=2 PASS  callouts=' + callouts);
    }

    // Test 3: stride=1 (triggers the bug)
    {
        Interceptor.attach(fnAddr, {
            onEnter() {
                Stalker.follow(this.threadId, {
                    events: { compile: false },
                    transform(iter) {
                        let ins;
                        while ((ins = iter.next()) !== null) {
                            iter.putCallout(() => {});
                            iter.keep();
                        }
                    }
                });
            },
            onLeave() { Stalker.unfollow(this.threadId); Stalker.flush(); Stalker.garbageCollect(); }
        });
        fn(buf);   // <-- access violation here without fix
        console.log('[3] stride=1 PASS (unexpected without fix)');
    }
}
setTimeout(run, 200);
```

---

Happy to submit a PR to frida-gum with this fix.

ltlly:
## Summary

When `putCallout()` is used on every instruction in a large branchless basic block (~1500+ instructions), each callout adds literal references to the code writer. The `gum_stalker_iterator_is_out_of_space()` check did not account for these pending literal pool bytes, allowing the pool to overflow past `slab_end` during flush and corrupt adjacent memory.

This manifests as SIGILL or SIGBUS crashes on ARM64 and ARM32 targets.

## Root Cause

In both ARM64 and ARM32 stalker backends, `is_out_of_space()` calculates remaining capacity as:

```c
capacity = slab_end - current_code_pointer;
return capacity < MIN_CAPACITY + snapshot_size;
```

But when the block is eventually flushed, `commit_literals()` writes the literal pool **at the current code pointer**, consuming `literal_refs.length * sizeof(guint64) + 4` bytes (ARM64) or `literal_refs.length * sizeof(guint32) + 4` bytes (ARM32). This space was never reserved, so with enough pending literals the pool overflows past `slab_end`.

## Fix

Include the pending literal pool size in the capacity check:

```c
pending_pool_size = cw->literal_refs.length * sizeof(guint64) + 4;
return capacity < MIN_CAPACITY + snapshot_size + pending_pool_size;
```

This causes blocks to split earlier when the literal pool grows large, preventing the overflow.

## Testing

Added regression tests for all affected architectures:
- **ARM64**: `many_callouts_should_not_overflow_literal_pool` — 1500 ADD instructions with a callout on every instruction
- **ARM32 (ARM mode)**: `arm_many_callouts_should_not_overflow_literal_pool`
- **ARM32 (Thumb mode)**: `thumb_many_callouts_should_not_overflow_literal_pool`

Verified locally on macOS ARM64:
- **Without fix**: test crashes with SIGILL (exit 132)
- **With fix**: test passes, all 46 Stalker tests pass with zero regressions

CI results on fork (armbe8 emulator confirms ARM32 fix works):
- native (linux-x86_64, linux-x86, windows-x86, windows-x86_64): ✅
- emulator (armbe8 — ARM32 BE, runs gum-tests in QEMU): ✅
- mobile (android-arm, android-arm64, android-x86, android-x86_64): ✅ (compile)
- manylinux (armhf, arm64, x86, x86_64): ✅

Closes https://github.com/frida/frida-gum/issues/1115

### DIFF
--- a/gum/backend-arm/gumstalker-arm.c
+++ b/gum/backend-arm/gumstalker-arm.c
@@ -2606,7 +2606,7 @@ gum_stalker_iterator_is_out_of_space (GumStalkerIterator * self)
   GumGeneratorContext * gc = self->generator_context;
   GumSlab * slab = &block->code_slab->slab;
   guint8 * cursor;
-  gsize capacity, snapshot_size;
+  gsize capacity, snapshot_size, pending_literals_size;
 
   cursor = gc->is_thumb
       ? gum_thumb_writer_cur (gc->thumb_writer)
@@ -2618,7 +2618,26 @@ gum_stalker_iterator_is_out_of_space (GumStalkerIterator * self)
       self->exec_context->stalker,
       gc->instruction->end - block->real_start);
 
-  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size;
+  if (gc->is_thumb)
+  {
+    GumThumbWriter * tw = gc->thumb_writer;
+    gsize alignment_padding = sizeof (guint16);
+
+    pending_literals_size = (tw->literal_refs.data != NULL)
+        ? (tw->literal_refs.length * sizeof (guint32)) + alignment_padding
+        : 0;
+  }
+  else
+  {
+    GumArmWriter * aw = gc->arm_writer;
+
+    pending_literals_size = (aw->literal_refs.data != NULL)
+        ? aw->literal_refs.length * sizeof (guint32)
+        : 0;
+  }
+
+  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size +
+      pending_literals_size;
 }
 
 void
--- a/gum/backend-arm64/gumstalker-arm64.c
+++ b/gum/backend-arm64/gumstalker-arm64.c
@@ -2924,16 +2924,22 @@ gum_stalker_iterator_is_out_of_space (GumStalkerIterator * self)
 {
   GumExecBlock * block = self->exec_block;
   GumSlab * slab = &block->code_slab->slab;
-  gsize capacity, snapshot_size;
+  GumArm64Writer * cw = self->generator_context->code_writer;
+  gsize capacity, snapshot_size, pending_literals_size;
 
   capacity = (guint8 *) gum_slab_end (slab) -
-      (guint8 *) gum_arm64_writer_cur (self->generator_context->code_writer);
+      (guint8 *) gum_arm64_writer_cur (cw);
 
   snapshot_size = gum_stalker_snapshot_space_needed_for (
       self->exec_context->stalker,
       self->generator_context->instruction->end - block->real_start);
 
-  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size;
+  pending_literals_size = (cw->literal_refs.data != NULL)
+      ? cw->literal_refs.length * sizeof (guint64)
+      : 0;
+
+  return capacity < GUM_EXEC_BLOCK_MIN_CAPACITY + snapshot_size +
+      pending_literals_size;
 }
 
 void
--- a/tests/core/arch-arm/stalker-arm-fixture.c
+++ b/tests/core/arch-arm/stalker-arm-fixture.c
@@ -105,6 +105,8 @@
  */
 #define INVOKER_IMPL_OFFSET 24
 
+#define MANY_CALLOUTS_INSN_COUNT 1500
+
 typedef struct _TestArmStalkerFixture TestArmStalkerFixture;
 typedef struct _UnfollowTransformContext UnfollowTransformContext;
 typedef struct _InvalidationTransformContext InvalidationTransformContext;
--- a/tests/core/arch-arm/stalker-arm.c
+++ b/tests/core/arch-arm/stalker-arm.c
@@ -102,6 +102,8 @@ TESTLIST_BEGIN (stalker)
   TESTENTRY (thumb_callout)
   TESTENTRY (arm_transformer_should_be_able_to_replace_call_with_callout)
   TESTENTRY (arm_transformer_should_be_able_to_replace_jumpout_with_callout)
+  TESTENTRY (arm_many_callouts_should_not_overflow_literal_pool)
+  TESTENTRY (thumb_many_callouts_should_not_overflow_literal_pool)
   TESTENTRY (unfollow_should_be_allowed_before_first_transform)
   TESTENTRY (unfollow_should_be_allowed_mid_first_transform)
   TESTENTRY (unfollow_should_be_allowed_after_first_transform)
@@ -177,6 +179,10 @@ static void replace_call_with_callout (GumStalkerIterator * iterator,
 static void replace_jumpout_with_callout (GumStalkerIterator * iterator,
     GumStalkerOutput * output, gpointer user_data);
 static void callout_set_cool (GumCpuContext * cpu_context, gpointer user_data);
+static void add_callout_for_every_instruction (GumStalkerIterator * iterator,
+    GumStalkerOutput * output, gpointer user_data);
+static void bump_num_callouts (GumCpuContext * cpu_context,
+    gpointer user_data);
 static void unfollow_during_transform (GumStalkerIterator * iterator,
     GumStalkerOutput * output, gpointer user_data);
 static void test_invalidation_for_current_thread_with_target (GumAddress target,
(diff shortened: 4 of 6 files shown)


oleavr:
Thanks! :raised_hands:

PR #1117, opened 2026-06-09
control-flow graph: Refine public naming

oleavr:
Follow-up naming cleanup for the recently-landed control-flow graph API,
before any of it ships in a release.

- Rename the block/instruction lookups to `find_block_containing()` and
  `find_instruction_containing()`, and the JS `blockContaining()` to
  `findBlockContaining()`, matching the `find*` convention used for
  nullable lookups elsewhere in the API.
- Rename the dominating-site `window` to `capacity`, aligning with the
  existing `GumRedirectWriteDetails` terminology for how many bytes a
  redirect may occupy.

No released API is affected. All gum tests pass locally (QJS + V8 + core).


oleavr:
Landed on main via fast-forward as f04c3e3d (subject hyphenated to 'control-flow-graph:' before merge).

PR #1119, opened 2026-06-12
Add script cancellation support

tpetsas:
Add `gum_script_cancel()` API that allows graceful termination of a running script from another thread.

**Implementation details:**

- **QuickJS:** Uses the engine's interrupt handler (`JS_SetInterruptHandler()`) to poll a per-script cancellation flag (`is_cancelled`) protected by a recursive mutex.
- **V8:** We use `Isolate::TerminateExecution()` directly and skip exceptions via a try-catch block.

**Key changes summary:**
- Add cancel vfunc to `GumScriptInterface`
- Protect dispose/unload state transitions with a per-script `GRecMutex`
- Move `cancellation_mutex` cleanup from `destroy_context()` to `finalize()` to prevent use-after-clear during dispose
- Auto-dispose the script if cancelled during load
- Handle interrupted/terminated exceptions in entrypoint execution
- Include 3 unit tests (run both on QJS and V8)

**Limitations:**

Only interrupts JavaScript execution. If a script is blocked in native code (e.g., a `NativeFunction` call or a blocking syscall), cancellation will take effect once control returns to the JS engine.

----

To quickly test only the added unit tests, do the following:
**QJS:**
```
> build/tests/gum-tests -p '/GumJS/Script/script_can_be_cancelled#QJS' -p '/GumJS/Script/script_cancel_is_idempotent#QJS' -p '/GumJS/Script/script_cancel_during_load_should_auto_dispose#QJS'
```

**V8:**
```
> build/tests/gum-tests -p '/GumJS/Script/script_can_be_cancelled#V8' -p '/GumJS/Script/script_cancel_is_idempotent#V8' -p '/GumJS/Script/script_cancel_during_load_should_auto_dispose#V8'
```

### DIFF
--- a/bindings/gumjs/gumquickscript.c
+++ b/bindings/gumjs/gumquickscript.c
@@ -56,6 +56,8 @@ struct _GumQuickScript
   GumQuickScriptBackend * backend;
 
   GumScriptState state;
+  GRecMutex cancellation_mutex;
+  gboolean is_cancelled;
   GSList * on_unload;
   JSRuntime * rt;
   JSContext * ctx;
@@ -242,6 +244,12 @@ static void gum_quick_script_complete_unload_task (GumQuickScript * self,
 static void gum_quick_script_try_unload (GumQuickScript * self);
 static void gum_quick_script_once_unloaded (GumQuickScript * self,
     GumUnloadNotifyFunc func, gpointer data, GDestroyNotify data_destroy);
+static void gum_quick_script_cancel (GumScript * script);
+
+static int gum_cancellable_interrupt_handler (JSRuntime * runtime,
+    void * opaque);
+static void gum_quick_register_interrupt_handler (GumQuickScript * script);
+static void gum_quick_remove_interrupt_handler (GumQuickScript * script);
 
 static void gum_quick_script_set_message_handler (GumScript * script,
     GumScriptMessageHandler handler, gpointer data,
@@ -327,6 +335,7 @@ gum_quick_script_iface_init (gpointer g_iface,
   iface->unload = gum_quick_script_unload;
   iface->unload_finish = gum_quick_script_unload_finish;
   iface->unload_sync = gum_quick_script_unload_sync;
+  iface->cancel = gum_quick_script_cancel;
 
   iface->set_message_handler = gum_quick_script_set_message_handler;
   iface->post = gum_quick_script_post;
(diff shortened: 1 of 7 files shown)


bindings/gumjs/gumquickscript.c
```diff
@@ -1033,6 +1086,65 @@ gum_quick_script_once_unloaded (GumQuickScript * self,
...
+  self->is_cancelled = TRUE;
+  g_rec_mutex_unlock (&self->cancellation_mutex);
+}
+
+void
+_gum_quick_script_dispose_cancelled_script (GumQuickScript * self)
+{
+  g_rec_mutex_lock (&self->cancellation_mutex);
+  if (self->is_cancelled)
+    gum_quick_script_dispose ((GObject *) self);
```
oleavr, line 1104: This should ideally not be called by user code, only implicitly through reference counting.

oleavr (changes requested):
Wow, love this feature! Thanks!

~Do you think we could get rid of the cancellation mutex?~ If we can ~get rid of that, and~ avoid calling dispose, this should be ready to land.

oleavr:
Actually, please disregard the part about the cancellation mutex; the scope mutex will obviously be held so we cannot use that. So that leaves only the dispose nitpick :pray:

oleavr:
Since it took me so long to get to this, and that last thing is such a tiny issue I'll just fix it so we can land this. Thanks! :raised_hands:

oleavr:
Thanks so much for this, @tpetsas — really nice feature and a clean first cut. 🙌

I've landed it on main as 6df6525c, squashed and with some follow-up changes on top, keeping you as the commit author.

Summary of what changed from the PR while integrating:
- Split the single cancel() into two verbs: gum_script_interrupt() (abort the in-flight JS once, leaving the script loaded and runnable — e.g. REPL Ctrl+C) and gum_script_terminate() (abort and tear down via the regular async unload path).
- Disposal now flows through reference counting / unload rather than calling dispose() directly.
- interrupt() is a no-op while nothing is executing, so a request arriving just as an operation finishes can't abort the next one — execution is tracked at the scope boundary in both engines.
- V8 now lifts TerminateExecution()'s sticky state on the way out so execution stays resumable; QuickJS swallows the uncatchable interrupt centrally.

Closing this PR as landed — thanks again for the contribution!

tpetsas:
Thanks! 🙌 

Yeah dispose was kind of abrupt; the unload path is much cleaner! And about the "nothing is executing" case, I hadn't thought of that — good catch!

PR #1121, opened 2026-06-23
Add capstone 6 compatibility in gumdefs.h

kripticni:
Capstone 6 renamed CS_ARCH_ARM64 -> CS_ARCH_AARCH64.  This adds a compatibility
#define using the CS_API_MAJOR version macro so frida-gum builds correctly with
both capstone 5 and 6.

The #if CS_API_MAJOR >= 6 guard ensures the alias is only defined when needed,
and !defined(CS_ARCH_ARM64) prevents redefinition if the capstone headers
already provide the backward-compatible alias.

### DIFF
--- a/gum/gumdefs.h
+++ b/gum/gumdefs.h
@@ -9,8 +9,13 @@
 #ifndef __GUMDEFS_H__
 #define __GUMDEFS_H__
 
+#include <capstone.h>
 #include <glib.h>
 #include <gum/gumenumtypes.h>
+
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks, @kripticni! 🙏🏻 Merged 🙌🏻

Tweaked two things first: moved the #include in with the others alphabetically, and dropped the !defined guard since CS_API_MAJOR >= 6 already implies the old name is gone.

PR #1129, opened 2026-07-06
Add missing locking to GumDarwinSymbolicator

comex:
Fixes a crash I encountered when symbolicating backtraces.

Symbolication is supposed to be thread-safe (as demonstrated in many other places, e.g. by the locking in `gum_try_obtain_symbolicator`, or by the fact that `gumjs_symbol_from_address` unlocks the JS engine while calling `gum_symbol_details_from_address`).

However, GumDarwinSymbolicator is not thread-safe.  It can perform write operations while looking up symbols, due to lazy initialization of (1) `objc_resolver`, (2) the module map, (3) the functions list for each module.  But there is no locking around any of those.

Fix by adding a read-write lock.

I considered also adding locking to the get_property/set_property implementations and `gum_darwin_symbolicator_load`.  But I decided against it since I'm not sure whether these are idiomatically expected to be thread-safe.  Regardless, it shouldn't make a difference in practice (as far as I can tell), since set_property and load are only called during initialization before multiple threads can have access to the object.

Also, drive-by fixes:

- Fix an overlong line in `gummoduleregistry-darwin.c` which `stylecheck.py` was complaining about.

- In `gum_darwin_symbolicator_synthesize_details_from_address`, initialize `module` to `NULL` to avoid an uninitialized load if we hit the first `goto beach`.

### DIFF
--- a/gum/backend-darwin/gumdarwinsymbolicator.c
+++ b/gum/backend-darwin/gumdarwinsymbolicator.c
@@ -55,6 +55,7 @@ struct _GumDarwinSymbolicator
 
   CSSymbolicatorRef handle;
 
+  GRWLock rwlock; /* guards following fields */
   GumApiResolver * objc_resolver;
   GumModuleMap * modules;
   GHashTable * functions;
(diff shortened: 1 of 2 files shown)


oleavr:
Nice one — you nailed the race in GumDarwinSymbolicator's lazy init. 🙏

Landed on main, shipping in 17.15.4. Split into three commits (locking / module init / overlong line), authorship preserved. Thanks!

PR #1131, opened 2026-07-09
Fix gum_memory_patch_code deadlock when threads are suspended in alloc

hsorbo:
On the non-RWX path, `gum_memory_patch_code_pages` suspends every other thread and then runs the apply callback while they're frozen. If a suspended thread is holding alloc lock and the callback allocates, the callback's allocation blocks on that lock forever — a permanent deadlock.

This fixes it by applying the patch into a scratch buffer first, where apply is free to allocate, and only then suspending threads — just long enough to copy the changed bytes back into place, lock-free.

Adds a regression test that reproduces the deadlock: a helper thread holds a lock across the patch while the apply callback contends for that same lock, standing in for a thread frozen mid-allocation.

Since the fix would push `gum_memory_patch_code_pages` past ~400 lines, I first split the three strategies into their own functions:
  - `gum_memory_patch_code_pages_via_remap`
  - `gum_memory_patch_code_pages_via_mprotect`
  - `gum_memory_patch_code_pages_via_code_segment`

### DIFF
--- a/gum/gummemory.c
+++ b/gum/gummemory.c
@@ -623,16 +623,15 @@ gum_memory_patch_code_pages_via_mprotect (GPtrArray * sorted_addresses,
   guint i;
   guint8 * apply_start, * apply_target_start;
   guint apply_num_pages;
-  GumPageProtection protection;
-  GumPageProtection * original_protections;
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! 🙌

PR #1132, opened 2026-07-10
Re-enable CodeSegment on newer XNUs

hsorbo:
The vm_remap OVERWRITE trick that ef6f67bb disabled for kernels >= 8020.142.0 (iOS >= 15.6.1) works again on newer iOS and macOS versions. My lowest test device is ipadOS 17.6 so that was chosen as cut-off. Test suite also runs on macOS 26.5.1

tested:
macOS 26.5.1: full suite.
ipadOS 17.6 (palera1n):  `/Core/Interceptor` and  `/Core/MemoryAccessMonitor`


oleavr:
Thanks! 🙌

Issue #1130, opened 2026-07-08
module-removed trigger too late for reverting implementations


SamSunNV:
## Summary

I've encountered a problem with a specific use case with Frida Gum: In my project, I want to hook certain libraries within the module-add trigger and revert them in module-removed to clear internal tracked states within that module. But module-removed is triggered after the module is unmapped from memory. gum_interceptor_revert will try replacing the implementation back to its original one. And during replacing, the mprotect call will fail since the memory is already unmapped. This makes me not able to call gum_interceptor_revert, leaving an internal tracked state after the module is unloaded. In the future, when a module is reloaded into that section of memory, it's likely that it will trigger a GUM_REPLACE_ALREADY_REPLACED error, even though the newly loaded module's implementation was not replaced.

Therefore, I think it would be better if we have an exposed API to clear the internal states or remove all the tracked states of a module when the module is removed.

## Suggested Fix

It would be ideal that we could have an API that allows clearing internal states of the module, or automatically clear them when the module is unloaded.

## Reproduction

I've made a small demo to reproduce the crash when calling gum_interceptor_revert in module-remove:

https://github.com/SamSunNV/frida-unload-crash-reproduce

PR #1134, opened 2026-07-14
module-registry: discard function contexts when module unloads

SamSunNV:
Discard function context when unloading modules.

Added a global registry for interceptor instances.

Added test that tests if the unmapped code will be touched during interceptor destruction.


SamSunNV:
#1130

oleavr:
Thanks! 🙌

PR #1139, opened 2026-08-04
arm64: Fix Interceptor crash on functions with internal CBZ targets

WHW0x455:
Hooking short ARM64 functions whose prologue contains a conditional branch into the same 16-byte window (notably
`dispatch_mach_msg_get_msg` in libdispatch, macOS) can crash the target with`EXC_BAD_INSTRUCTION` / `SIGILL`.

On macOS, attaching Interceptor to `dispatch_mach_msg_get_msg` and exercising the `x1 == NULL` path crashes, 
e.g. in process `lsd`:

 ```
 exception: EXC_BAD_INSTRUCTION
 frame:     dispatch_mach_msg_get_msg + 12
 x1 = 0
```

The function looks like:

```asm
dispatch_mach_msg_get_msg:
    cbz  x1, +12
    ldr  x8, [x0, #0x50]
    str  x8, [x1]
    ldr  w8, [x0, #0x48]   ; +12
    add  x0, x0, #0x58
    ...
    ret
```

Root cause:

 1. Interceptor prefers a 16-byte full redirect when
    gum_arm64_relocator_can_relocate(..., 16, CHECKED, ...) succeeds.
 2. Relocated cbz/tbz/b.cond are rewritten as absolute jumps back to
    the original target address.
 3. With x1 == NULL, the rewritten CBZ jumps to function+12, which
    sits inside the 16-byte patch (part of the absolute address literal)
    → illegal instruction.

 GUM_RELOCATION_CHECKED was supposed to prevent branches into the
 middle of the patch, but it only started its reachability scan at
 rl.input_cur (after the bytes already chosen for relocation). It never
 inspected branches inside the relocated instructions themselves, so
 the leading cbz x1, +12 was invisible and can_relocate(16) returned
 true.

 Fix:

 In gum_arm64_relocator_can_relocate() under GUM_RELOCATION_CHECKED:

 - Scan already-read instructions for B/CBZ/TBZ targets.
 - If a target falls inside (0, n), shrink n so the target is outside
   the patch; queue out-of-range targets for the existing walk.
 - Rewind rl.input_cur / rl.input_pc to the new boundary before the
   post-range reachability scan.
 - Factor branch-target extraction into
   gum_arm64_relocator_extract_branch_target().

 For dispatch_mach_msg_get_msg, can_relocate(16) now fails with
 maximum == 12, and Interceptor falls back to an 8-byte near redirect
 that leaves +12 intact.

Testcase cannot_relocate_with_internal_cbz_target is added for covering this prologue shape.

### DIFF
--- a/gum/arch-arm64/gumarm64relocator.c
+++ b/gum/arch-arm64/gumarm64relocator.c
@@ -24,6 +24,7 @@ struct _GumCodeGenCtx
 
 static gboolean gum_arm64_relocator_register_is_free (
     const GumArm64Relocator * self, guint n, arm64_reg reg);
+static gpointer gum_arm64_relocator_extract_branch_target (const cs_insn * insn);
 
 static gboolean gum_arm64_branch_is_unconditional (const cs_insn * insn);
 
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! 🙌  Landed in ddc10c5559cbb41a3dd72866bfba6ff3945ffa5c with a few minor tweaks.

PR #1141, opened 2026-08-08
elf-module: Handle relocated program headers

tracyliving:
Read program headers using file-coordinate data, while preserving bounded live-memory fallback when no backing file is available.

Add focused Linux x86-64 regressions for relocated program headers and fallback bounds.


oleavr:
Merged as 8a54c262 — thanks!

Right diagnosis. The unbounded live-data fallback was the easy half to miss.

Tidied on the way in: one guard macro instead of five copies, the synthetic tests now run on any Linux, and the two must-not-open paths got a name that says so.

The registry guessed the load bias the same way — fixed in 2448b0bb and friends.

PR #1145, opened 2026-08-21
Move cmodule api to gum

cputnam-a11y:
Move the cmodule api in gumcmodule.h and gumcmodule.c to the gum subproject, allowing it's consumers (me) to use it without depending on gumjs, as it doesn't really depend on js at all, and can be useful even when consuming the c api.

I'm sure I've missed some things as this is my first time working with meson and cpp in a larger project, so I welcome any guidance as for what I've missed


oleavr:
Thanks! 🙌

PR #1146, opened 2026-08-26
Modify gum_pointer_scan_process_tile to use GumExceptor

IPMegladon:
The new `gum_memory_find_pointers()` / `Memory.findPointers()` is great and exactly what I needed. 
However without a GumExceptor like `Memory.scan` it becomes extremely unstable when scanning across large ranges like `Process.enumerateRanges('rw')`.

Thus had the clanker add a GumExceptor to gum_pointer_scan_process_tile.


oleavr:
Thanks! 🙌

PR #1148, opened 2026-08-31
arm64: Dodge BTI where we have no landing pad

### LINKED ISSUE #1147: arm64 Linux kernel-module agent: Interceptor.attach panics with Oops - BTI (bti fix incomplete for far trampolines)
## Summary

On an arm64 kernel built with `CONFIG_ARM64_BTI_KERNEL=y` (+ `CONFIG_ARM64_PTR_AUTH_KERNEL=y`), the **Linux kernel-module** flavour of the barebone agent (`frida-agent.ko`) panics with `Oops - BTI` as soon as `Interceptor.attach()` activates a hook. The upstream BTI fix from #731 / #735 only covers one of several indirect branches Gum generates, and the kernel-module case trips a path those two PRs did not need to fix.

## Environment

- Kernel: `5.15.180-android13` (OnePlus SM8550, Android GKI), `CONFIG_ARM64_BTI_KERNEL=y`, `CONFIG_ARM64_PTR_AUTH_KERNEL=y`
- frida-core / frida-gum @ `17.17.0`
- Agent: `frida-core/src/barebone/agent/linux` (the `/dev/frida` kernel module), host = `frida-server --device barebone`

## Repro

```js
Interceptor.attach(DebugSymbol.getFunctionByName('__arm64_sys_getuid'), {
  onEnter() { console.log('enter'); }
});
```

Kernel logs:

```
Internal error: Oops - BTI: 0000000036000001 [#1] PREEMPT SMP
Modules linked in: frida_agent(OE) ...
Call trace:
 0xffffffc03731d000            <- vmalloc, Gum's trampoline (no symbol)
 ...
Kernel panic - not syncing: Oops - BTI: Fatal exception
```

## Root cause

#731/#735 changed only `gum_emit_epilog()` to use `ret` instead of `br` when `HAVE_PTRAUTH` is unset. That is sufficient for the userland case in frida/frida#2389, where the trampoline is allocated within ±128 MB of the target and the redirect therefore uses `put_b_imm` (a direct branch, which BTI does not inspect).

The kernel-module case is different: `gum_code_allocator` places the trampoline in the vmalloc region (`0xffffffc0…`) while the hooked kernel function sits in the module region (`0xffffffd1…`), ~73 GB apart — far beyond the ±128 MB reach of `b`/`bl`. The redirect therefore falls into `guminterceptor-arm64.c` `_gum_interceptor_backend_activate_trampoline()`'s `redirect_code_size == 8` / `GUM_INTERCEPTOR_FULL_REDIRECT_SIZE` cases, which emit `br x16` to the trampoline. The trampoline entry has no `bti` landing pad, so the `br` faults.

The same applies to the other runtime-generated indirect branches that target dynamic code: the enter/leave thunks, the deflector dispatchers in `gumcodeallocator.c`, the relocator's rewritten `b`/`cbz`/`tbz`, and Stalker's block hops.

## Suggested fix

Generalise the #731/#735 approach at the writer layer: when pointer authentication is unsupported, make `gum_arm64_writer_put_br_reg()` / `put_br_reg_no_auth()` emit `ret xN` instead of `br xN`. `ret` performs the identical unconditional jump but is exempt from the BTI check, and on non-PTRAUTH kernels it does not authenticate. This single change covers every indirect branch Gum generates instead of patching ~25 call sites individually.

```c
gboolean
gum_arm64_writer_put_br_reg (GumArm64Writer * self, arm64_reg reg)
{
  if (self->ptrauth_support == GUM_PTRAUTH_SUPPORTED)
    return gum_arm64_writer_put_br_reg_with_extra (self, reg, 0x81f);

  return gum_arm64_writer_put_ret_reg (self, reg);
}
```

(and the analogous change for `put_br_reg_no_auth`). Darwin arm64e keeps the existing behaviour because `ptrauth_support == GUM_PTRAUTH_SUPPORTED` there.

I have a PR ready with this change; happy to open it if this approach is acceptable.

inforcqb:
## What

Stop `br` from trapping on BTI-guarded pages, so the runtime-generated
trampolines and thunks work on kernels built with `CONFIG_ARM64_BTI_KERNEL=y`.

Fixes #1147.

## Why

`br xN` faults with `Oops - BTI` unless the target begins with a `bti` landing
pad. Gum emits its trampolines, thunks and relocated code at run time without
one, and on the Linux kernel-module flavour the trampoline lives in vmalloc
while the hooked function lives in the module region (~73 GB apart), so the
redirect cannot use a direct `b` and falls back to `br`.

## Change

Split the problem by who owns the branch target.

**Targets that are ours** — the enter and leave thunks get a `bti` landing pad
and keep `br`. This is the same shape as the x86 side, where
`gum_x86_writer_put_endbr()` is a plain primitive and the backend decides where
pads go.

**Targets that aren't ours** — use `ret`, which is exempt from the BTI check,
for the branches Gum cannot annotate: resuming a hooked function, the
relocator's rewritten `b`/`b.cond`/`cbz`/`tbz`, the epilog's next hop, the
activation redirect (which for `GUM_INTERCEPTOR_TYPE_FAST` targets the caller's
replacement function), and the code deflector's dispatcher and thunk. This
generalises the trick `gum_emit_epilog()` already used behind `#ifndef
HAVE_PTRAUTH`, which now folds into the new API.

The writer stays a faithful assembler: `put_br_reg()` and `put_br_reg_no_auth()`
emit `br`, and callers opt into the exemption through a separate operation-level
pair.

```c
/* mnemonic layer */                  /* operation layer */
put_br_reg (aw, reg);                 put_jmp_reg (aw, reg);
put_br_reg_no_auth (aw, reg);         put_jmp_reg_no_auth (aw, reg);
```

`put_jmp_reg*` mean "get to this address by whatever survives BTI" — `br` where
pointer authentication is available, `ret` otherwise — so `_no_auth` stays a
single orthogonal modifier. Darwin arm64e is unchanged.

## Not covered

Stalker's block dispatch still uses `br`. Its targets are mostly its own blocks,
so it wants landing pads at block entries rather than the `ret` treatment:
routing its hot path through `ret` would pop the return-address stack on every
block transition and mispredict real returns process-wide. That touches slab
layout and `GUM_RESTORATION_PROLOG_SIZE`, so it is left for a follow-up, and a
kernel-flavour Stalker still trips BTI.

## Testing

`Arm64Writer` covers the new `put_jmp_reg*` and `put_bti`, and still asserts
that `put_br_reg` emits `br`. The `Arm64Relocator` expectations were updated by
hand from `br x16` to `ret x16`; they are gated behind `cs_support
(CS_ARCH_ARM64)` and need arm64 CI to confirm. Nothing here has been exercised
on a BTI-enforcing kernel yet.

### DIFF
--- a/gum/arch-arm64/gumarm64relocator.c
+++ b/gum/arch-arm64/gumarm64relocator.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2014-2026 Ole André Vadla Ravnås <oleavr@nowsecure.com>
  * Copyright (C) 2026 Haiwei Wang <haiwei.wang1109@gmail.com>
+ * Copyright (C) 2026 inforcqb <fanjiawei080615@qq.com>
  *
  * Licence: wxWindows Library Licence, Version 3.1
  */
@@ -758,7 +759,7 @@ gum_arm64_relocator_rewrite_b (GumArm64Relocator * self,
 
   gum_arm64_writer_put_ldr_reg_address (ctx->output, ARM64_REG_X16,
       gum_arm64_writer_sign (ctx->output, target->imm));
-  gum_arm64_writer_put_br_reg (ctx->output, ARM64_REG_X16);
+  gum_arm64_writer_put_jmp_reg (ctx->output, ARM64_REG_X16);
 
   return TRUE;
 }
@@ -778,7 +779,7 @@ gum_arm64_relocator_rewrite_b_cond (GumArm64Relocator * self,
   gum_arm64_writer_put_label (ctx->output, is_true);
   gum_arm64_writer_put_ldr_reg_address (ctx->output, ARM64_REG_X16,
       gum_arm64_writer_sign (ctx->output, target->imm));
-  gum_arm64_writer_put_br_reg (ctx->output, ARM64_REG_X16);
+  gum_arm64_writer_put_jmp_reg (ctx->output, ARM64_REG_X16);
 
   gum_arm64_writer_put_label (ctx->output, is_false);
 
@@ -817,7 +818,7 @@ gum_arm64_relocator_rewrite_cbz (GumArm64Relocator * self,
   gum_arm64_writer_put_label (ctx->output, is_true);
   gum_arm64_writer_put_ldr_reg_address (ctx->output, ARM64_REG_X16,
       gum_arm64_writer_sign (ctx->output, target->imm));
-  gum_arm64_writer_put_br_reg (ctx->output, ARM64_REG_X16);
+  gum_arm64_writer_put_jmp_reg (ctx->output, ARM64_REG_X16);
 
   gum_arm64_writer_put_label (ctx->output, is_false);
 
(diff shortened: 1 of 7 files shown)


oleavr:
Thanks! 🙌

PR #1149, opened 2026-09-02
interceptor: Don't reap live frames across stackful coroutines

### LINKED ISSUE #1142: Interceptor: live invocation-stack entry reaped under stackful coroutines → SIGSEGV in _gum_function_context_end_invocation
## Summary

On a target whose scripting runtime uses **stackful coroutines / green threads** (multiple logical call stacks multiplexed onto one OS thread, e.g. `swapcontext`/`makecontext`-style fiber schedulers), attaching an `onLeave` (or `Interceptor.replace`) to a native function that is entered and returns across a coroutine suspend/resume leads to a **probabilistic `SIGSEGV`** inside the Interceptor machinery.

The crashing frame (#0) is `_gum_function_context_end_invocation` (`gum/guminterceptor.c`), at:

```c
stack_entry = gum_invocation_stack_peek_top (interceptor_ctx->stack);
*next_hop = gum_sign_code_pointer (stack_entry->caller_ret_addr);   // <-- fault
```

`stack_entry` is a wild pointer because the thread's invocation stack is **empty** at this point, yet the value is dereferenced unconditionally.

## Environment

- frida-gum 17.17.0 (the relevant code is also present on `main`).
- Reproduced on Android arm64, embedded as Gadget. The analysis is platform-independent.
- Signal: `SIGSEGV / SEGV_MAPERR`, fault address is a large out-of-bounds pointer (not a small NULL+offset).

## Root cause

17.x added two "reap" steps that assume calls nest strictly by **descending stack address** on a single OS thread:

1. `gum_function_context_begin_invocation()` calls `gum_invocation_stack_reap_unwound(stack, live_sp)` on every on-enter, popping every entry whose `stack_address < live_sp` (treated as "unwound past").
2. `_gum_function_context_end_invocation()` calls `gum_invocation_stack_reap_unwound_above(stack, returning_ctx)`.

With stackful coroutines this assumption breaks. Consider one OS thread running two fibers whose stacks live at **unrelated** addresses:

- Fiber A (low-address stack) enters hooked function `F` → entry `E_F` pushed with `stack_address = SP_A`.
- The scheduler suspends A and resumes fiber B (higher-address stack). B enters hooked function `G` at `SP_B` where `SP_B > SP_A`.
- `begin_invocation` for `G` runs `reap_unwound(stack, SP_B)`. Since `SP_A < SP_B`, the **still-live** entry `E_F` is misclassified as unwound and popped.
- Later A resumes and `F` returns; its on-leave trampoline fires, but `E_F` is gone. The thread's invocation stack may now be **empty**, so `gum_invocation_stack_peek_top()` returns `NULL`.

Because `_gum_function_context_end_invocation` dereferences the peek result unconditionally, the compiler treats the `if (stack->len == 0) return NULL;` inside `gum_invocation_stack_peek_top` as dead (dereferencing NULL is UB), and eliminates it. The inlined peek then computes `&g_array_index(stack, GumInvocationStackEntry, (guint)(len - 1))` with `len == 0`, i.e. index `0xFFFFFFFF`, producing a wild pointer → `SIGSEGV` on the `caller_ret_addr` load.

Disassembly (arm64, inlined peek at the fault) confirms: `len` is loaded, `sub len,#1` (no zero check), `&data[len-1]` computed via `umaddl`, then `ldr x0,[x_entry,#8]` faults.

Pre-17.x Gum did not have these reap steps and did not crash in this scenario (it simply peeked the current top).

## Reproduction (conceptual)

1. In a process using a runtime with stackful coroutines/green threads (fiber scheduler that swaps the C stack on the same OS thread).
2. `Interceptor.attach(fn, { onLeave() { ... } })` (or `Interceptor.replace`) on a native function that is entered on one fiber and returns after a suspend/resume, so its frame is "crossed" by another hooked call on a higher-address fiber stack.
3. Under load the process crashes probabilistically in `_gum_function_context_end_invocation`.

## Suggested fixes (any one addresses the crash)

1. **Guard the empty/mismatch case** in `_gum_function_context_end_invocation`: if `gum_invocation_stack_peek_top()` returns `NULL` (or the top's `function_ctx != function_ctx`), do not dereference; recover gracefully instead of relying on UB. This also makes `gum_invocation_stack_peek_top`'s NULL guard actually effective (currently DCE'd away by the unconditional caller deref).
2. **Don't reap purely by stack-pointer comparison.** Since fiber stacks are not monotonic across coroutines, `reap_unwound` should not treat `stack_address < live_sp` as proof of unwinding. Reap only entries that are provably dead (e.g. require the returning `function_ctx` to be present, and never empty a non-empty stack that still holds unrelated live frames).
3. Restoring the classic non-reaping behavior fixes stackful-coroutine workloads; a targeted guard is preferable so exception/`longjmp` cleanup still works.

I'm happy to send a PR — please advise which direction you'd prefer (defensive guard in `end_invocation` vs. reworking the reap heuristic).

wertyutreethgfd:
## Summary

Fixes #1142.

On a target that multiplexes several C stacks onto one OS thread (`swapcontext` / fibers / stackful coroutines), Interceptor 17.x can `SIGSEGV` in `_gum_function_context_end_invocation` because reap treated `stack_address < live_sp` as proof a frame was unwound — including a still-live frame on another C stack.

## Approach

Each invocation-stack frame records the **memory range of its own C stack** at enter:

1. `gum_thread_try_get_ranges()` when the SP is on the pthread / Darwin thread stack / current Windows fiber (`GetCurrentThreadStackLimits`).
2. Otherwise the RW mapping that contains the SP (`_gum_process_enumerate_ranges`).

Reap and leave matching are **per range**:

- **Enter:** only SP-compare frames that share the live range. Frames on other stacks are skipped, even if they sit at a lower address or above us in the mixed invocation array.
- **Leave:** find the returning `function_ctx` **on that same range** (two fibers can call the same hooked function). Reap only intervening frames on that range (exception / `longjmp` on that C stack). `onLeave` runs on the matching frame; it is removed by index, so another fiber may remain stacked above it.
- **Current invocation / depth / replacement bypass** look at the top frame of the *current* C stack, not the mixed array top.

If the matching frame is missing, `next_hop` is not taken from a wild `peek_top()` (that path is UB and compilers delete the NULL check). That is an invariant failure, not a substitute for identifying the stack.

If the OS cannot describe any mapping (bare metal), the address space is one stack — the correct model when there is only one C stack, not a return to the old cross-stack SP compare.

## Tests

- `longjmp_reaps_skipped_leave` — thread stack: inner `onLeave` skipped by `longjmp` is reaped when the outer frame returns.
- `stackful_coroutine_does_not_reap_live_frames` — Linux, two `mmap` fibers, different hooked functions.
- `stackful_coroutine_same_function` — two fibers enter/leave the **same** hooked function; leave must not bind the other fiber's frame.
- `stackful_coroutine_longjmp_reaps_on_fiber` — `longjmp` on fiber B reaps B's inner frame and leaves fiber A intact.

### DIFF
--- a/gum/guminterceptor.c
+++ b/gum/guminterceptor.c
@@ -232,14 +232,17 @@ static GumInvocationStackEntry * gum_invocation_stack_push (
     gpointer caller_ret_addr, gpointer stack_address,
     gboolean only_invoke_unignorable_listeners);
 static gpointer gum_invocation_stack_pop (GumInvocationStack * stack);
+static void gum_invocation_stack_remove_entry (GumInvocationStack * stack,
+    GumFunctionContext * function_ctx, gpointer stack_address);
 static void gum_invocation_stack_reap_unwound (GumInvocationStack * stack,
     gpointer live_stack_address);
-static void gum_invocation_stack_reap_unwound_above (
+static GumInvocationStackEntry * gum_invocation_stack_reap_unwound_above (
     GumInvocationStack * stack, GumFunctionContext * returning_ctx);
 static void gum_invocation_stack_entry_snapshot_cpu_context (
     GumInvocationStackEntry * entry, const GumCpuContext * cpu_context);
 static gboolean gum_invocation_stack_entry_was_unwound_past (
-    const GumInvocationStackEntry * entry, gpointer live_stack_address);
+    const GumInvocationStackEntry * entry, gpointer live_stack_address,
+    const GumMemoryRange * ranges, guint n_ranges);
 static void gum_invocation_stack_entry_release_trampoline (
     const GumInvocationStackEntry * entry);
 static GumInvocationStackEntry * gum_invocation_stack_peek_top (
@@ -2143,6 +2146,7 @@ _gum_function_context_end_invocation (GumFunctionContext * function_ctx,
   InterceptorThreadContext * interceptor_ctx;
   GumInvocationStackEntry * stack_entry;
   GumInvocationContext * invocation_ctx;
+  gpointer stack_address;
   GPtrArray * listener_entries;
   gboolean only_invoke_unignorable_listeners;
   guint i;
@@ -2159,11 +2163,22 @@ _gum_function_context_end_invocation (GumFunctionContext * function_ctx,
 
   interceptor_ctx = get_interceptor_thread_context ();
 
-  gum_invocation_stack_reap_unwound_above (interceptor_ctx->stack,
+  stack_entry = gum_invocation_stack_reap_unwound_above (interceptor_ctx->stack,
       function_ctx);
+  if (G_UNLIKELY (stack_entry == NULL))
+  {
+    /*
+     * The enter-side frame is gone, so the original return address is too.
+     * Never peek an empty invocation stack: that is UB, and compilers
+     * delete peek_top()'s NULL check, turning len==0 into a wild pointer.
+     */
+    *next_hop = NULL;
+    gum_tls_key_set_value (gum_interceptor_guard_key, NULL);
+    return;
+  }
 
-  stack_entry = gum_invocation_stack_peek_top (interceptor_ctx->stack);
   *next_hop = gum_sign_code_pointer (stack_entry->caller_ret_addr);
+  stack_address = stack_entry->stack_address;
 
   invocation_ctx = &stack_entry->invocation_context;
   invocation_ctx->cpu_context = cpu_context;
@@ -2211,7 +2226,8 @@ _gum_function_context_end_invocation (GumFunctionContext * function_ctx,
 
   gum_thread_set_system_error (invocation_ctx->system_error);
 
-  gum_invocation_stack_pop (interceptor_ctx->stack);
+  gum_invocation_stack_remove_entry (interceptor_ctx->stack, function_ctx,
+      stack_address);
 
   gum_tls_key_set_value (gum_interceptor_guard_key, NULL);
 
@@ -2529,9 +2545,36 @@ gum_invocation_stack_pop (GumInvocationStack * stack)
 }
 
 static void
+gum_invocation_stack_remove_entry (GumInvocationStack * stack,
+                                   GumFunctionContext * function_ctx,
+                                   gpointer stack_address)
+{
+  guint i;
+
+  for (i = stack->len; i != 0; i--)
+  {
+    GumInvocationStackEntry * entry;
+
+    entry = (GumInvocationStackEntry *)
+        &g_array_index (stack, GumInvocationStackEntry, i - 1);
+    if (entry->function_ctx == function_ctx &&
+        entry->stack_address == stack_address)
+    {
+      g_array_remove_index (stack, i - 1);
+      return;
+    }
+  }
+}
+
+static void
 gum_invocation_stack_reap_unwound (GumInvocationStack * stack,
                                    gpointer live_stack_address)
 {
+  GumMemoryRange ranges[2];
+  guint n_ranges;
+
+  n_ranges = gum_thread_try_get_ranges (ranges, G_N_ELEMENTS (ranges));
+
   while (stack->len != 0)
   {
     GumInvocationStackEntry * entry;
@@ -2539,7 +2582,7 @@ gum_invocation_stack_reap_unwound (GumInvocationStack * stack,
     entry = (GumInvocationStackEntry *)
         &g_array_index (stack, GumInvocationStackEntry, stack->len - 1);
     if (!gum_invocation_stack_entry_was_unwound_past (entry,
-        live_stack_address))
+        live_stack_address, ranges, n_ranges))
       break;
 
     gum_invocation_stack_entry_release_trampoline (entry);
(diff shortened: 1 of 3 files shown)


oleavr:
Thanks for this — the diagnosis and the tests are good, and I've kept you as
author of the interceptor commit. I've reworked the branch and pushed it here;
rebased onto current `main` too, since the old base predated the `G_OS_NONE`
guard around the unwind broker and would have reverted it.

What changed, and why:

**Identify a stack by the top of its region, not the whole range.** Stacks grow
downwards, so the range's base moves — Linux grows the main thread's `[stack]`
VMA, and Windows commits fiber stack pages one at a time. If that happened
between a hook's enter and its leave, the frame no longer matched, and the
`*next_hop = NULL` bail-out that followed is a jump to NULL. Keying on the top
is stable, and it also removes the need to pin the pthread ranges.

**Don't query the memory map on every call.** `gum_thread_try_get_ranges()` ran
unconditionally on every enter, leave and peek, ahead of the cache — on Linux
that's `pthread_getattr_np()` plus a `sigaltstack()` syscall per hooked call.
There's now a `gum_memory_query_region()` doing a real point query
(`VirtualQuery` / `PROC_PIDREGIONINFO` / `PROCMAP_QUERY`), consulted only on a
cache miss.

**Removed the `NULL` bail-out** and replaced the search-by-`(function_ctx,
stack_address)` removal with the index the reap already knows.

**On Go**, which I'm guessing is what you're aiming at: this won't separate
goroutines. Go carves goroutine stacks out of shared, growing heap arena spans —
I measured 200 goroutines landing in 11 regions, 93 of them in one, with a given
base reported at different sizes as the arena is carved. So leave-time matching
now falls back to the returning function alone when no frame is found on the
current C stack, which is what `main` does today: Go ends up safe, but not
supported. Doing it properly needs identity from the runtime rather than the
memory map.

One thing I could not do: the three `stackful_coroutine_*` tests are Linux-only
(`makecontext`/`swapcontext` don't exist on macOS/arm64), so they have never
been run. If you can confirm they pass on Linux, that would help.

wertyutreethgfd:
The three `stackful_coroutine_*` tests pass on Linux, and so does `longjmp_reaps_skipped_leave`. Ran on commit `26f08c3af5a7ec3893e72c8483bf3375c81b4c89` (`interceptor: Track which C stack each frame is on`), Ubuntu 24.04, Linux 6.8.0-62-generic x86_64, glibc 2.39. Build was `meson setup build -Dtests=enabled && ninja -C build`. All four were registered (`GUM_TEST_HAVE_FIBERS`), then each was run on its own with `-p`:

```
./build/tests/gum-tests -p /Core/Interceptor/longjmp_reaps_skipped_leave
# random seed: R02S7847ca88f6797e57581b76cae16ce7b6
# Start of Core tests
# Start of Interceptor tests
ok 1 /Core/Interceptor/longjmp_reaps_skipped_leave
# End of Interceptor tests
# End of Core tests
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_does_not_reap_live_frames
# random seed: R02Sde315201af25c3b79dc3a1965f9bc374
ok 1 /Core/Interceptor/stackful_coroutine_does_not_reap_live_frames
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_same_function
# random seed: R02S630c3191dac43db598f28d9a6d108c6d
ok 1 /Core/Interceptor/stackful_coroutine_same_function
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_longjmp_reaps_on_fiber
# random seed: R02Se244521b9799da6b53825659e545939d
ok 1 /Core/Interceptor/stackful_coroutine_longjmp_reaps_on_fiber
1..1
exit:0
```

`gumtest.c` also prints `Ran 448 tests in 0.00 seconds` after each of these. That number is `get_number_of_tests_in_suite()` of the registered suite, not the filter. The TAP plan above is what actually ran.

This is not a Go target. The process runs a scripting runtime with stackful coroutines — Python and Lua fibers — multiplexed onto one OS thread. Each fiber has its own C stack and is switched with a `swapcontext`-style scheduler, and those stacks sit in separate mappings. Identifying a stack by the top of the region that holds it is the case we need. Goroutines carved out of one shared, growing heap arena are not what we are instrumenting, so the leave-time fallback to the returning function alone is not the path this workload hits.

wertyutreethgfd:
The three `stackful_coroutine_*` tests pass on Linux, and so does `longjmp_reaps_skipped_leave`. Ran on commit `26f08c3af5a7ec3893e72c8483bf3375c81b4c89` (`interceptor: Track which C stack each frame is on`), Ubuntu 24.04, Linux 6.8.0-62-generic x86_64, glibc 2.39. Build was `meson setup build -Dtests=enabled && ninja -C build`. All four were registered (`GUM_TEST_HAVE_FIBERS`), then each was run on its own with `-p`:

```
./build/tests/gum-tests -p /Core/Interceptor/longjmp_reaps_skipped_leave
# random seed: R02S7847ca88f6797e57581b76cae16ce7b6
# Start of Core tests
# Start of Interceptor tests
ok 1 /Core/Interceptor/longjmp_reaps_skipped_leave
# End of Interceptor tests
# End of Core tests
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_does_not_reap_live_frames
# random seed: R02Sde315201af25c3b79dc3a1965f9bc374
ok 1 /Core/Interceptor/stackful_coroutine_does_not_reap_live_frames
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_same_function
# random seed: R02S630c3191dac43db598f28d9a6d108c6d
ok 1 /Core/Interceptor/stackful_coroutine_same_function
1..1
exit:0

./build/tests/gum-tests -p /Core/Interceptor/stackful_coroutine_longjmp_reaps_on_fiber
# random seed: R02Se244521b9799da6b53825659e545939d
ok 1 /Core/Interceptor/stackful_coroutine_longjmp_reaps_on_fiber
1..1
exit:0
```

`gumtest.c` also prints `Ran 448 tests in 0.00 seconds` after each of these. That number is `get_number_of_tests_in_suite()` of the registered suite, not the filter. The TAP plan above is what actually ran.

This is not a Go target. The process runs a scripting runtime with stackful coroutines — Python and Lua fibers — multiplexed onto one OS thread. Each fiber has its own C stack and is switched with a `swapcontext`-style scheduler, and those stacks sit in separate mappings. Identifying a stack by the top of the region that holds it is the case we need. Goroutines carved out of one shared, growing heap arena are not what we are instrumenting, so the leave-time fallback to the returning function alone is not the path this workload hits.

PR #1150, opened 2026-09-11
darwin: Use jailbreak-provided memory hooks

devnoname120:
Dopamine monkey-patches Frida's syscall instructions to route executable-memory changes through hookd. This workaround is broken with current Frida on iOS 26, where Frida operations can crash the iPhone and force a reboot.

Use an optional, versioned memory provider for explicit hookd calls during initialization, code patching, and VM protection. This removes the dependency on Frida's instruction layout and avoids debugger page-plan traps; existing behavior is retained without a provider.

Together with opa334/Dopamine#1000, addresses opa334/Dopamine#938, opa334/Dopamine#979, opa334/Dopamine#985 and frida/frida#3773.

Validated on an iPhone 11/iOS 26.0 standalone target: attach, patch, Interceptor replace/revert, detach, and full reboot.

### DIFF
--- a/gum/backend-darwin/gummemory-darwin.c
+++ b/gum/backend-darwin/gummemory-darwin.c
@@ -10,7 +10,9 @@
 #include "gum/gumdarwin.h"
 #include "gumdarwin-priv.h"
 #include "gummemory-priv.h"
+#include "gummemory-jailbreak.h"
 
+#include <dlfcn.h>
 #include <errno.h>
 #include <unistd.h>
 #include <libkern/OSCacheControl.h>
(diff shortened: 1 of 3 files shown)


TheDauntless:
Tested in iPhone 11 Pro, iOS 26.0.1. Hooking with Interceptor.attach works after jailbreaking with modified Dopamine version.

morzelowski:
Tested successfully on **iPhone 11 / iOS 26.0** as part of the full 3-patch set:

- opa334/Dopamine#999
- opa334/Dopamine#1000
- frida/frida-gum#1150

I built Frida from the 17.18.0 tree with this frida-gum patch and used a custom Dopamine build containing both corresponding Dopamine PRs.

I also verified that the resulting Frida build was using `jb_get_memory_hooks`.

`frida-ps -U` worked and Frida injection no longer caused a kernel panic/reboot.

This full patch set is working for me on iOS 26.0.

devnoname120:
@oleavr Great! Once the Dopamine PRs are merged frida support on jailbroken devices will be fixed once and for good :)

PR #1151, opened 2026-09-16
module-registry: Validate auxv program headers

tntljc:
## Summary

This fixes a crash found when using frida-gum with Wine's preloader. The fix
validates the complete program-header table from `/proc/self/auxv` before
parsing it. If that table is stale or inaccessible, the existing stack-auxv
fallback supplies the current image metadata.

## Problem

When frida-gum initializes inside a process launched through Wine's preloader,
ModuleRegistry may read stale program-header metadata from `/proc/self/auxv`
and crash. Wine's preloader replaces the initial ELF image without the kernel
refreshing that auxv data, then reserves the former `AT_PHDR` mapping with
`PROT_NONE`.

The proc auxv therefore points into an inaccessible reservation, while the
reconstructed auxv on the current stack contains valid headers for the target
image. Treating the proc-auxv result as unavailable lets the existing fallback
select the valid stack data.

## Testing

- Current main: native Linux x86-64 build, ModuleRegistry tests, and the
  mixed-page `gum_memory_is_readable()` test
- 17.15.5 backport: Linux x86 and x86-64 builds and the same focused tests,
  in Debug and Release
- Wine 9 preloader injection on x86, x86-64, x86-64-to-x86 exec, and
  alternating-bitness launches


oleavr:
Thanks! 🙌

PR #1153, opened 2026-09-19
Hook functions that use both intra-procedure scratch registers

jiska2342:
`gum_arm64_relocator_can_relocate()` hands Interceptor a scratch register, and accepts one only if the relocated block never mentions it, trying X16 and then X17. An arm64e authentication prologue uses both:

    mov   x17, x1
    ldr   x16, [x17, #0x10]!
    cbz   x16, #...
    autia x16, x17

so there is no scratch register and `Interceptor.attach()` fails with `"unable to intercept function; please file a bug"`. On iOS 27 that includes commonly used functions across libdispatch, libobjc and libsystem_c, including `dispatch_sync`, `dispatch_once`, `dispatch_retain`, `objc_msgSendSuper2` and `_objc_msgForward`.

Tested on an SRD with iOS 27.0 beta 7.


oleavr:
Thanks a lot for this, Jiska — really solid work, and a great catch on iOS 27. 🙏

I reworked the scratch-register selection around liveness analysis (so it also holds for mid-function probes and picks per-exit registers), and pushed it to your branch along with a few things it surfaced:

- arm64: Pick scratch registers by liveness — your commit, reworked.
- darwin: Bless allocated code pages via page plan — trampoline pages weren't made executable under the jailed debugger, so hooked calls hung; this was the real blocker on jailed iOS 27.
- interceptor: Distinguish undecodable targets — attaching to non-code now reports clearly instead of "please file a bug".
- gumjs: Name function or instruction in errors — message reflects what you passed.

Verified on an iPhone 17 Pro Max (iOS 27): dispatch_once, dispatch_sync, objc_msgSendSuper2 and friends all hook and run now.

Merging — thanks again! 🎉

